PR #7, opened 2021-01-10
Switching off k-mesh spacing by k_mesh_spacing=None

jan-janssen:
Transfer https://github.com/pyiron/pyiron/pull/1237/


jan-janssen:
@samwaseda Can you take a look at this pull request?

samwaseda:
I'm fine with this change, but I don't really understand what this was made for.

jan-janssen:
I thought the issue was that the kpoint spacing was set before the structure. For example in phonon calculations when phonopy is increasing the supercell size. At least that is what I thought.

samwaseda:
But isn't it `modify_kpoints` that's called when the structure is set afterwards? Is `set_kpoints` also called?

jan-janssen:
Ok, I notice that nobody is really sure about this right now. 

@yury-lysogorskiy Maybe you can jump in and take a look. 

Otherwise I would suggest to close this pull request and wait until somebody is looking for this functionality.

PR #10, opened 2021-01-11
Add  property spin

### LINKED ISSUE #9: Atomic spins could be an attribute
Right now we need `structure.set_initial_magnetic_moment(*args, **kwargs)` to add spins to our atomic structures. But spin is just a per-atom property like position -- can we just set these as an attribute, e.g.

```python
structure = pr.create.structure.ase_bulk('Fe', cubic=True)
print(structure.spins)
>>> None

structure.spins = [0.1, 0.5]
# Runs a regular magnetic calculation

structure.spins = np.array([
    [0., 0., 0.1],
    [0.1, 0., 0.]
])
# Runs non-collinear magnetism
```

sudarsan-surendralal:
See linked issue

### DIFF
--- a/pyiron_atomistics/atomistics/structure/atoms.py
+++ b/pyiron_atomistics/atomistics/structure/atoms.py
@@ -219,7 +219,7 @@ class Atoms(ASEAtoms):
             else:
                 self.new_array("initial_magmoms", val)
         else:
-            del self.arrays["initial_magmoms"]
+            self.set_array('initial_magmoms', None)
 
     @property
     def visualize(self):
--- a/pyiron_atomistics/atomistics/structure/atoms.py
+++ b/pyiron_atomistics/atomistics/structure/atoms.py
@@ -213,6 +213,7 @@ class Atoms(ASEAtoms):
     @spins.setter
     def spins(self, val):
         if val is not None:
+            val = np.asarray(val)
             if self.has("initial_magmoms"):
                 self.set_array("initial_magmoms", val)
             else:
--- a/pyiron_atomistics/atomistics/structure/atoms.py
+++ b/pyiron_atomistics/atomistics/structure/atoms.py
@@ -2548,12 +2548,9 @@ class Atoms(ASEAtoms):
                 self.add_tag(spin=None)
             for ind, spin in enumerate(magmoms):
                 self.spin[ind] = spin
-        # ASE part
-        if magmoms is None:
-            self.set_array('initial_magmoms', None)
-        else:
-            magmoms = np.asarray(magmoms)
-            self.arrays['initial_magmoms'] = magmoms
+        if magmoms is not None:
+            self.spins = None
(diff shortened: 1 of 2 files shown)


pyiron_atomistics/atomistics/structure/atoms.py
```diff
@@ -196,6 +196,31 @@ def __init__(
...
+        Magnetic spins for each atom in the structure
+
+        Returns:
+            numpy.ndarray/list: The magnetic moments for reach atom as a single value or a vector (non-collinear spins)
+
+        """
+        if self.has("initial_magmoms"):
+            return self.get_array("initial_magmoms")
+        else:
+            return
```
pmrv, line 211: Unless this is how we do it already with other attributes, I think this should be an `AttributeError`.

sudarsan-surendralal: If we have to be compatible to ase, we shouldn't raise an `AttributeError` and should simple return `None`.

pmrv: Ah, bummer. :+1:

sudarsan-surendralal:
@samwaseda can you review this?

samwaseda (commented):
This is again a super dangerous kind of `property` usage, because the following lines are probably super counter-intuitive:

```
>>> structure.set_initial_magnetic_moments(len(structure)*[2])
>>> structure.spins[0] = 0
>>> print(structure.spins[0])
2
```

sudarsan-surendralal:
> 
> 
> This is again a super dangerous kind of `property` usage, because the following lines are probably super counter-intuitive:
> 
> ```
> >>> structure.set_initial_magnetic_moments(len(structure)*[2])
> >>> structure.spins[0] = 0
> >>> print(structure.spins[0])
> 2
> ```

Thanks for pointing this out. This is now fixed. The getter now points to the actual array which is how ase does it.

pyiron_atomistics/atomistics/structure/atoms.py
```diff
@@ -2523,12 +2549,9 @@ def set_initial_magnetic_moments(self, magmoms=None):
                 self.add_tag(spin=None)
             for ind, spin in enumerate(magmoms):
                 self.spin[ind] = spin
-        # ASE part
-        if magmoms is None:
-            self.set_array('initial_magmoms', None)
-        else:
-            magmoms = np.asarray(magmoms)
-            self.arrays['initial_magmoms'] = magmoms
+        if magmoms is not None:
```
samwaseda: Isn't this line redundant? This means, if `magmoms` is not `None`, it's explicitly set to `None` and if `magmoms` is `None`, it's set to `magmoms`, which is `None`.

sudarsan-surendralal: This should now be fixed

PR #12, opened 2021-01-13
add docstrings

samwaseda:
I start putting docstrings to some of the most important classes.


samwaseda:
Since it's only about the DocString, I merge it now. Feel free to update it if you see a need.

PR #25, opened 2021-01-18
Lammps: Use change_box when switching from orthogonal box to triclini…

jan-janssen:
…c box


pyiron_atomistics/lammps/interactive.py
```diff
@@ -114,12 +114,18 @@ def interactive_cells_setter(self, cell):
             warnings.warn('set_relative() is deprecated as of 2020-02-26. It is not guaranteed from pyiron_atomistics vers. 0.3')
```
samwaseda, line 114: It's totally unrelated, but since this feature is badly deprecated, I guess we shouldn't forget to erase it after this PR.

pyiron_atomistics/lammps/interactive.py
```diff
@@ -114,12 +114,18 @@ def interactive_cells_setter(self, cell):
...
 
         if is_skewed and is_scaled:
+            self._interactive_lib_command(
+                "change_box all triclinic"
+            )
             self._interactive_lib_command(
                 "change_box all x final 0 %f y final 0 %f z final 0 %f \
                  xy final %f xz final %f yz final %f triclinic remap units box"
```
samwaseda: If we are setting `change_box all triclinic` in the command above, we can probably erase `triclinic` here (same for the command below).

samwaseda (changes requested):
This doesn't work if we go from a skewed box to an orthogonal box, because the shear components will be ignored (although to be more precise, this problem was already present before this PR).

jan-janssen:
I need this fix for the old elastic constants job which is currently not working. 

With the current master branch I get the following error message: 
```
ERROR: Cannot change box tilt factors for orthogonal box (../change_box.cpp:249)
```

liamhuber (commented):
My Lammps-fu isn't good enough to just look at this and understand what the impact will be. I don't have any objections to the change, but it would be very helpful if the tests were also updated to demonstrate the behaviour you're trying to get.

samwaseda:
> I need this fix for the old elastic constants job which is currently not working.
> 
> With the current master branch I get the following error message:
> 
> ```
> ERROR: Cannot change box tilt factors for orthogonal box (../change_box.cpp:249)
> ```

Sorry I don't understand what you want to say with this.

jan-janssen:
> This doesn't work if we go from a skewed box to an orthogonal box, because the shear components will be ignored (although to be more precise, this problem was already present before this PR).

I agree that this is an issue which should be addressed but that is not part of this pull request.

samwaseda:
> I agree that this is an issue which should be addressed but that is not part of this pull request.

There I would slightly (but not totally) disagree. In the previous implementation it was not possible to go from orthogonal to skewed or other way around. If you make it possible to go from orthogonal to skewed with the new implementation I would find it more logical to make it also possible to go from skewed to orthogonal. Especially regarding the fact that going from skewed to orthogonal doesn’t even raise an error, so pyiron silently gives back wrong output and the user might never know that.

jan-janssen:
> In the previous implementation it was not possible to go from orthogonal to skewed or other way around.

In the previous implementation all structures were triclinic, this changed in https://github.com/pyiron/pyiron/pull/1083 so all I want is to restore functionality we had before.

samwaseda:
> In the previous implementation all structures were triclinic, this changed in [pyiron/pyiron#1083](https://github.com/pyiron/pyiron/pull/1083) so all I want is to restore functionality we had before.

That's perfectly fine by me, but the new implementation still doesn't rectify the problem that it's not possible to go from a skewed box to an orthogonal box.

pyiron_atomistics/lammps/interactive.py
```diff
@@ -114,23 +114,35 @@ def interactive_cells_setter(self, cell):
...
+            self._interactive_lib_command(
+                "change_box all ortho"
+            )
             self._interactive_lib_command(
                 "change_box all x final 0 %f y final 0 %f z final 0 %f remap units box"
                 % (lx, ly, lz)
             )
         else:  # is neither skewed nor scaled
+            self._interactive_lib_command(
+                "change_box all ortho"
```
jan-janssen, line 144: @samwaseda Is this what you meant?

samwaseda: No this won't work either, because `xy` will still retain the previous value.

jan-janssen: @samwaseda Then I am not sure what you want and prefer to merge this pull request and address your request in a separate pull request.

samwaseda: Wait "I don't understand what you're saying so I'm gonna merge it" is anyway not quite acceptable. Let me explain the problem one more time and then we're gonna talk about whether to still ignore the problem and merge it or not, ok?

So what pyiron does is set `lx ly lz xy xz yz` if the box is skewed, and `lx ly z` if the box is orthogonal. Imagine we have an orthogonal box at the beginning and go through the following case:

1. Set `xy=0.1` -> run
2. Set `xy=0` -> run

What happens inside pyiron is:

1. `xy=0.1`: box is skewed -> `change_box lx ly lz xy xz yz` -> run
2. `xy=0`: box is orthogonal -> `change_box lx ly lz` -> run


And as you can see in the second step, `xy` remains untouched, while pyiron should have actively set `xy=0`.

I hope this explanation was clear.

jan-janssen: I understand the physical implications - I just say it was not working before and I do not know how to fix it, because we would have to read the current box inside lammps and then make a decision how to change this box which most likely leads to performance penalties. Therefore I suggest to merge the current pull request because it at least fixes the old elastic constant protocol developed by yury, which is not working with the current master branch.

liamhuber (changes requested):
I am still not tracking the Lammps super specifically, but I am trying to follow the discussion. Below is my current understanding. Ideal would be some test to ensure the desired pyiron behaviour, but I'm not going to force that issue.

# Old

EDIT: Sam corrected my understanding
EDIT again: @samwaseda is the post correct now?
- Change skew to orthogonal) ~~Fail with error~~ Silently does the wrong thing -- AAAHHH!
- Change orthogonal to skew) ~~Fail with error~~ ~~Silently does the wrong thing -- AAAHHH!~~ Fails with error

# New (unacceptable)

- Change skew to orthogonal) Silently does the wrong thing
- Change orthogonal to skew) Works

# Requested

- Change skew to orthogonal) Fails with error
- Change orthogonal to skew) Works

pyiron_atomistics/lammps/interactive.py
samwaseda: I have the feeling that it's better to leave it then, because, again, going from skewed to orthogonal won't raise an error, but silently returns wrong output. So by making it possible to go from orthogonal to skewed with this PR, we might be making it even more difficult to realise where the problem comes from later.

So I would say, either we solve both problems or nothing. But I'll accept also other decisions if other people prefer a different way. What do you guys think? @liamhuber @pmrv @sudarsan-surendralal

samwaseda:
> # Old
> * Change skew to orthogonal) Fail with error

No the old version also does the wrong thing silently.

pyiron_atomistics/lammps/interactive.py
liamhuber: Aha. Also from your other comment I understand a bit better. Failing silently is the worst of devils and should be exorcised immediately. I agree entirely that making it work in one direction but not the other makes this an even worse trap.

We definitely don't need the functionality to work both ways, but we should make it fail ASAP, *especially* if we now make it work one way.

samwaseda: Ok then maybe the quickest solution is to set `change_box orthogonal`, because then LAMMPS complains, right?

liamhuber: > because then LAMMPS complains, right?

*cough* need tests *cough* 😉

jan-janssen:
This starts to get fundamental. This pull request addresses one part of the issue above: 

# Old 
* Change orthogonal to skew - fails 

# New 
* Change orthogonal to skew - works

While the other changes would be nice to have they are not part of this pull request, mainly because I do not know how to get the information about the current structure from LAMMPS in an efficient way. 

The tests for this part would require running lammps and I thought we agreed that we do not want to run simulation codes in the regular unit tests.

samwaseda:
> # Old
> EDIT: Sam corrected my understanding
> 
> * Change orthogonal to skew) ~Fail with error~ Silently does the wrong thing -- AAAHHH!

No, there it fails 😎

jan-janssen:
The failing unit tests are addressed in https://github.com/pyiron/pyiron_atomistics/pull/28

liamhuber:
> This starts to get fundamental. This pull request addresses one part of the issue above:

My argument is that these parts are not necessarily separable issues. 

(A,B):Fails silently, (B,A):Fails <-- Bad

(A,B):Fails silently, (B,A):Works <-- *much worse*

Silent failures should be high priority to squash. Making a reverse operation fully functional only exacerbates the problem and raises the priority. The reverse doesn't need to work, it just needs to fail cleanly.

> The tests for this part would require running lammps and I thought we agreed that we do not want to run simulation codes in the regular unit tests.

Agreed, we should probably avoid this in the regular unit tests, which is why I'm absolutely *not* going to insist such a test be part of this PR. However, this PR definitely highlights the need for tests, be they unit or integration.

jan-janssen:
If we do not fix this the calculation of elastic constants using the old class from Yury does not work in interactive mode. But this class is already used in current high through projects like http://atomistictools.org . This functionality was broken when we moved away from having triclinic cells as the default. I am against dropping existing functionality in favour of abstract ideals.

samwaseda:
> If we do not fix this the calculation of elastic constants using the old class from Yury does not work in interactive mode. But this class is already used in current high through projects like http://atomistictools.org . This functionality was broken when we moved away from having triclinic cells as the default. I am against dropping existing functionality in favour of abstract ideals.

- It should still work if the job is static.
- Whether it works or not in the interactive mode depends also on the order of execution: if the orthogonal box doesn't happen to be the very first run, then it returns wrong results silently, which is even worse than now.

And I wouldn't call it an "abstract ideal". It's a real problem.

jan-janssen:
So then the solution is to reset `xy`, `xz` and `yz` to 0.0 and afterwards change the box to orthogonal?

liamhuber:
> So then the solution is to reset `xy`, `xz` and `yz` to 0.0 and afterwards change the box to orthogonal?

Shouldn't they be `none` for orthogonal?

jan-janssen:
> Shouldn't they be `none` for orthogonal?

I do not think LAMMPS supports `none` here that is why afterwards I change the cell to `orthogonal`.

samwaseda:
I guess the first thing we can try is to set `change_box orthogonal` as you did in one of the commits and hope that LAMMPS complains if we go from triclinic to orthogonal. If it doesn't complain, we probably have to find a different solution...

liamhuber:
> I guess the first thing we can try is to set `change_box orthogonal` as you did in one of the commits and hope that LAMMPS complains if we go from triclinic to orthogonal. If it doesn't complain, we probably have to find a different solution...

The Lammps-crash error message is a little cryptic, but if this does manage to crash things I'd put this in the "good enough" category. 

Since the check needs to be performed anyway, it would be lovely to have it codified as a tiny integration test.

jan-janssen:
> I guess the first thing we can try is to set `change_box orthogonal` as you did in one of the commits and hope that LAMMPS complains if we go from triclinic to orthogonal. If it doesn't complain, we probably have to find a different solution...

That is what is used in the current version, first I resent `xy`, `xz` and `yz` to 0.0 and then I switch to `ortho`

liamhuber:
Then is Lammps crashing on the reverse transformation as hoped? If not some further change to force the crash (either in pyiron or just directly in lammps) is still needed

jan-janssen:
> Then is Lammps crashing on the reverse transformation as hoped? If not some further change to force the crash (either in pyiron or just directly in lammps) is still needed

No, also the second part is now fixed.

liamhuber (approved):
I won't insist on it, but I think it would be excellent if you included whatever code you're currently using to verify the forward and backward transitions as an integration test (might need to make a new folder for it, but can follow the same format as contrib; also no immediate need to get integration in the ci as even having it there for manual running would be super helpful IMO)

samwaseda:
> 
> That is what is used in the current version, first I resent `xy`, `xz` and `yz` to 0.0 and then I switch to `ortho`

Ah now I get it. I also thought about this possibility, but wasn’t sure if the performance is not badly affected by it. The question is also whether it really makes sense to use both triclinic and orthogonal, or we simply use only triclinic with all shear components equal to zero. Anyway, for now I would accept the solution you suggested (although again I think we should do a performance test).

liamhuber (approved):

Issue #45, opened 2021-01-24
Difference between _key_from_property and _lst_from_property?


pmrv:
Does anyone with interactive experience know what the difference between these two functions is?

https://github.com/pyiron/pyiron_atomistics/blob/bc4c32156d67d584c00f97a1877ebad598768261/pyiron_atomistics/atomistics/job/interactive.py#L444

https://github.com/pyiron/pyiron_atomistics/blob/bc4c32156d67d584c00f97a1877ebad598768261/pyiron_atomistics/atomistics/job/interactive.py#L424

They seem do to the same to me, so if I don't hear from you I'll throw one of them under the bus.

jan-janssen:
As far as I can see the only difference is that for one tilmestep we either store a single entry - like the potential energy - or a matrix like the cell dimensions.

Issue #41, opened 2021-01-22
get_structure() is slow


pmrv:
I did some brief profiling that identified two problems:
1. the connection to the database can take a significant amount of time, a few seconds per call and it seems to get triggered a few times.
2. conversions of numpy arrays from/to python lists when reading the positions and indices.

To load the final structure of ~1000 atoms from a 20000 steps trajectory takes around *30s*, around 1/3 of that in the database connection and 2/3 in the numpy conversions.  I'll look into these conversions and cut out as much as I can, but I will need some help on the database part, since I'm not familiar with its setup.

I put a notebook and an exported project containing this job on the cluster world-readable at
/u/zora/get_structure_profile.tar.gz
/u/zora/GetStructureProfile.ipynb

Below some example traces.  `tottime` is the total internal time, i.e. how long the function takes minus how long its callees take.
``` 
        40013 function calls (39849 primitive calls) in 38.418 seconds

   Ordered by: internal time
   List reduced from 1010 to 10 due to restriction <10>

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
        7   14.327    2.047   14.327    2.047 {built-in method psycopg2._psycopg._connect}
      108   12.191    0.113   12.744    0.118 {built-in method numpy.array}
        3   10.082    3.361   10.082    3.361 {method 'tolist' of 'numpy.ndarray' objects}
        1    1.048    1.048   24.091   24.091 /u/zora/software/pyiron_atomistics/pyiron_atomistics/atomistics/job/interactive.py:491(positions)
        4    0.553    0.138    0.553    0.138 /u/system/SLES12/soft/pyiron/dev/anaconda3/lib/python3.7/site-packages/h5py/_hl/dataset.py:947(read_direct)
        7    0.052    0.007    0.052    0.007 {method 'execute' of 'psycopg2.extensions.cursor' objects}
        1    0.033    0.033    3.043    3.043 /u/zora/software/pyiron_atomistics/pyiron_atomistics/atomistics/job/interactive.py:467(indices)
        7    0.028    0.004    0.028    0.004 {method 'rollback' of 'psycopg2.extensions.connection' objects}
       11    0.027    0.002    0.027    0.002 {method '_g_list_group' of 'tables.hdf5extension.Group' objects}
       62    0.009    0.000    0.011    0.000 {method '_g_new' of 'tables.hdf5extension.File' objects}
```

```
         40013 function calls (39849 primitive calls) in 33.868 seconds

[...]

PR #46, opened 2021-01-24
Get structure fast

pmrv:
See #41.

This cuts the time on the same sample structure down to \~2s.  There's a similar overhead (\~.5s) for other properties of `GenericInteractiveOutput` that are read via `_key_from_property`.  Pending feedback on #45 I'll cut that too.  I'll assume the rest is due to us reading the full arrays from HDF5 and then indexing them even if we only need a tiny part of them.  But to leverage that we'd have to touch HDF5 code and anyways the current speed up is enough to make it bearable for the POTENTIALS workshop.

For some reason this patch also makes the waiting time on the SQL connection disappear.

Example trace after the patch
```
         70225 function calls (69858 primitive calls) in 1.956 seconds

   Ordered by: internal time
   List reduced from 1008 to 10 due to restriction <10>

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
      115    0.803    0.007    1.309    0.011 {built-in method numpy.array}
        8    0.504    0.063    0.505    0.063 /u/system/SLES12/soft/pyiron/dev/anaconda3/lib/python3.7/site-packages/h5py/_hl/dataset.py:947(read_direct)
        2    0.304    0.152    0.304    0.152 {method 'tolist' of 'numpy.ndarray' objects}
       14    0.155    0.011    0.156    0.011 {built-in method psycopg2._psycopg._connect}
       14    0.049    0.003    0.049    0.003 {method 'execute' of 'psycopg2.extensions.cursor' objects}
      124    0.021    0.000    0.024    0.000 {method '_g_new' of 'tables.hdf5extension.File' objects}
        2    0.011    0.005    0.545    0.273 /u/zora/software/pyiron_atomistics/pyiron_atomistics/atomistics/job/interactive.py:477(cells)
       14    0.008    0.001    0.008    0.001 {method 'rollback' of 'psycopg2.extensions.connection' objects}
      286    0.004    0.000    0.004    0.000 {method '_g_get_objinfo' of 'tables.hdf5extension.Group' objects}
     5548    0.004    0.000    0.004    0.000 /u/system/SLES12/soft/pyiron/dev/anaconda3/lib/python3.7/site-packages/tables/group.py:838(__setattr__)
```

### DIFF
--- a/pyiron_atomistics/atomistics/job/interactive.py
+++ b/pyiron_atomistics/atomistics/job/interactive.py
@@ -443,26 +443,28 @@ class GenericInteractiveOutput(GenericOutput):
 
     def _lst_from_property(self, key, prop=None):
         """
+        Fetch latest values for the given key and property.
+
+        Values are first looked up in the interactive cache, then in the interactive group in the HDF5 file
+        ('output/interactive/<key>') and if not found there via the given prop function.
 
         Args:
-            key (str):
-            prop (function):
+            key (str): output key
+            prop (function): Unbound property getter from GenericOutput
 
         Returns:
-
+            :class:`numpy.ndarray`: collected values from all previous steps
         """
-        return_lst = self._lst_from_cache(key)
-        hdf5_output = self._key_from_hdf(key)
-        if hdf5_output is not None and len(hdf5_output) != 0:
-            if isinstance(hdf5_output[-1], list):
-                return_lst = [np.array(out) for out in hdf5_output] + return_lst
-            else:
-                return_lst = hdf5_output.tolist() + return_lst
-        elif prop is not None:
-            prop_result = prop(self)
-            if prop_result is not None:
-                return_lst = prop_result.tolist() + return_lst
-        return np.array(return_lst)
+        cached = self._lst_from_cache(key)
+        fetched = self._key_from_hdf(key)
+        if fetched is None or len(fetched) == 0 and prop is not None:
+            fetched = prop(self)
+        if fetched is None or len(fetched) == 0:
+            return cached
+        elif len(cached) == 0:
+            return fetched
+        else:
+            return np.concatenate((cached, fetched))
 
     @property
     def indices(self):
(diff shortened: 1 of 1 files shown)


jan-janssen (changes requested):
I am aware that we do not have the required unit tests, but I have the feeling we are missing some case if we remove the property part. The cases from my memory are: 

- some output already written to the HDF5 file `[output/interactive]`, some output in the cache - this seems to be handled correctly. 
- but for the case that the job was executed previously and the output is already copied to `[output/generic]` then we need to call the property to get this information, this is no longer included in the new version.

pyiron_atomistics/atomistics/job/interactive.py
```diff
@@ -441,32 +445,34 @@ def _key_from_property(self, key, prop):
...
-                return_lst = hdf5_output.tolist() + return_lst
-        elif prop is not None:
-            prop_result = prop(self)
-            if prop_result is not None:
-                return_lst = prop_result.tolist() + return_lst
-        return np.array(return_lst)
+        cached = self._lst_from_cache(key)
+        fetched = self._key_from_hdf(key)
+        if fetched is None or len(fetched) == 0:
+            fetched = getattr(super(), key)
```
pmrv: Loading from `output/generic` should be handled here, in line 405.  `super()` resolves to `GenericOutput` and the `getattr()` then gets the value of `key` on that instance, which invokes the property getter.  I'll add a comment.  I'll be honest though and say that I didn't do testing beyond my sample notebook.

EDIT for clarity.

pmrv:
I've checked that it works with interactive jobs during and after the run as well as after reloading with the attached notebook.  Ideally I'd want to also look into #45, but I don't understand the code well enough and lack time for it.

Notebook is [here](https://gist.github.com/pmrv/c8a1ecb5b423f371d6c23dee7f7f15c3).

jan-janssen (approved):
Looks good to me - I guess adding the integration tests can be done in a separate pull request

PR #50, opened 2021-01-28
Faster ase to pyiron

### LINKED ISSUE #48: ase_to_pyiron is slow
I'm currently importing a large number of ase structures into a structure container and have noticed it is exceedingly slow.  Importing 1000 structures takes ~40s, 10000 ~4min or longer.  After some profiling I identified `ase_to_pyiron` and this line

https://github.com/pyiron/pyiron_atomistics/blob/d9145e632d8871cfe8f6b4bdd1e867be8f7d773a/pyiron_atomistics/atomistics/structure/atoms.py#L163

as the culprits.  It seems for the element conversion `mendelev` consults some kind of SQL database (???) that takes a lot of time.  Can I skip this conversion somehow?

sudarsan-surendralal:


pmrv:
Profiling shows no change, so I'm assuming the SQL database from mendeleev did some caching already.  Since it's not critical for the structure import that the other workshop organizers depend on, let's discuss the problem in the meeting tomorrow.

PR #63, opened 2021-02-03
use pyiron-internal is_skewed

samwaseda:
Certain fixes in LAMMPS do not work with `fix orthogonal` etc., which used to work before (essentially before [this commit](https://github.com/pyiron/pyiron_atomistics/commit/04ad909b78b056ea8cfa9433e03050cf7ee25f64)). As I didn't really see how to check whether LAMMPS sees `triclinic` or `orthogonal` from pyiron, I decided to implement it inside pyiron. Hopefully this one is acceptable...

### DIFF
--- a/pyiron_atomistics/atomistics/structure/atoms.py
+++ b/pyiron_atomistics/atomistics/structure/atoms.py
@@ -1948,6 +1948,25 @@ class Atoms(ASEAtoms):
         return self._analyse.pyscal_voronoi_volume()
     get_voronoi_volume.__doc__ = Analyse.pyscal_voronoi_volume.__doc__
 
+    def is_skewed(self, tolerance=1.0e-8):
+        """
+        Check whether the simulation box is skewed/sheared. The algorithm compares the box volume
+        and the product of the box length in each direction. If these numbers do not match, the box
+        is considered to be skewed and the function returns True
+
+        Args:
+            tolerance (float): Relative tolerance above which the structure is considered as skewed
+
+        Returns:
+            (bool): Whether the box is skewed or not.
+        """
+        volume = self.get_volume()
+        prod = np.linalg.norm(self.cell, axis=-1).prod()
+        if volume > 0:
+            if abs(volume-prod)/volume < tolerance:
+                return False
+        return True
+
     def find_mic(self, v, vectors=True):
         """
         Find vectors following minimum image convention (mic). In principle this
--- a/pyiron_atomistics/lammps/interactive.py
+++ b/pyiron_atomistics/lammps/interactive.py
@@ -123,9 +123,6 @@ class LammpsInteractive(LammpsBase, GenericInteractive):
             )
         elif was_skewed:
             self._interactive_lib_command(
-                "change_box all triclinic"
-            )
-            self._interactive_lib_command(
                 "change_box all x final 0 %f y final 0 %f z final 0 %f \
                 xy final %f xz final %f yz final %f remap units box"
                 % (lx, ly, lz, 0.0, 0.0, 0.0)
--- a/pyiron_atomistics/lammps/interactive.py
+++ b/pyiron_atomistics/lammps/interactive.py
@@ -108,8 +108,8 @@ class LammpsInteractive(LammpsBase, GenericInteractive):
                 "Warning: setting upper trangular matrix might slow down the calculation"
             )
 
-        is_skewed = self._structure_current.is_skewed()
-        was_skewed = self._structure_previous.is_skewed()
+        is_skewed = self._structure_current.is_skewed(tolerance=1.0e-8)
+        was_skewed = self._structure_previous.is_skewed(tolerance=1.0e-8)
 
         if is_skewed:
             if not was_skewed:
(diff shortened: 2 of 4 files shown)


jan-janssen:
But is this the same? For example in the case when we do a monte carlo simulation in combination with minimising each step. Then all cells in pyiron are orthogonal while the cells inside LAMMPS are not.

jan-janssen:
> Certain fixes in LAMMPS do not work with `fix orthogonal` etc., 

Can you provide some examples what does not work?

samwaseda:
> But is this the same? For example in the case when we do a monte carlo simulation in combination with minimising each step. Then all cells in pyiron are orthogonal while the cells inside LAMMPS are not.

That's not true. If the pyiron cell is orthogonal, then we use `change_box all ortho`, in which case it's not allowed to become triclinic. I don't see a problem with changing this, but it's unrelated to this PR.

jan-janssen:
You are right, still I would be interested what fails. As I thought orthogonal boxes are the default for most cases.

samwaseda:
> You are right, still I would be interested what fails. As I thought orthogonal boxes are the default for most cases.

It's actually the callback function. If you are really really interested I can try to think of an easy example, but that's gonna probably still be somewhat cumbersome (both to create one and to read it)...

samwaseda:
But actually it's maybe anyway time to think about making simple examples for callback functions, so I'm gonna make one anyway.

pyiron_atomistics/atomistics/structure/atoms.py
```diff
@@ -1948,6 +1948,25 @@ def get_voronoi_volume(self):
         return self._analyse.pyscal_voronoi_volume()
     get_voronoi_volume.__doc__ = Analyse.pyscal_voronoi_volume.__doc__
 
+    def is_skewed(self, tolerance=1.0e-8):
```
pmrv, line 1951: This should work, but isn't it more straightforward to check the dot product between cell vectors?  Your reasoning is documented though, so I don't mind too much.

samwaseda: Uh that's a good point. I don't have a strong opinion about it. The only one thing is that I want to insist on making `tolerance` a dimensionless value, in which case the computational cost is probably very much the same. If other people prefer dot product I can also change it.

pyiron_atomistics/lammps/interactive.py
```diff
@@ -108,30 +108,20 @@ def interactive_cells_setter(self, cell):
...
-                "change_box all triclinic"
-            )
-            self._interactive_lib_command(
-                "change_box all x final 0 %f y final 0 %f z final 0 %f \
-                xy final %f xz final %f yz final %f units box"
-                % (lx, ly, lz, xy, xz, yz)
-            )
-        elif not is_skewed and is_scaled:
+        elif was_skewed:
             self._interactive_lib_command(
```
pmrv, line 125: I know that this line is not part of the PR, but why is it setting first triclinic and then reverting to orhonormal?

tests/atomistics/structure/test_atoms.py
```diff
@@ -1607,6 +1607,14 @@ def test_get_wrapped_coordinates(self):
             np.linalg.norm(position-structure.cell*0.1), 0
         )
 
+    def test_is_skewed(self):
```
pmrv, line 1610: I think this would easier to read if `is_skewed` is called with an explicit tolerance.  Also adding a message to the assert would make it failure clearer when the tests fail in the future.

pyiron_atomistics/lammps/interactive.py
samwaseda: Good point. I changed it.

It was because in the old implementation we didn't check beforehand whether the box was skewed or not. With this, we made sure that the box first becomes triclinic, which allowed us to change the shear components to 0.

tests/atomistics/structure/test_atoms.py
samwaseda: Done!

samwaseda: I put a message. Was this more or less what you suggested?

pmrv: Perfect, thanks!

pmrv (approved):

samwaseda:
I found out that actually it was not the callback function that causes the problem, but apparently `pressure=0` is not compatible with the old implementation. So the following lines failed:

```
job = pr.create_job('Lammps', 'md')
job.structure = pr.create_ase_bulk('Al', cubic=True)
job.calc_md(temperature=300, pressure=0)
job.interactive_open()
job.run()
job.run()
job.interactive_close()
```

And with this PR this problem was fixed.

With this, I guess it's clear to everyone that this PR is indispensable.

samwaseda:
And since I promised @jan-janssen , here's a super simple notebook for callback:

[callback.ipynb.txt](https://github.com/pyiron/pyiron_atomistics/files/5926653/callback.ipynb.txt)

jan-janssen (approved):
Looks good to me

PR #74, opened 2021-02-11
automatic triclinic when potentially triclinic

### LINKED ISSUE #38: Automatic triclinic boxes in LAMMPS?
When I run a LAMMPS simulation with a barostat and I also apply a barostat to shear components, the simulation can fail when the input simulation cell is orthogonal:

```
job2 = pr.create_job(pr.job_type.Lammps, "test2")
job2.structure = <something orthogonal>
job2.potential = ...
job2.calc_md(
    temperature=1000.0,
    temperature_damping_timescale=100.0,
    pressure=[0.0, 0.0, 0.0, 0.0, 0.0, 0.0],
    pressure_damping_timescale=500.0,
    time_step=1.0,
    n_ionic_steps=100000,
)
```
The reason is that LAMMPS needs to be told if shear of a box is allowed. The command is `change_box all triclinic`. Shouldn't this command be added automatically when applying a barostat with shear components?

samwaseda:
Closes #38


samwaseda:
I'm not really sure if I should create a flag in `Atoms` or in `LammpsBase`. I guess I go for `LammpsBase` for now, but maybe it should be moved to `Atoms` at some point if such a thing is required later on.

samwaseda:
I keep it still Draft because I haven't written unit tests, but it should do what it should.

There's one thing I don't really like about this one: it creates a new flag `_force_skewed` in LAMMPS control, which behaves like an independent flag, even though it's very much dependent on what's written in the input script. I don't really think there's an automatic way to detect "potential skewness" (i.e. possibility that the box becomes skewed during simulation) by looking at the input script, but if there is one, I might prefer that.

t-brink:
As noted in #38, currently it is not allowed to have skewed boxes that are not fully periodic. This check can now be circumvented, I believe. There is good reason for that, because in *some* cases LAMMPS will either fail or give nonsensical results. In *some* cases it works fine. I cannot currently remember the details, they should be documented in the LAMMPS docs.

Some Qs I'm unsure about:

* Aren't there other cases, such as applying a strain rate? There I can imagine starting from a cubic box and shearing it. Or is this still experimental and not officially supported yet?

* What speaks agains always writing a triclinic box, as long as it is fully periodic (this is not a rhetoric question)? And worrying about non-periodic boxes later?
    * Alternative: always write a triclinic box and have checks in calc_minimize()/calc_md() that ensure that the pressure tensor is compatible with the boundaries. I kinda like that idea, thoughts?

* (There are a bunch of changes in this PR which just seem to correct spelling and whitespace issues in docstrings. I did not read them. Are there any real doc changes that should be checked?)

samwaseda:
> * Aren't there other cases, such as applying a strain rate? There I can imagine starting from a cubic box and shearing it. Or is this still experimental and not officially supported yet?

There's no possibility to do something like this right now in pyiron, except if it's an interactive job, in which case it's fine.

> * What speaks agains always writing a triclinic box, as long as it is fully periodic (this is not a rhetoric question)? And worrying about non-periodic boxes later?
>   

No idea. The only one thing I can imagine is efficiency. If you show us that it's exactly the same, I guess it's indeed something we should think about.

>   * Alternative: always write a triclinic box and have checks in calc_minimize()/calc_md() that ensure that the pressure tensor is compatible with the boundaries. I kinda like that idea, thoughts?

I guess currently we don't store the input pressures as such, so for the case that the structure is set after the pressures are defined, we have to make a proper input field for pressures.

> * (There are a bunch of changes in this PR which just seem to correct spelling and whitespace issues in docstrings. I did not read them. Are there any real doc changes that should be checked?)

No you don't need to. It was just super irritating to me when I was editing the codes because those lines were wrapped in a super weird way.

t-brink:
> > * What speaks agains always writing a triclinic box, as long as it is fully periodic (this is not a rhetoric question)? And worrying about non-periodic boxes later?
> 
> No idea. The only one thing I can imagine is efficiency. If you show us that it's exactly the same, I guess it's indeed something we should think about.

Ran each three times on my private computer (bit older), so judge for yourself.

ortho:
```
Loop time of 58.4871 on 4 procs for 10000 steps with 6912 atoms
Loop time of 58.1516 on 4 procs for 10000 steps with 6912 atoms
Loop time of 60.8539 on 4 procs for 10000 steps with 6912 atoms
```

triclinic:
```
Loop time of 61.9571 on 4 procs for 10000 steps with 6912 atoms
Loop time of 63.2048 on 4 procs for 10000 steps with 6912 atoms
Loop time of 62.5691 on 4 procs for 10000 steps with 6912 atoms
```

The difference is small but visible in this example. It probably depends on all sorts of details (number of atoms/core, potential, ...).

samwaseda:
> Ran each three times on my private computer (bit older), so judge for yourself.
> 
> ortho:
> 
> ```
> Loop time of 58.4871 on 4 procs for 10000 steps with 6912 atoms
> Loop time of 58.1516 on 4 procs for 10000 steps with 6912 atoms
> Loop time of 60.8539 on 4 procs for 10000 steps with 6912 atoms
> ```
> 
> triclinic:
> 
> ```
> Loop time of 61.9571 on 4 procs for 10000 steps with 6912 atoms
> Loop time of 63.2048 on 4 procs for 10000 steps with 6912 atoms
> Loop time of 62.5691 on 4 procs for 10000 steps with 6912 atoms
> ```
> 
> The difference is small but visible in this example. It probably depends on all sorts of details (number of atoms/core, potential, ...).

Ok I agree that it's a very small difference, which we might actually ignore. But considering the fact that it's going to be also a relatively special case that the current implementation fails (that would require a static LAMMPS version which shears box internally where the starting box is orthogonal), I guess we can keep the changes suggested in this PR. If such a special case should ever occur, it is still possible to set `_force_skewed = True` manually anyway.

Again, I don't strongly insist on this point, so feel free to oppose :)

t-brink:
I have no strong opinion on always writing "triclinic" boxes or not. It seemed a bit easier to implement to me, but that's up to whoever implements it.


But:

> As noted in #38, currently it is not allowed to have skewed boxes that are not fully periodic. This check can now be circumvented, I believe. There is good reason for that, because in _some_ cases LAMMPS will either fail or give nonsensical results. In _some_ cases it works fine. I cannot currently remember the details, they should be documented in the LAMMPS docs.

This part is still not addressed. Usually there is a check here: https://github.com/pyiron/pyiron_atomistics/blob/da7f89f1a180cde50389938c39f09c90cf6e6e5f/pyiron_atomistics/lammps/structure.py#L118-L121

This check is too broad, but it is safe. If non-PBC can be used with a skewed box depends on which fixes are applied in LAMMPS, unfortunately. At least in the past, LAMMPS did not raise an error, but did something silly when the PBCs where not correct, which we would probably like to avoid. But I will try to find time to see what modern LAMMPS does. If it raises an error that might be good enough to remove any checks instead and rely on LAMMPS to figure this out?

----

From the LAMMPS docs:

> There is no requirement that a triclinic box be periodic in any dimension, though it typically should be in at least the second dimension of the tilt (y in xy) if you want to enforce a shift in periodic boundary conditions across that boundary. Some commands that work with triclinic boxes, e.g. the fix deform and fix npt commands, require periodicity or non-shrink-wrap boundary conditions in specific dimensions. See the command doc pages for details.

Only the docs for fix npt mention these issues, but agrees with the above:

> Xy, xz, and yz can only be barostatted if the simulation domain is triclinic and the second dimension in the keyword (y dimension in xy) is periodic.

----

I would ultimately like everything that is possible in LAMMPS to be supported. That will likely require storing the input stresses and checking them against the PBCs when running the job, I guess. For now, I would be happy with any safe way to do it that allows at least full PBC simulations in order to advance with this.

samwaseda:
After today's discussion (summarised [here])(https://github.com/orgs/pyiron/teams/pyiron/discussions/46/comments/8)), it looks like we're going to have a major update in April or May, so that the features requested here will be more easily handled afterwards. I would suggest that we keep the current structure for now. Would this be acceptable @t-brink ?

t-brink:
What do you mean by "keeping the current structure"? You mean merging the PR as-is or do you mean not merging the PR and waiting for the other changes?

If you want to merge the PR, I would suggest to add something like this

```
        if self._force_skewed and (not (pbc[0] and pbc[1] and pbc[2])):
            raise RuntimeError(
                "Skewed lammps cells MUST have " "PBC == True in all directions!"
            )
```

in `lammps.structure.simulation_cell()` to have the same checks as before.

If you want to hold off the PR in favor of waiting for redesign that's OK for me.

samwaseda:
> If you want to merge the PR, I would suggest to add something like this
> 
> ```
>         if self._force_skewed and (not (pbc[0] and pbc[1] and pbc[2])):
>             raise RuntimeError(
>                 "Skewed lammps cells MUST have " "PBC == True in all directions!"
>             )
> ```
> 
> in `lammps.structure.simulation_cell()` to have the same checks as before.

Fine by me.

pyiron_atomistics/lammps/structure.py
```diff
@@ -115,7 +115,7 @@ def fold(vec, pvec, i):
 
         self.A = apre
 
-        if self.is_skewed() and (not (pbc[0] and pbc[1] and pbc[2])):
+        if self._force_skewed and (not (pbc[0] and pbc[1] and pbc[2])):
```
t-brink: Is it by design that `self.is_skewed() == True` no longer triggers this check? I would perhaps assume that this could be the case, since simulations with skewed and non-PBC are safe as long as the pressure tensor does not touch the shear directions? This tripped me a bit, so maybe a comment explaining the rationale is in order?

t-brink (approved):
I think I approve, see my one comment for a small doubt.

pyiron_atomistics/lammps/structure.py
samwaseda: So your suggestion really totally makes sense, but `_force_skewed` is not part of the class and therefore we'd have to include it either as an argument or an attribute of this class, both of which would be somewhat ugly. Besides, since it's not a clear error, as long as the user knows what they are doing, I think it goes too far if we clearly forbid it. Therefore I replaced the error by a warning.

In any case, we're planning to overhaul the input structure in the near future, which would allow us to have pressures in the input as well, so if the solution I propose here somewhat solves a practical problem, I'd say we can merge it for now. Later when the input structure is there, we can make it more elegant.

t-brink: Erm, I'm not sure I understand your answer?

_You_ added the check using `self._force_skewed`, it was `self.is_skewed()` before. I was wondering if that was relaxing the checks too much, because you can now force skewed boxes in cases where it was not allowed before, namely when you do have a non-orthogonal box and no PBCs, but do not apply any shear stress. This behavior, introduced by your patch, might be OK or not.

samwaseda: > You added the check using self._force_skewed

Yes, I added the check according to what you wrote, but then as the tests failed I realised that `_force_skewed` was not part of the class. I don't really see a nice way to check whether the PBC setting and skewness are compatible from the pyiron side, so I'd rather leave it to LAMMPS to decide.

Actually I don't really understand your comment either...

samwaseda: To be honest, as long as we don't have a clear way to check whether the box setting is compatible with what LAMMPS allows, we should leave it to LAMMPS to decide whether it's allowed or not. Therefore I think it's totally appropriate to make it a warning and not an error here.

t-brink: I was about to write this:

> OK, let's make this easier: Why not emit a warning in `LammpsStructure.simulation_cell()`, which is the only place where the structure is written and which has both `is_skewed() and `_force_skewed` available?
> 
> Add the check inside the `if` here: https://github.com/pyiron/pyiron_atomistics/blob/ed3e79c2e599fbb7c302b5203167d2082ea61cd5/pyiron_atomistics/lammps/structure.py#L286

But then I realized that the `pbc` argument is not passed to `UnfoldingPrism.__init__()` here: https://github.com/pyiron/pyiron_atomistics/blob/ed3e79c2e599fbb7c302b5203167d2082ea61cd5/pyiron_atomistics/lammps/structure.py#L276

So I don't understand if this check in the current code is applied at all or if it completely useless. If you want to figure this out, go ahead, but I'm OK with the code as-is. Especially if an overhaul is planned.

I also agree that a warning in lieu of an error is appropriate.

PR #81, opened 2021-02-19
get voronoi vertices + clustering

samwaseda:
Voronoi vertices and clustering of nearby vertices. This would have been super helpful for our analysis of interstitial sites for H in Ni with grain boundaries.

Doc + tests follow

### DIFF
--- a/pyiron_atomistics/atomistics/structure/analyse.py
+++ b/pyiron_atomistics/atomistics/structure/analyse.py
@@ -207,8 +207,8 @@ class Analyse:
         """    Calculate the Voronoi volume of atoms        """
         return analyse_voronoi_volume(atoms=self._structure)
 
-    def get_voronoi_vertices(self, epsilon=2.6544356738490314e-4, distance_threshold=0):
-        voro = Voronoi(self._structure.get_extended_positions(3)+epsilon)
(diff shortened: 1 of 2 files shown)


pyiron_atomistics/atomistics/structure/analyse.py
```diff
@@ -206,3 +207,41 @@ def pyscal_voronoi_volume(self):
         """    Calculate the Voronoi volume of atoms        """
         return analyse_voronoi_volume(atoms=self._structure)
 
+    def get_voronoi_vertices(self, epsilon=2.6544356738490314e-4, distance_threshold=0, width_buffer=10):
```
pmrv: If the purpose is to find interstitial sites, I believe we can just change the name to reflect that and mention in the docs that it uses a Voronoi teselation. 

Also where's that specific epsilon coming from?

samwaseda: In bcc it doesn't find the octahedral interstitial sites, which is deadly for Fe-C.

The epsilon value comes from my notebook. I just wanted to have a fixed small random number.

pmrv (approved):

pyiron_atomistics/atomistics/structure/analyse.py
niklassiemer: Could we just use `2.5e-4`? This way it looks like it might be important to get the digits right.

PR #94, opened 2021-03-01
Migrate quickff and yaff to pyiron_gpl

### LINKED ISSUE #91: Moving yaff and quickff to pyiron_gpl
@SanderBorgmans Based on the strict interpretation of the `GPL` license it is not allowed to include a `GPL` library in an `BSD` licensed project. This is solved by the `LGPL`. So my suggestion would be to migrate `yaff` and `quickff` to `pyiron_gpl`. From the user side nothing changes as long as you import `pyiron`, because both `pyiron_atomistics` and `pyiron_gpl` are loaded in the background when they are installed. But when the user only loads `pyiron_atomistics` they have to separately import `pyiron_gpl`.

jan-janssen:


jan-janssen:
Once the tests passed this can be merged as the same modules are already included in `pyiron_gpl` https://github.com/pyiron/pyiron_gpl/pull/7

PR #95, opened 2021-03-01
Support for multiple potential file locaitons

jan-janssen:
This allows the user to have an additional potential file in their home directory


jan-janssen:
I tested it with [Binder](https://mybinder.org/v2/gh/jan-janssen/potential_testing/resources)

PR #98, opened 2021-03-02
Fail hard instead of silently returning None

liamhuber:
`pr.structure.create.surface` currently silently returns `None` if the string-valued surface code is something unexpected. I'd rather raise an exception.


sudarsan-surendralal (approved):
Looks good to me!

PR #100, opened 2021-03-03
Get chemical symbols

samwaseda:
I added `neigh.chemical_symbols` which gives the chemical symbols of the neighboring atoms.

I was not sure if I should call it `neigh.get_chemical_symbols()` or make it a property. Since there's no argument to add, I decided to make it a property, but I'm open to suggestions.

### DIFF
--- a/pyiron_atomistics/atomistics/structure/neighbors.py
+++ b/pyiron_atomistics/atomistics/structure/neighbors.py
@@ -571,9 +571,7 @@ class Neighbors(Tree):
 
     @property
     def chemical_symbols(self):
-        """
-            Returns chemical symbols of the neighboring atoms.
-        """
+        """ Returns chemical symbols of the neighboring atoms.  """
         return self._ref_structure.get_chemical_symbols()[self.indices]
 
     @property
--- a/pyiron_atomistics/atomistics/structure/neighbors.py
+++ b/pyiron_atomistics/atomistics/structure/neighbors.py
@@ -570,6 +570,13 @@ class Neighbors(Tree):
         return to_return.replace('given positions', 'each atom')
 
     @property
+    def chemical_symbols(self):
+        """
+            Returns chemical symbols of the neighboring atoms.
+        """
+        return self._ref_structure.get_chemical_symbols()[self.indices]
+
+    @property
     def shells(self):
         """
             Returns the cell numbers of each atom according to the distances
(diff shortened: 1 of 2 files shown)


pmrv (approved):
lgtm

liamhuber:
> I was not sure if I should call it `neigh.get_chemical_symbols()` or make it a property. Since there's no argument to add, I decided to make it a property, but I'm open to suggestions.

I like this choice. My rationale is this:
- Requires arguments --> Method (obviously 😛)
- Alters parent object in place --> Method
- Best described by a verb --> ...Method? Usually at least. But in this case it typically alters the parent object in place.
- Else --> Property. ❤️ properties.

pyiron_atomistics/atomistics/structure/neighbors.py
```diff
@@ -569,6 +569,13 @@ def __repr__(self):
         to_return = super().__repr__()
         return to_return.replace('given positions', 'each atom')
 
+    @property
+    def chemical_symbols(self):
+        """
+            Returns chemical symbols of the neighboring atoms.
```
liamhuber: codacy is going to whine that this could be on one line. I don't actually care strongly about this formatting choice, but I do like making codacy shut up 😉

tests/atomistics/structure/test_neighbors.py
```diff
@@ -339,6 +339,13 @@ def test_norm_order(self):
         with self.assertRaises(ValueError):
             neigh.norm_order = 3
 
+    def test_chemical_symbols(self):
```
liamhuber, line 343: :heart: tests

tests/atomistics/structure/test_neighbors.py
```diff
@@ -339,6 +339,13 @@ def test_norm_order(self):
...
+        basis = CrystalStructure("Fe", bravais_basis="bcc", lattice_constants=1)
```
liamhuber: I see that `CrystalStructure` just invokes `_CrystalStructure` and that `_CrystalStructure` claims it exists only for historical reasons. I suggest but don't insist that we use `StructureFactory().bulk("Fe", a=1, cubic=True)` instead.

liamhuber (approved):
Cool, cool cool cool.

Change suggestions in the review, but nothing barring merging IMO, just see what you like.

pyiron_atomistics/atomistics/structure/neighbors.py
samwaseda: Oh ok I didn't know that.

PR #117, opened 2021-03-16
User defined tags are added to `__dir__`

### LINKED ISSUE #116: selective_dynamics doesn't appear in the list of tab completion
From what I know even after setting:

```python
structure.add_tag(selective_dynamics=[True, True, True])
```

`selective_dynamics` does not appear in the tab completion list, even though it can (or maybe even *should*?) still be accessed via `structure.selective_dynamics`. I wouldn't call it a bug but it would be really nice to have it tab-completable.

sudarsan-surendralal:
These tags can now be accessed using auto-complete


samwaseda (approved):
Very good!

nit: According to @liamhuber, `super()` doesn't require arguments anymore as long as the direct parent class is meant.

PR #118, opened 2021-03-16
Drop support for old python < 3.6

pmrv:
We do this already for all other repos, I guess this one just slipped through.


jan-janssen (approved):
Looks good to me

PR #119, opened 2021-03-16
Add workflow for integration notebooks

pmrv:
Add the for interactive output as first example.

This is on PR for now to test it, will change it back to on pull only if it's working.


pmrv:
Errors are coming from broken depends from `sqlalchemy`, hope that #122 catches this.

jan-janssen (approved):
Looks good to me

PR #122, opened 2021-03-17
Check pip status using pip check

jan-janssen:


jan-janssen:
The issue seems to be related to https://github.com/conda-forge/sqlalchemy-feedstock/pull/64

PR #136, opened 2021-03-22
Run integration notebooks only on tag

pmrv:
As discussed [here](https://github.com/pyiron/pyiron_base/pull/237).


jan-janssen (approved):
Looks good to me

PR #138, opened 2021-03-23
Quasi newton

samwaseda:
Quasi Newton as implemented in the Hackathon session.

DocStrings and tests will follow.


jan-janssen:
Should this be part of `pyiron_atomistics` or `pyiron_contrib` ?

samwaseda:
> Should this be part of `pyiron_atomistics` or `pyiron_contrib` ?

I had to think about it as well. Here are my arguments for `'pyiron_atomistics`:

- We should have a base Quasi Newton algorithm in pyiron, which can be flexibly extended
- There's probably no functionality in this class that is going to be **removed**, meaning even if someone starts using it, it's unlikely that things won't work in the future
- Moving a class from `pyiron_contrib` to `pyiron_atomistics` will require an extra step some time in the future.

But these are weak arguments. If you find a decisive argument I can also move it to `pyiron_contrib`.

jan-janssen:
> But these are weak arguments. If you find a decisive argument I can also move it to `pyiron_contrib`.

I am open for both, we should just make sure that we come up with some more or less general rules, so that we do not have to decide on a case by case bases. My understanding was that everything which is not directly required by another part of `pyiron_atomistics` should first go into `pyiron_contrib` and only when there are multiple users or use cases for a given feature that it is moved to `pyiron_atomistics` afterwards - but that is just my personal opinion.

samwaseda:
> everything which is not directly required by another part of `pyiron_atomistics` should first go into `pyiron_contrib` and only when there are multiple users or use cases for a given feature that it is moved to `pyiron_atomistics` afterwards - but that is just my personal opinion.

Yeah that's a good point, but the second part is already fulfilled, since @mostafajoulaian is going to use the class according to @freyso . And anyone who wants to do a saddle point calculation will probably want to give this one a try first, before going to ART or NEB, because this one requires hardly any parameter.

liamhuber:
> we should just make sure that we come up with some more or less general rules, so that we do not have to decide on a case by case bases.

👍 

- Good (or even perfect 😉) unit test coverage
  - Here: None. (No blame -- it's only "draft" and they were mentioned! It's just on the TODO list)
- Uses up-to-date syntax
  - Here: `QuasiNewtonOutput`, `Input ` --> `DataContainer` 
- Written in a maintainable way, i.e. maintainable three years later when everyone who wrote it is gone.
  - This one is really hard to get perfect and obviously subjective, but we can try
  - Here, subjective e.g.: Refactor out `run_static` so it is much shorter, just calls a bunch of well-named method calls so that the flow of the algorithm can be immediately understood. Similarly `get_positions` and `update_hessian`.
- Verified demand
  - Here: satisfied, as discussed.

samwaseda:
I’m starting to feel that this class is not quite as general as I thought so I’m gonna move it to pyiron_mpie until someone starts using it.

PR #141, opened 2021-03-24
Add 'msd' flag to lammps calc_md to get mean squared displacements directly from lammps

raynol-dsouza:
I usually manually calculate msd from the total displacements. Getting msd directly from lammps seemed more efficient. It is not super important to merge this PR, though.

### DIFF
--- a/pyiron_atomistics/atomistics/job/atomistic.py
+++ b/pyiron_atomistics/atomistics/job/atomistic.py
@@ -244,6 +244,7 @@ class AtomisticGenericJob(GenericJobCore):
         tloop=None,
         initial_temperature=True,
         langevin=False,
+        msd=False,
     ):
         self._generic_input["calc_mode"] = "md"
         self._generic_input["temperature"] = temperature
@@ -891,3 +892,15 @@ class GenericOutput(object):
             return hdf5_path.list_nodes()
         else:
             return []
+
+    @property
+    def mean_squared_displacement(self):
+        """
+        The mean-squared displacement computed directly within Lammps.
+
+        WARNINGS:
+            1. This works only if the job is a `Lammps` job.
+            2. Returns an output ONLY when the `msd` flag in set to True in `calc_md` for a Lammps job.
+            3. Returns None if used with `calc_static`, `calc_minimize` and `calc_vcsgc` for a Lammps job.
+        """
+        return self._job["output/generic/mean_squared_displacement"]
--- a/pyiron_atomistics/lammps/base.py
+++ b/pyiron_atomistics/lammps/base.py
@@ -603,6 +603,9 @@ class LammpsBase(AtomisticGenericJob):
                 )
                 df["mean_pressures"] = pressures.tolist()
 
+            if "v_msd" in df:
+                df["mean_squared_displacement"] = df.pop("v_msd")
+
             with self.project_hdf5.open("output/generic") as hdf_output:
                 # This is a hack for backward comparability
                 for k, v in df.items():
@@ -671,6 +674,7 @@ class LammpsBase(AtomisticGenericJob):
         langevin=False,
         delta_temp=None,
         delta_press=None,
+        msd=False,
     ):
         # Docstring set programmatically -- Ensure that changes to signature or defaults stay consistent!
         if self.server.run_mode.interactive_non_modal:
@@ -690,6 +694,7 @@ class LammpsBase(AtomisticGenericJob):
             tloop=tloop,
             initial_temperature=initial_temperature,
             langevin=langevin,
+            msd=msd,
         )
         self.input.control.calc_md(
             temperature=temperature,
@@ -706,7 +711,8 @@ class LammpsBase(AtomisticGenericJob):
             delta_temp=delta_temp,
             delta_press=delta_press,
             job_name=self.job_name,
-            rotation_matrix=rotation_matrix
+            rotation_matrix=rotation_matrix,
+            msd=msd
         )
     calc_md.__doc__ = LammpsControl.calc_md.__doc__
 
--- a/pyiron_atomistics/lammps/control.py
+++ b/pyiron_atomistics/lammps/control.py
@@ -427,7 +427,8 @@ class LammpsControl(GenericParameters):
         delta_temp=None,
         delta_press=None,
         job_name="",
-        rotation_matrix=None
+        rotation_matrix=None,
+        msd=False
     ):
         """
         Set an MD calculation within LAMMPS. Nosé Hoover is used by default.
@@ -471,6 +472,8 @@ class LammpsControl(GenericParameters):
             job_name (str): Job name of the job to generate a unique random seed.
             rotation_matrix (numpy.ndarray): The rotation matrix from the pyiron to Lammps coordinate
                 frame.
+            msd (bool): (True or False) Calculate the mean-squared displacement over all the atoms in the structure
+                for a Lammps job
         """
         if self["units"] not in LAMMPS_UNIT_CONVERSIONS.keys():
             raise NotImplementedError
@@ -605,6 +608,14 @@ class LammpsControl(GenericParameters):
                 seed=seed,
                 gaussian=True,
                 job_name=job_name
+            )
+
+        if msd:
+            self.modify(
+                compute___1="all msd com yes",
+                variable___msd=" equal c_1[4]",
+                thermo_style=self.get("thermo_style") + " v_msd",
+                append_if_not_present=True,
             )
 
     def calc_vcsgc(
--- a/tests/lammps/test_base.py
+++ b/tests/lammps/test_base.py
@@ -619,6 +619,13 @@ class TestLammps(unittest.TestCase):
         self.assertTrue(np.isclose(float(m.group(7)), 0.0 * cnv))
         self.assertTrue(np.isclose(float(m.group(8)), 0.0 * cnv))
 
+        # test if msd flag works
+        self.md_control_job.calc_md(temperature=300., msd=True)
+        self.assertEqual(self.md_control_job.input.control["compute___1"], "all msd com yes")
+        self.assertEqual(self.md_control_job.input.control["variable___msd"], " equal c_1[4]")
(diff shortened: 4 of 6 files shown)


samwaseda:
Can you maybe give a specific example where it's indeed more efficient? At least we keep the unwrapped positions in pyiron so the total MSD should be exactly the same.

raynol-dsouza:
I agree, the total MSD is the same. 

For longer MD runs with < 5k snapshots, for example, I believe getting them directly from lammps may be a BIT more efficient. I can write a notebook to test it out.

samwaseda:
> I agree, the total MSD is the same.
> 
> For longer MD runs with < 5k snapshots, for example, I believe getting them directly from lammps may be a BIT more efficient. I can write a notebook to test it out.

I would be a bit surprised if it makes a huge difference though, since MSD is not quite a computationally expensive calculation.

raynol-dsouza:
Right, so getting the MSD from the unwrapped positions takes roughly about 550ms, for a structure with 500 atoms, with 10k snapshots.

samwaseda:
> Right, so getting the MSD from the unwrapped positions takes roughly about 550ms, for a structure with 500 atoms, with 10k snapshots.

Ok but is that too slow?

In principle I have nothing against making LAMMPS-internal MSD available. The problem with pyiron, however, is that for the LAMMPS-internal commands there's hardly any structure. It looks like a totally unconnected patchwork of different commands, some of which are probably even totally redundant. So even though this is not a problem that's created in this PR, I might be slightly concerned about making the LAMMPS class even more difficult if we keep adding stuff in the current format.

pyiron_atomistics/atomistics/job/atomistic.py
```diff
@@ -891,3 +892,15 @@ def __dir__(self):
...
+
+    @property
+    def mean_squared_displacement(self):
+        """
+        The mean-squared displacement computed directly within Lammps.
+
+        WARNINGS:
+            1. This works only if the job is a `Lammps` job.
+            2. Returns an output ONLY when the `msd` flag in set to True in `calc_md` for a Lammps job.
+            3. Returns None if used with `calc_static`, `calc_minimize` and `calc_vcsgc` for a Lammps job.
```
liamhuber, line 904: I share @samwaseda's concerns about cluttering the Lammps codebase when this feature is so specific. What is the exact use-case you have in mind (I think I know -- FTS?) and what is the cost savings compared to post-processing in python from unwrapped positions?

pyiron_atomistics/atomistics/job/atomistic.py
```diff
@@ -891,3 +892,15 @@ def __dir__(self):
...
+        """
+        return self._job["output/generic/mean_squared_displacement"]
```
liamhuber, line 906: Do we need to reference the HDF path itself? My understanding is that job should be a live object and we can use the real output path `job.output.mean_squared_displacement`. 

As much as possible we need to break free from strings and explicit dependence on storage implementation details.

pyiron_atomistics/lammps/base.py
```diff
@@ -671,6 +674,7 @@ def calc_md(
         langevin=False,
         delta_temp=None,
         delta_press=None,
+        msd=False,
```
liamhuber, line 677: `calc_vcsgc` is anyhow invoking `calc_md` -- if we do go through with this why *shouldn't* it work for vcasgc?

pyiron_atomistics/lammps/control.py
```diff
@@ -607,6 +610,14 @@ def calc_md(
                 job_name=job_name
             )
 
+        if msd:
+            self.modify(
+                compute___1="all msd com yes",
```
liamhuber, line 615: `compute___1` seems like a super dangerous name likely to lead to conflict. Rather `compute___msd`?

tests/lammps/test_base.py
```diff
@@ -619,6 +619,13 @@ def test_calc_md_input(self):
         self.assertTrue(np.isclose(float(m.group(7)), 0.0 * cnv))
         self.assertTrue(np.isclose(float(m.group(8)), 0.0 * cnv))
 
+        # test if msd flag works
+        self.md_control_job.calc_md(temperature=300., msd=True)
+        self.assertEqual(self.md_control_job.input.control["compute___1"], "all msd com yes")
+        self.assertEqual(self.md_control_job.input.control["variable___msd"], " equal c_1[4]")
+        self.assertEqual(self.md_control_job.input.control["thermo_style"],
+                         "custom step temp pe etotal pxx pxy pxz pyy pyz pzz vol v_msd")
+
```
liamhuber, line 628: A separate integration test would be nice too.

liamhuber (commented):
@raynol-dsouza I'm not sure I understand, is this for getting MSD(t = {t0, t1, ..., t_n), not just MSD(t=t_n), right? If it's just the last frame then we should definitely only do this in python with the unwrapped positions.

Could you anyhow take a look at the first review comment and give us a sense how much time Lammps would actually save?

pyiron_atomistics/atomistics/job/atomistic.py
raynol-dsouza: > Right, so getting the MSD from the unwrapped positions takes roughly about 550ms, for a structure with 500 atoms, with 10k snapshots.

As I mentioned in a previous comment, the cost for smaller structures is not worth fretting over. I do agree with Sam's and your concerns over cluttering the Lammps codebase with something that can be done outside of Lammps. 

The use case I had in mind was definitely to calculate diffusion coefficients, as many Lammps users are more familiar with the `compute msd` command in the script. I also wanted to calculate approximate spring constants from the msd as mentioned in [this paper](https://doi.org/10.1016/j.commatsci.2015.10.050).

pyiron_atomistics/atomistics/job/atomistic.py
raynol-dsouza: All the `job.output.<something>` functions reference the HDF path.

pyiron_atomistics/lammps/base.py
raynol-dsouza: It can for sure. I only need to add the flag in `calc_vcsgc`. For vcsgc, however, the atom species are different, and I thought calculating the msd for all the atoms as a single scalar would not be accurate.

pyiron_atomistics/atomistics/job/atomistic.py
liamhuber: Suppose we change the storage path. If you explicitly reference the path here, we need to update this code *and* `job.output.<something>`. If you directly reference `job.output.<something>` we only need to change it there. Plus strings are a huge pain in the butt to find and update 😛

raynol-dsouza: Yes, yes. But this seems to be the case for all outputs, not just `mean_squared_displacement`.

For eg. `job.output.energy_pot `returns `self._job["output/generic/energy_pot"]`

liamhuber: Right, my bad. I see now you're extending `GenericOutput` not the job itself.

Basically, IMO, this code is a hot mess where we have our storage path implementation hard-coded everywhere and transitioning over to using `DataContainer` is going to be a huge headache. That's not your fault, but it means that I don't want to make that eventual transition worse by piling on more baggage...

What *does* need fixing in this PR is that the output field is defined in the wrong place. I didn't appreciate this on the first readthrough, but having it live on the `Generic` level and then explicitly warn that it only works for Lammps jobs is a no-go. 
  The simples solution is to go over to `pyiron_atomistics/lammps/base.py` and introduce a new `class _LammpsOutput(GenericOutput)` which extends functionality with this new attribute. I would also then wrap the return in a try/except (not sure the exact error when the HDF field doesn't exist) and return `None` when it's not there, maybe raising a warning then that you had to ask for it. Then your docstring doesn't need all these disclaimers. Don't forget to overwrite `self.output = _LammpsOutput(job=self)` in `LammpsBase.__init__`.

pyiron_atomistics/lammps/base.py
liamhuber: Yeah, I don't know to what degree it's physically meaningful, but I have  a mild preference to maintain that the `calc_md` signature is a perfect subset of the `calc_vcsgc` signature.

raynol-dsouza:
Considering all of the comments, I think it may not really be beneficial to have this PR merged, as it would add more complications to the Lammps codebase within pyiron. Furthermore, what this PR aims to do can also be done at negligible computational cost from the existing output (`unwrapped_positions`) from pyiron's Lammps job. 

If its okay by the reviewers, I would prefer to close this PR without merging it.

samwaseda:
> Considering all of the comments, I think it may not really be beneficial to have this PR merged, as it would add more complications to the Lammps codebase within pyiron. Furthermore, what this PR aims to do can also be done at negligible computational cost from the existing output (`unwrapped_positions`) from pyiron's Lammps job.

I just want to make sure that you know I find it very nice that you opened this PR. Especially in times of working from home we tend to have little communication and I appreciate anything that allows me to see what other people do/think. So I'd like to thank you for bringing up the topic and the discussion, and I'd still encourage you to do the same if similar issues arise.

So I agree that we can close the PR for now, but it made us also see the urgency to making LAMMPS input more flexible. Maybe we can bring it up at the next pyiron meeting.

raynol-dsouza:
Hahaa! Not all PRs need to be merged. :P Besides, it gave me the incentive of having to review the `pyiron_atomistics` code and learn a lot more about how things work under the hood. 

I will bring up the issue of having more flexible Lammps input/output at tomorrow’s pyiron meeting.

liamhuber:
> I just want to make sure that you know I find it very nice that you opened this PR.

I also wanted to quickly make sure it wasn't too demoralizing of a process, so I'm glad Sam beat me to it! 

> Not all PRs need to be merged.

Yeah, at the end of the day this is spot-on. I would also echo your guys' sentiments that the conversation was helpful for me to better understand the issue and how we should keep development going overall!

PR #148, opened 2021-03-29
Add arguments in group_points_by_symmetry which are available in get_…

samwaseda:
…symmetry


samwaseda:
I merge it because it's a super minor issue.

PR #161, opened 2021-04-19
Document energy_pot_per_atom

pmrv:
I'm not sure if the HDF output path should be "hard-coded" here, but I do think it's the crucial information needed here.

In general, I'd like that function to be more visible, but am a bit unsure where place it, any thoughts?

### DIFF
--- a/pyiron_atomistics/lammps/control.py
+++ b/pyiron_atomistics/lammps/control.py
@@ -793,6 +793,9 @@ class LammpsControl(GenericParameters):
             raise NotImplementedError(key+' is not implemented')
 
     def energy_pot_per_atom(self):
+        """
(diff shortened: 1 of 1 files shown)


liamhuber (approved):
Nice touch-up, I'm happy to merge.

> I'm not sure if the HDF output path should be "hard-coded" here, but I do think it's the crucial information needed here.
>
> In general, I'd like that function to be more visible, but am a bit unsure where place it, any thoughts?

Yeah, I absolutely agree. I think the medium-term solution is to restructure job I/O with specially subclassed`DataContainer`s. This involves going all the way back to `GenericJob` though, which currently still fills `input` with a `GenericInput`. Right now hard-coding HDF paths is ubiquitous throughout the code, so I'm not worried about trying to solve it in this one place right away.

PR #164, opened 2021-04-22
Fix vasp import

### LINKED ISSUE #160: Importing unfinished POSCAR from VASP
Importing unfinished POSCAR from VASP (after several steps of ion relaxation) results in empty HDF5 file.

sudarsan-surendralal:
These changes are to ensure importing VASP calculations work even if the POTCAR file is corrupt


samwaseda:
Why does it try to get the species from POTCAR in the first place, if POSCAR already has it?

sudarsan-surendralal:
> 
> 
> Why does it try to get the species from POTCAR in the first place, if POSCAR already has it?

POSCAR files don't always have this. (the older versions don't have the species info)

pmrv (approved):
I might want to rename the directory from "another_glitch" to something more meaningful like "invalid_potcar" or so, unless this is obvious to VASP users.

samwaseda:
> 
> POSCAR files don't always have this. (the older versions don't have the species info)

ok but from what I can see from `atoms_from_string()` in [`structure.py`](https://github.com/pyiron/pyiron_atomistics/blob/master/pyiron_atomistics/vasp/structure.py), it doesn't really do anything about the species when there's not species found in POSCAR (I might be completely wrong because it was just a quick glance). Isn't it better to issue a warning either when pyiron finds no POTCAR or the species couldn't be specified? My preference would be a short warning `no POTCAR found`. (I don't have a strong feeling for this so feel free to ignore it if you don't agree)

samwaseda (approved):

sudarsan-surendralal:
> 
> 
> > POSCAR files don't always have this. (the older versions don't have the species info)
> 
> ok but from what I can see from `atoms_from_string()` in [`structure.py`](https://github.com/pyiron/pyiron_atomistics/blob/master/pyiron_atomistics/vasp/structure.py), it doesn't really do anything about the species when there's not species found in POSCAR (I might be completely wrong because it was just a quick glance). Isn't it better to issue a warning either when pyiron finds no POTCAR or the species couldn't be specified? My preference would be a short warning `no POTCAR found`. (I don't have a strong feeling for this so feel free to ignore it if you don't agree)

Actually, it raises an error rather than a warning when it can't find anything. Which seems reasonable to me

samwaseda:
> Actually, it raises an error rather than a warning when it can't find anything. Which seems reasonable to me

Where is the line which does it?

sudarsan-surendralal:
> 
> 
> > Actually, it raises an error rather than a warning when it can't find anything. Which seems reasonable to me
> 
> Where is the line which does it?

python raises the error if it can't find anything in this line: https://github.com/pyiron/pyiron_atomistics/blob/025a46ea9e0f87a0de824b37841e3629987365ed/pyiron_atomistics/vasp/structure.py#L69

samwaseda:
Sorry I'm starting to be confused. So this PR fixes the case when POTCAR is present, but it's corrupted, right? Shouldn't it tell the user that pyiron failed to read the POTCAR in this case? (Again, I'm gladly overruled if you guys disagree)

And sorry I said *pyiron finds no POTCAR* but what I meant was *pyiron fails to understand the species in POTCAR*.

sudarsan-surendralal:
> Sorry I'm starting to be confused. So this PR fixes the case when POTCAR is present, but it's corrupted, right? Shouldn't it tell the user that pyiron failed to read the POTCAR in this case? (Again, I'm gladly overruled if you guys disagree)

Ok, I misunderstood. It should now raise a warning when this happens!

gongyilun:
possible to merge this fix to pyiron as well?

PR #165, opened 2021-04-22
Grain thickness

liamhuber:
Add kwargs to the aimsgb binding so we can specify the thickness of grains (in terms of the unit cell provided). 

Note that structural repetitions *inside* the GB plane are already trivially easy with the `repeat` call to structures. It's just the thickness that was not possible.


liamhuber:
Ugh, and of course it raises the un-solvable codacy nit...

usaikia (approved):

pmrv (approved):
Looks like you've beat me to it! ;)

PR #167, opened 2021-04-23
Extract get_structure to abstract Mixin

pmrv:
A number of our classes define a `get_structure` with consistent semantics, but there was no one place for its definition.  I pulled it out into an ABC and added two other methods 

```python
number_of_structures
iter_structures
```

where the first gives the maximum "iteration_step" that can be passed to `get_structure` and the other just iterates over all structures.   Previously you would (and our code indeed does) check the size of the cells/positions array in the output of an atomistic job, but that's not so OOP.

On the implementation side, I've added extra methods

```python
_number_of_structures_impl
_get_structure_imp
```

that sub classes should override instead of the ones above, because that guarantees stable interfaces and correct docstrings.

Rationale: I want to have a common and guaranteed interface of obtaining structures from objects, for use in the structure/training container.  Both when adding structures to the container where something like

```python
container.include(job)
```

would be nice to have, but also when adding jobs to potential fitting jobs, I want structure/training containers to behave as the other atomistic jobs.

WIP: renaming all the previously defined `get_structures` to `_get_structure_impl` and deriving from `HasStructure`.

Christmas Wishlist: We also have a `structure` property, that I'd like to see on this class, but I thought touching `get_structure` is intrusive enough for now.

### DIFF
--- a/pyiron_atomistics/atomistics/job/atomistic.py
+++ b/pyiron_atomistics/atomistics/job/atomistic.py
@@ -616,18 +616,7 @@ class AtomisticGenericJob(GenericJobCore, HasStructure):
         """
         return self.get_structure(iteration_step=-1)
 
-    def get_structure(self, iteration_step=-1, wrap_atoms=True):
-        """
-        Gets the structure from a given iteration step of the simulation (MD/ionic relaxation). For static calculations
-        there is only one ionic iteration step
-
-        Args:
-            iteration_step (int): Step for which the structure is requested
-            wrap_atoms (bool): True if the atoms are to be wrapped back into the unit cell
-
-        Returns:
-            pyiron.atomistics.structure.atoms.Atoms: The required structure
-        """
+    def _get_structure_impl(self, iteration_step=-1, wrap_atoms=True):
         if self.structure is None:
             raise AssertionError('Structure not set')
         snapshot = self.structure.copy()
@@ -636,7 +625,7 @@ class AtomisticGenericJob(GenericJobCore, HasStructure):
                 snapshot.cell = self.output.cells[iteration_step]
             except IndexError:
                 if wrap_atoms:
-                    raise IndexError('cell at step ', iteration_step, ' not found')
+                    raise IndexError('cell at step ', iteration_step, ' not found') from None
                 snapshot.cell = None
         if self.output.indices is not None:
             try:
@@ -653,9 +642,8 @@ class AtomisticGenericJob(GenericJobCore, HasStructure):
                 snapshot.positions += self.output.total_displacements[iteration_step]
         return snapshot
 
-    def get_number_of_structures(self):
+    def _number_of_structures_impl(self):
         return self.output.positions.shape[0]
-    get_number_of_structures.__doc__ = HasStructure.get_number_of_structures.__doc__
 
     def map(self, function, parameter_lst):
         master = self.create_job(
--- a/pyiron_atomistics/atomistics/job/atomistic.py
+++ b/pyiron_atomistics/atomistics/job/atomistic.py
@@ -616,30 +616,30 @@ class AtomisticGenericJob(GenericJobCore, HasStructure):
         """
         return self.get_structure(iteration_step=-1)
 
-    def _get_structure(self, iteration_step=-1, wrap_atoms=True):
+    def _get_structure(self, frame=-1, wrap_atoms=True):
         if self.structure is None:
             raise AssertionError('Structure not set')
         snapshot = self.structure.copy()
         if self.output.cells is not None:
             try:
-                snapshot.cell = self.output.cells[iteration_step]
+                snapshot.cell = self.output.cells[frame]
             except IndexError:
                 if wrap_atoms:
-                    raise IndexError('cell at step ', iteration_step, ' not found') from None
+                    raise IndexError('cell at step ', frame, ' not found') from None
                 snapshot.cell = None
         if self.output.indices is not None:
             try:
-                snapshot.indices = self.output.indices[iteration_step]
+                snapshot.indices = self.output.indices[frame]
             except IndexError:
                 pass
         if self.output.positions is not None:
             if wrap_atoms:
-                snapshot.positions = self.output.positions[iteration_step]
+                snapshot.positions = self.output.positions[frame]
                 snapshot.center_coordinates_in_unit_cell()
-            elif len(self.output.unwrapped_positions) > max([iteration_step, 0]):
-                snapshot.positions = self.output.unwrapped_positions[iteration_step]
+            elif len(self.output.unwrapped_positions) > max([frame, 0]):
+                snapshot.positions = self.output.unwrapped_positions[frame]
             else:
-                snapshot.positions += self.output.total_displacements[iteration_step]
+                snapshot.positions += self.output.total_displacements[frame]
         return snapshot
 
     def _number_of_structures(self):
--- a/pyiron_atomistics/atomistics/job/atomistic.py
+++ b/pyiron_atomistics/atomistics/job/atomistic.py
@@ -616,7 +616,7 @@ class AtomisticGenericJob(GenericJobCore, HasStructure):
         """
         return self.get_structure(iteration_step=-1)
 
-    def _get_structure_impl(self, iteration_step=-1, wrap_atoms=True):
+    def _get_structure(self, iteration_step=-1, wrap_atoms=True):
         if self.structure is None:
             raise AssertionError('Structure not set')
         snapshot = self.structure.copy()
@@ -642,7 +642,7 @@ class AtomisticGenericJob(GenericJobCore, HasStructure):
                 snapshot.positions += self.output.total_displacements[iteration_step]
         return snapshot
 
-    def _number_of_structures_impl(self):
+    def _number_of_structures(self):
         return self.output.positions.shape[0]
 
     def map(self, function, parameter_lst):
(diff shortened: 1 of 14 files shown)


pyiron_atomistics/atomistics/structure/has_structure.py
```diff
@@ -0,0 +1,108 @@
...
+            wrap_atoms (bool): True if the atoms are to be wrapped back into the unit cell
+
+        Returns:
+            :class:`pyiron_atomistics.atomistics.structure.atoms.Atoms`: the requested structure
+
+        Raises:
+            IndexError: if not 0 < iteration_step < :property:`.number_of_structures`
+        """
+        num_structures = self.number_of_structures
+        if iteration_step < 0:
```
liamhuber: Edge case: Rather take a mod, as this will fail for numbers < -`number_of_structures`.

pyiron_atomistics/atomistics/structure/has_structure.py
```diff
@@ -0,0 +1,108 @@
...
+            iteration_step += num_structures
+        if not (0 <= iteration_step < num_structures):
+            raise IndexError(f"iteration_step {iteration_step} out of range [0, {num_structures}).")
+
+        return self._get_structure_impl(iteration_step=iteration_step, wrap_atoms=wrap_atoms)
+
+    @abstractmethod
+    def _get_structure_impl(self, iteration_step=-1, wrap_atoms=True):
```
liamhuber: I'm not very familiar with mixins... I assume `_impl` is short for "implicit"? Or? Is this standard practice for mixins? I could see rather just `_get_structure` instead (sim. for number of structures).

pyiron_atomistics/atomistics/master/murnaghan.py
```diff
@@ -590,7 +591,7 @@ def pouriertarantola(V, E0, B0, BP, V0):
 
 
 # ToDo: not all abstract methods implemented
-class Murnaghan(AtomisticParallelMaster):
+class Murnaghan(AtomisticParallelMaster, HasStructure):
```
liamhuber: I would really like that `AtomisticParallelMaster` has some `Atomistic` parent that already gives us the `HasStructure` inheritance. Doesn't have to happen in this PR though, and I haven't looked yet what the current tree is.

liamhuber (commented):
I love this idea.

pyiron_atomistics/atomistics/structure/has_structure.py
pmrv: I only want to allow iteration_steps within [-N, N), so I do want this to fail for number < -N.  This is not reflected in the error message, so I'll touch that up, though.

pyiron_atomistics/atomistics/structure/has_structure.py
pmrv: I put it as "implementation", but you're right dropping it, makes it shorter and it's still clearly private, so I'll do that.

pyiron_atomistics/atomistics/master/murnaghan.py
pmrv: I'll check the tree if it is necessary.

pmrv:
The (newly added) murnaghan test case fails, because for the dummy jobs setup in the test suite the calculation does not converge (doesn't find a minimum), so I've checked manually whether it works. It does. I'll try to fiddle with the parameters, but in case it doesn't get better I suggest I'll take the test out again and we merge as is.

pyiron_atomistics/atomistics/job/atomistic.py
```diff
@@ -9,6 +9,7 @@
 import warnings
 
 from pyiron_atomistics.atomistics.structure.atoms import Atoms
+from ..structure.has_structure import HasStructure
```
liamhuber: I don't have a specific complaint about the relative import, but it raises a yellow flag for me since I'm pretty sure we avoid it everywhere else.

pyiron_atomistics/atomistics/job/atomistic.py
```diff
@@ -635,7 +625,7 @@ def get_structure(self, iteration_step=-1, wrap_atoms=True):
                 snapshot.cell = self.output.cells[iteration_step]
             except IndexError:
                 if wrap_atoms:
-                    raise IndexError('cell at step ', iteration_step, ' not found')
+                    raise IndexError('cell at step ', iteration_step, ' not found') from None
```
liamhuber: TIL about `from`ing exceptions 👍 
Not 100% sure how I feel about it, but I don't object at any rate.

pyiron_atomistics/atomistics/master/murnaghan.py
```diff
@@ -11,6 +11,7 @@
 import warnings
 from pyiron_atomistics.atomistics.structure.atoms import Atoms, ase_to_pyiron
 from pyiron_atomistics.atomistics.master.parallel import AtomisticParallelMaster
+from ..structure.has_structure import HasStructure
```
liamhuber: `Murnaghan(AtomisticParallelMaster(ParallelMaster, AtomisticGenericJob(GenericJobCore, HasStructure)))` and `HasStructure` otherwise doesn't appear in `murnaghan.py`. Isn't this import redundant?

pyiron_atomistics/atomistics/master/parallel.py
```diff
@@ -41,11 +41,11 @@ def structure(self, basis):
                 "A structure can only be set after a reference job has been assinged."
             )
 
-    def get_structure(self, iteration_step=-1):
-        if iteration_step == 0:
-            return self.structure
-        else:
-            raise ValueError("iteration_step should be either 0.")
+    def _get_structure(self, iteration_step=0, wrap_atoms=True):
```
liamhuber: I'm confused why `AtomisticParallelMaster` isn't an `ABC`, in which case these could (should?) be `abstractmethod`s. What am I missing?

pyiron_atomistics/atomistics/structure/has_structure.py
```diff
@@ -0,0 +1,116 @@
...
+    Cell([0.0, 0.0, 0.0])
+    <BLANKLINE>
+
+    >>> isinstance(f, HasStructure)
+    True
+    """
+
+    def get_structure(self, iteration_step=-1, wrap_atoms=True):
+        """
+        Gets the structure from a given iteration step of the simulation (MD/ionic relaxation). For static calculations
```
liamhuber: "(MD/ionic relaxation)" is too restrictive. Consider, e.g. Monte Carlo on chemistry. I would change this to "(e.g. MD, ionic relaxation, Monte Carlo, etc...)" or just remove it.

liamhuber: (I lean weakly towards removal)

pyiron_atomistics/lammps/interactive.py
```diff
@@ -51,11 +51,14 @@ def structure(self, structure):
         self._prism = UnfoldingPrism(structure.cell)
         GenericInteractive.structure.fset(self, structure)
 
-    def get_structure(self, iteration_step=-1, wrap_atoms=True):
-        return super(GenericInteractive, self).get_structure(
+    def _get_structure(self, iteration_step=-1, wrap_atoms=True):
```
liamhuber: I checked out the branch and it looks to me like `super()._get_structure` and `_number_of_structures` trace back to the methods defined on `AtomisticGenericJob`; is this correct? If so, why does the `super` call need an explicit parent path, and why are these methods needed at all since they're just naked supers calls?

pyiron_atomistics/sphinx/interactive.py
```diff
@@ -51,8 +51,8 @@ def structure(self, structure):
                 element_lst=structure.get_species_symbols().tolist()
             )
 
-    def get_structure(self, iteration_step=-1, wrap_atoms=True):
-        return GenericInteractive.get_structure(
+    def _get_structure(self, iteration_step=-1, wrap_atoms=True):
```
liamhuber: Yeah, in contrast I definitely see the need for the explicit super call here since the `_get_structure` method differs depending whether you wind up at `GenericDFTJob` or `AtomisticGenericJob`.

liamhuber (commented):
Looking good. There's a couple places I think I found some redundant code, so either you can clean things up a bit or I can learn something new 😉 It's all pretty minor stuff though, and I'm looking forward to having this in the codebase!

pyiron_atomistics/atomistics/job/atomistic.py
pmrv: It's nice I think when you're trying to keep the backtrace clean.  The original IndexError is an implementation detail, so it's nice to hide it.

pyiron_atomistics/atomistics/master/murnaghan.py
pmrv: Yup, forgot to remove that after I dropped `HasStructure` from `Murnaghan`.

pyiron_atomistics/atomistics/master/parallel.py
pmrv: In principle you could instantiate `AtomisticParallelMaster` directly with a custom generator, so it doesn't feel like an `ABC` to me.  But then again I just didn't want to touch too many things and just renamed the methods.

pyiron_atomistics/atomistics/structure/has_structure.py
pmrv: Agree, this also came up [here](https://github.com/pyiron/pyiron_contrib/pull/153).  Maybe it's best to deprecate `iteration_step` and rename it to something more generic together with the docstring.

liamhuber: Yeah, I like `frame` personally

pyiron_atomistics/lammps/interactive.py
pmrv: Back then, I just renamed it blindly.  But I think you're right both inheritance paths go to `AtomisticGenericJob`, so the super calls should be unnecessary.

PR #172, opened 2021-04-23
Handling chemical formula larger than 30 chars

sudarsan-surendralal:
This is an intermediate fix till a proper solution is found.

### DIFF
--- a/pyiron_atomistics/atomistics/job/atomistic.py
+++ b/pyiron_atomistics/atomistics/job/atomistic.py
@@ -394,7 +394,10 @@ class AtomisticGenericJob(GenericJobCore):
                 parent_structure = self.structure.get_parent_basis()
             else:
                 parent_structure = self.structure.copy()
-            db_dict["ChemicalFormula"] = parent_structure.get_chemical_formula()
+            chem_formula = parent_structure.get_chemical_formula()
+            if len(chem_formula) > 30:
+                chem_formula = chem_formula[:29]+"+"
+            db_dict["ChemicalFormula"] = chem_formula
         return db_dict
 
     def restart(self, job_name=None, job_type=None):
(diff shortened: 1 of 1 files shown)


pyiron_atomistics/atomistics/job/atomistic.py
```diff
@@ -394,7 +394,10 @@ def db_entry(self):
                 parent_structure = self.structure.get_parent_basis()
             else:
                 parent_structure = self.structure.copy()
-            db_dict["ChemicalFormula"] = parent_structure.get_chemical_formula()
+            chem_formula = parent_structure.get_chemical_formula()
+            if len(chem_formula) > 30:
+                chem_formula = chem_formula[:29]+"+"
```
liamhuber: My concern with leaving any of the real formula intact is that then certain parsing operations will silently return the wrong results. e.g. imagine that our limits were much shorter and we would up with something like "Fe302Mn1000" -> "Fe302Mn10" (wrong count), or "O1234P56He" -> "O1234P56H" (wrong species).

I would suggest instead a value like "OVERFLOW_ERROR", and let the user exploit pyiron tables to extract the chemistry directly from the stored structures.

liamhuber (changes requested):
Cool, nice and fast! I would suggest only to change the actual stored value to avoid silent errors.

jan-janssen:
I do not think we should make cluster dependent bug fixes, as the issue is already fixed in https://github.com/pyiron/pyiron_base/pull/290 . If you want to modify the version on our cluster that is fine for me but I am against including this workaround in the master branch.

sudarsan-surendralal:
I close this since @muh-hassani has already proposed a better solution in https://github.com/pyiron/pyiron_base/pull/295

PR #179, opened 2021-04-28
Bader analysis in pyiron

### LINKED ISSUE #62: Request for function for doing bader charge analysis
**Summary**

Hello everyone,
I am looking for a function in pyiron to do Bader charge analysis. 

**Detailed Description**

Usually, it can be done with an executable developed by Prof. Henkelman. [(http://theory.cm.utexas.edu/henkelman/code/bader/)](url)

This executable generates a file (ACF.dat) with the charges of the atoms. 
It would be nice to have a function that can run this executable, generate the ACF.dat file and read the charges of the individual atoms from this file.

The best-case scenario would be when the user can access the Bader charges as an attribute of the job.

[(http://theory.cm.utexas.edu/henkelman/code/bader/)](url)

sudarsan-surendralal:
This PR interfaces the Bader charge analysis code http://theory.cm.utexas.edu/henkelman/code/bader/ in pyiron

Two new outputs:

- `job["output/generic/dft/valence_charges"]` the valence charge on each atom
- `job["output/generic/dft/bader_charges"]` the bader charge on each atom (after correcting for the valence charge). Determined if the necessary files for a Bader analysis are present!


sudarsan-surendralal:
I see that `bader` on conda-forge is available only for Linux. Any way around this @muh-hassani @niklassiemer?

niklassiemer:
[bader](http://theory.cm.utexas.edu/henkelman/code/bader/) itself offers also a Mac executable. Thus, it should be possible to add this to the `bader` on conda-forge - and Jan is the only maintainer of this conda-forge package. For Windows support compiling is always more problematic.

jan-janssen:
I added an Mac Os X version https://github.com/conda-forge/bader-feedstock/pull/1 but I have no experience with compiling Fortran on Windows, if somebody wants to give it a try feel free to open a pull request on the conda-forge recipe.

sudarsan-surendralal:
@jan-janssen Thank you! It's probably very difficult (or even impossible) to compile this for windows and anyway, I don't expect people to use this in a windows environment where one can't run DFT calculations. In that case, I would rather put these tests as an integration notebook (which runs only in a Linux environment).  This seems reasonable right?

niklassiemer:
You may also check the os.name and skip if it is Windows in the unittests.

sudarsan-surendralal:
> 
> 
> You may also check the os.name and skip if it is Windows in the unittests.

Sure. But wouldn't we still need a separate environment file (`.ci-support/environment.yml`) for windows where we exclude `bader`? Otherwise the CI would fail when trying to install this

niklassiemer:
Yes, that is true...

pyiron_atomistics/dft/job/bader.py
```diff
@@ -0,0 +1,124 @@
...
+__author__ = "Sudarsan Surendralal"
+__copyright__ = (
+    "Copyright 2021, Max-Planck-Institut für Eisenforschung GmbH - "
+    "Computational Materials Design (CM) Department"
+)
+__version__ = "1.0"
+__maintainer__ = "Sudarsan Surendralal"
+__email__ = "surendralal@mpie.de"
+__status__ = "production"
+__date__ = "Sep 1, 2017"
```
liamhuber: 🔴

pyiron_atomistics/vasp/base.py
```diff
@@ -1584,7 +1619,7 @@ def compress(self, files_to_compress=None):
             files_to_compress = [
                 f
                 for f in list(self.list_files())
-                if f not in ["CHGCAR", "CONTCAR", "WAVECAR", "STOPCAR"]
+                if f not in ["CHGCAR", "CONTCAR", "WAVECAR", "STOPCAR", "AECCAR0", "AECCAR1", 'AECCAR2']
```
liamhuber: Nit: quotation consistency

pyiron_atomistics/dft/job/bader.py
```diff
@@ -0,0 +1,124 @@
...
+        Initialize the Bader module
+
+        Args:
+            job (pyiron_atomistics.dft.job.generic.GenericDFTJob): A DFT job instance (finished/converged job)
+        """
+        self.job = job
+        self._working_directory = job.working_directory
+        self._structure = job.structure
+
+    def create_cube_files(self):
```
liamhuber: can be private?

pyiron_atomistics/dft/job/bader.py
```diff
@@ -0,0 +1,124 @@
...
+        """
+        self.create_cube_files()
+        error_code = call_bader(foldername=self._working_directory, extra_arguments=extra_arguments)
+        if error_code > 0:
+            self.remove_cube_files()
+            raise ValueError("Invoking Bader charge analysis failed!")
+        self.remove_cube_files()
+        return self.parse_charge_vol()
+
+    def remove_cube_files(self):
```
liamhuber: can be private?

pyiron_atomistics/dft/job/bader.py
```diff
@@ -0,0 +1,124 @@
...
+        """
+        Delete created CUBE files
+        """
+        os.remove(os.path.join(self._working_directory, "valence_charge.CUBE"))
+        os.remove(os.path.join(self._working_directory, "total_charge.CUBE"))
+
+    def parse_charge_vol(self):
```
liamhuber: can be private?

pyiron_atomistics/dft/job/bader.py
```diff
@@ -0,0 +1,124 @@
...
+
+        Returns:
+            tuple: charges and volumes
+
+        """
+        filename = os.path.join(self._working_directory, "ACF.dat")
+        return parse_charge_vol_file(structure=self._structure, filename=filename)
+
+
+def call_bader(foldername, extra_arguments=None):
```
liamhuber: can be private?

pyiron_atomistics/dft/job/bader.py
```diff
@@ -0,0 +1,124 @@
...
+    cmd_2 = list()
+    cmd_2.append("bader")
+    cmd_2.append("{0}/valence_charge.CUBE -ref {0}/total_charge.CUBE".format(foldername))
+    if extra_arguments is not None:
+        cmd_2.append(extra_arguments)
+    cmd_2 = " ".join(cmd_2)
+    return subprocess.call(";".join([cmd_1, cmd_2]), shell=True)
+
+
+def parse_charge_vol_file(structure, filename="ACF.dat"):
```
liamhuber: can be private?

pyiron_atomistics/dft/job/bader.py
```diff
@@ -0,0 +1,124 @@
...
+
+    def create_cube_files(self):
+        """
+        Create CUBE format files of the total and valce charges to be used by the Bader program
+        """
+        cd_val, cd_total = self.job.get_valence_and_total_charge_density()
+        cd_val.write_cube_file(filename=os.path.join(self._working_directory, "valence_charge.CUBE"))
+        cd_total.write_cube_file(filename=os.path.join(self._working_directory, "total_charge.CUBE"))
+
+    def call_bader_from_job(self, extra_arguments=None):
```
liamhuber: As far as I can tell, this is the only public interface with this functionality. Instead of `Bader.call_bader_from_job`, what about just `Bader.run` or `Bader.calculate`?

pyiron_atomistics/dft/job/generic.py
```diff
@@ -243,6 +243,11 @@ def get_kpoints(self):
             "The get_kpoints() function is not implemented for this code."
         )
 
+    def get_valence_and_total_charge_density(self):
```
liamhuber, line 246: If this is only implemented for VASP, is there a reason (other than optimism that it will be implemented for other codes later) that this lives in `GenericDFT` instead of `VASP`? (And of course the same for `get_bader_*`.

liamhuber (commented):
Cool! I'm not super well qualified to evaluate the function, but I'm overall happy with the form. I always get a little uncomfortable with `get` methods instead of properties and with referencing the HDF path directly, but I'm pretty sure this is the standard operating procedure for the DFT code so I don't mind keeping things consistent.

I have a few nits, and it looks like we can make a bunch of stuff private -- but I didn't check out the branch so I might be missing an out-of-scope call. I flagged all the suspects though.

My only big complaint is pretty easy to fix: why does `bader.py` live in `.../dft/job`? It does not inherit from `GenericJob`, so we should put it somewhere else (maybe just `.../dft`).

pyiron_atomistics/dft/job/bader.py
sudarsan-surendralal: Oops!

pyiron_atomistics/dft/job/generic.py
sudarsan-surendralal: This is straightforward to implement all other DFT codes as well. It's just that I'm not familiar with the format of the volumetric data files for these codes. With @samwaseda's help, we can definitely get this done for SPHInX as well.

pyiron_atomistics/dft/job/bader.py
sudarsan-surendralal: I thought about this. But since this module is technical (edit: **not**) a "job", I consciously avoided `run` or `calculate`. Perhaps `compute_bader_charges` is a more suitable name?

tests/vasp/test_vasp_import.py
```diff
@@ -45,10 +45,9 @@ def test_import(self):
         folder_path = os.path.join(
             self.file_location, "../static/vasp_test_files/full_job_minor_glitch"
         )
-        with warnings.catch_warnings(record=True) as w:
+        with warnings.catch_warnings(record=True) as _:
             warnings.simplefilter("always")
             self.project.import_from_path(path=folder_path, recursive=False)
-            self.assertEqual(len(w), 4)
```
jan-janssen: Is it necessary to remove this line? While I agree that counting the number of warnings is not very intuitive it helped to identify the recent HDF5 warnings, so I would prefer to keep it. In any case please make sure to merge the master branch ones as I have the feeling this pull request started before the recent HDF5 version update which might be the reason why you removed this part.

pyiron_atomistics/dft/bader.py
```diff
@@ -0,0 +1,124 @@
...
+    """
+    # Go to the working directory to do this
+    cmd_1 = "cd {}".format(foldername)
+    cmd_2 = list()
+    cmd_2.append("bader")
+    cmd_2.append("{0}/valence_charge.CUBE -ref {0}/total_charge.CUBE".format(foldername))
+    if extra_arguments is not None:
+        cmd_2.append(extra_arguments)
+    cmd_2 = " ".join(cmd_2)
+    return subprocess.call(";".join([cmd_1, cmd_2]), shell=True)
```
jan-janssen: I guess this call can be simplified. On the one hand you can set the current working directory of the subprocess using `cwd` and on the other hand for reading the code it is confusing when lists and strings are combined. Either make each parameter a separate element in a list or build one long string, both is fine but the combination is confusing for me.

pyiron_atomistics/vasp/base.py
```diff
@@ -395,6 +396,22 @@ def collect_output(self):
         except (IOError, ValueError, FileNotFoundError):
             pass
 
+        # Bader analysis
+        if os.path.isfile(self.working_directory + "/AECCAR0") and os.path.isfile(self.working_directory + "/AECCAR2"):
```
jan-janssen: please use `os.path.join` rather than strings.

pyiron_atomistics/vasp/base.py
```diff
@@ -395,6 +396,22 @@ def collect_output(self):
...
+                charges_orig, volumes_orig = bader.call_bader_from_job()
+                charges, volumes = charges_orig.copy(), volumes_orig.copy()
+                charges[self.sorted_indices] = charges_orig
+                volumes[self.sorted_indices] = volumes_orig
+                if "valence_charges" in self._output_parser.generic_output.dft_log_dict.keys():
+                    valence_charges = self._output_parser.generic_output.dft_log_dict["valence_charges"]
+                    # Positive values indicate electron depletion
+                    self._output_parser.generic_output.dft_log_dict["bader_charges"] = valence_charges - charges
+                self._output_parser.generic_output.dft_log_dict["bader_volumes"] = volumes
+            except ValueError:
```
jan-janssen: I have the feeling this `try` and `except` statement is a bit lengthy. Possible errors could be changes in the Bader code which lead to different output files or simply the Bader executable not being available on a given system, but both would result in the same error message making it hard to debug.

pyiron_atomistics/vasp/base.py
```diff
@@ -395,6 +396,22 @@ def collect_output(self):
...
+                warnings.warn("Invoking Bader charge analysis failed")
+                self.logger.warning("Invoking Bader charge analysis failed")
```
niklassiemer: How about 
```suggestion
            try:
                charges_orig, volumes_orig = bader.call_bader_from_job()
            except ValueError:
                warnings.warn("Invoking Bader charge analysis failed")
                self.logger.warning("Invoking Bader charge analysis failed")
            else:
                charges, volumes = charges_orig.copy(), volumes_orig.copy()
                charges[self.sorted_indices] = charges_orig
                volumes[self.sorted_indices] = volumes_orig
                if "valence_charges" in self._output_parser.generic_output.dft_log_dict.keys():
                    valence_charges = self._output_parser.generic_output.dft_log_dict["valence_charges"]
                    # Positive values indicate electron depletion
                    self._output_parser.generic_output.dft_log_dict["bader_charges"] = valence_charges - charges
                self._output_parser.generic_output.dft_log_dict["bader_volumes"] = volumes
```

pyiron_atomistics/dft/job/bader.py
liamhuber: Yeah, I don't object to that name. IMO even `compute` is sufficient given the context that it is the one public method of the `Bader` class. 

> But since this module is technical a "job"...

I'm still confused by this. As per my main review comment, isn't it explicitly *not* a job? In this case I also see a good justification for avoiding `run` (job's have `run` and this isn't a job). But I still understand it to be not-a-job, and don't think we need to specially avoid `calculate`, `compute`, or similar.

tests/vasp/test_vasp_import.py
liamhuber: > While I agree that counting the number of warnings is not very intuitive...

Indeed! If we keep it, it would be nice to have something in the `msg` kwarg explaining what warnings we're expecting to see here.

pyiron_atomistics/dft/job/bader.py
sudarsan-surendralal: > I'm still confused by this. As per my main review comment, isn't it explicitly _not_ a job? In this case I also see a good justification for avoiding `run` (job's have `run` and this isn't a job). But I still understand it to be not-a-job, and don't think we need to specially avoid `calculate`, `compute`, or similar.

Sorry my bad, I meant "technically **not** a job" in my previous comment. I shouldn't post comments after 8 PM

tests/vasp/test_vasp_import.py
sudarsan-surendralal: I removed it precisely because it isn't that intuitive. I like Liam's idea of checking the message but I don't know what to expect and maybe this needs to be done in a separate PR.

pyiron_atomistics/vasp/base.py
sudarsan-surendralal: Thanks @niklassiemer, I've fixed this!

pyiron_atomistics/vasp/base.py
sudarsan-surendralal: Done

pyiron_atomistics/dft/bader.py
sudarsan-surendralal: Done!

pyiron_atomistics/dft/job/bader.py
liamhuber: > I shouldn't post comments after 8 PM

😆 fair! I can relate!

jan-janssen (approved):
Looks good to me

PR #183, opened 2021-04-29
Steinhardt parameter implemented

samwaseda:
I implemented it because it was super easy + I wanted to have Steinhardt parameters of unoccupied interstitial sites.

### DIFF
--- a/pyiron_atomistics/atomistics/structure/neighbors.py
+++ b/pyiron_atomistics/atomistics/structure/neighbors.py
@@ -531,6 +531,8 @@ class Tree:
         Args:
             l (int/numpy.array): Degree of the harmonic (int); must have ``l >= 0``.
             m (int/numpy.array): Order of the harmonic (int); must have ``|m| <= l``.
+            cutoff_radius (float): maximum neighbor distance to include (default: inf, i.e. all
+            atoms included in the neighbor search).
             rotation (numpy.array/list): Rotation to make sure phi does not become nan
 
         Returns:
@@ -564,6 +566,8 @@ class Tree:
         """
         Args:
             l (int/numpy.array): Order of Steinhardt parameter
+            cutoff_radius (float): maximum neighbor distance to include (default: inf, i.e. all
+            atoms included in the neighbor search).
 
         See more on https://pyscal.org/part3/steinhardt.html
 
--- a/pyiron_atomistics/atomistics/structure/neighbors.py
+++ b/pyiron_atomistics/atomistics/structure/neighbors.py
@@ -533,7 +533,7 @@ class Tree:
             m (int/numpy.array): Order of the harmonic (int); must have ``|m| <= l``.
             cutoff_radius (float): maximum neighbor distance to include (default: inf, i.e. all
             atoms included in the neighbor search).
-            rotation (numpy.array/list): Rotation to make sure phi does not become nan
+            rotation ( (3,3) numpy.array/list): Rotation to make sure phi does not become nan
 
         Returns:
             (numpy.array) spherical harmonic values
@@ -542,6 +542,9 @@ class Tree:
 
         Y^m_l(\theta,\phi) = \sqrt{\frac{2l+1}{4\pi} \frac{(l-m)!}{(l+m)!}}
         e^{i m \theta} P^m_l(\cos(\phi))
+
+        The angles are calculated based on `self.vecs`, where the azimuthal angle is defined on the
+        xy-plane and the polar angle is along the z-axis.
 
         See more on: scipy.special.sph_harm
 
--- a/pyiron_atomistics/atomistics/structure/neighbors.py
+++ b/pyiron_atomistics/atomistics/structure/neighbors.py
@@ -548,7 +548,7 @@ class Tree:
         if rotation is not None:
             vecs = np.einsum('ij,nkj->nki', rotation, vecs)
         within_cutoff = self.distances<cutoff_radius
-        if np.any(~np.all(within_cutoff, axis=-1)):
+        if np.any(np.all(~within_cutoff, axis=-1)):
             raise ValueError('cutoff_radius too small - some atoms have no neighbors')
         phi = np.zeros_like(self.distances)
         theta = np.zeros_like(self.distances)
--- a/pyiron_atomistics/atomistics/structure/neighbors.py
+++ b/pyiron_atomistics/atomistics/structure/neighbors.py
@@ -548,6 +548,8 @@ class Tree:
         if rotation is not None:
             vecs = np.einsum('ij,nkj->nki', rotation, vecs)
         within_cutoff = self.distances<cutoff_radius
+        if np.any(~np.all(within_cutoff, axis=-1)):
+            raise ValueError('cutoff_radius too small - some atoms have no neighbors')
         phi = np.zeros_like(self.distances)
         theta = np.zeros_like(self.distances)
         theta[within_cutoff] = np.arctan2(vecs[within_cutoff,1], vecs[within_cutoff,0])
@@ -571,7 +573,7 @@ class Tree:
         random_rotation = Rotation.from_mrp(np.random.random(3)).as_matrix()
         return np.sqrt(4*np.pi/(2*l+1)*np.sum([
             np.absolute(self.get_spherical_harmonics(
-                l=l, m=m, cutoff_radius=cutoff_radius, random_rotation=random_rotation
+                l=l, m=m, cutoff_radius=cutoff_radius, rotation=random_rotation
             ))**2
             for m in np.arange(-l, l+1)
         ], axis=0))
--- a/pyiron_atomistics/atomistics/structure/neighbors.py
+++ b/pyiron_atomistics/atomistics/structure/neighbors.py
@@ -564,6 +564,9 @@ class Tree:
             l (int/numpy.array): Order of Steinhardt parameter
 
         See more on https://pyscal.org/part3/steinhardt.html
+
+        Note: This function does not have an internal algorithm to calculate a suitable cutoff
+        radius. For automated uses, see Atoms.analyse.pyscal_steinhardt_parameter()
         """
         random_rotation = Rotation.from_mrp(np.random.random(3)).as_matrix()
         return np.sqrt(4*np.pi/(2*l+1)*np.sum([
--- a/pyiron_atomistics/atomistics/structure/neighbors.py
+++ b/pyiron_atomistics/atomistics/structure/neighbors.py
@@ -536,7 +536,7 @@ class Tree:
             rotation ( (3,3) numpy.array/list): Rotation to make sure phi does not become nan
 
         Returns:
-            (numpy.array) spherical harmonic values
+            ( (natoms,) numpy.array) spherical harmonic values
 
         Spherical harmonics defined as follows
 
@@ -571,6 +571,9 @@ class Tree:
             l (int/numpy.array): Order of Steinhardt parameter
             cutoff_radius (float): maximum neighbor distance to include (default: inf, i.e. all
             atoms included in the neighbor search).
+
+        Returns:
+            ( (natoms,) numpy.array) Steinhardt parameter values
 
         See more on https://pyscal.org/part3/steinhardt.html
 
--- a/pyiron_atomistics/atomistics/structure/neighbors.py
+++ b/pyiron_atomistics/atomistics/structure/neighbors.py
@@ -526,7 +526,7 @@ class Tree:
                 return True
         return False
 
-    def get_spherical_harmonics(self, l, m, rotation=None):
+    def get_spherical_harmonics(self, l, m, cutoff_radius=np.inf, rotation=None):
         """
         Args:
             l (int/numpy.array): Degree of the harmonic (int); must have ``l >= 0``.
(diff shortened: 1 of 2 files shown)


liamhuber (commented):
lgtm.

For the tests, you might be able to lean heavily on stuff @yuzie007 did for [one of my repos](https://github.com/liamhuber/clustergb/tree/master/tests). At the very least he has the right parameters tabulated for a bunch of structures in the `.dat` file so you have something quickly compare against 😄

samwaseda:
> lgtm.
> 
> For the tests, you might be able to lean heavily on stuff @yuzie007 did for [one of my repos](https://github.com/liamhuber/clustergb/tree/master/tests). At the very least he has the right parameters tabulated for a bunch of structures in the `.dat` file so you have something quickly compare against 😄

I just compared my values, the ones you suggested and those of pyscal's. I got the same values for mine and pyscal but they seem to differ for hcp from your data. I'm gonna use pyscal's values, ok?

liamhuber:
> > lgtm.
> > For the tests, you might be able to lean heavily on stuff @yuzie007 did for [one of my repos](https://github.com/liamhuber/clustergb/tree/master/tests). At the very least he has the right parameters tabulated for a bunch of structures in the `.dat` file so you have something quickly compare against 😄
> 
> I just compared my values, the ones you suggested and those of pyscal's. I got the same values for mine and pyscal but they seem to differ for hcp from your data. I'm gonna use pyscal's values, ok?

Sounds good. Are the HCP values close at least? My guess the c/a ratio differs for the input or something. You can see the exact positions Yuji was using [here](https://github.com/liamhuber/clustergb/blob/46583e09059d4ec90d89c5e24aac4579460445fb/tests/positions.py#L130) and see if they're different from yours. Anyhow, since HCP is the only one differing, I have to imagine it's the input rather than the code that differs so it's not a huge deal IMO.

samwaseda:
> 
> Anyhow, since HCP is the only one differing, I have to imagine it's the input rather than the code that differs so it's not a huge deal IMO.

Sorry I said something super misleading - I compared fcc and hcp (because bcc was missing in Yuji's data) and hcp differed.

liamhuber:
> Sorry I said something super misleading - I compared fcc and hcp (because bcc was missing in Yuji's data) and hcp differed.

I guess BCC is 'cube' in Yuji's notation. All his positions seem to be relative to an atom at [0,0,0]. If your/pyscal's calculations and Yuji's disagree for HCP and the input is the same I guess this warrants looking into. By democracy I guess it's probably Yuji's code that's wrong^1, but in case he's right it's important to fix pyscal!

^1 still way less wrong than my implementation he was correcting, for the record.

samwaseda:
> I guess BCC is 'cube' in Yuji's notation. All his positions seem to be relative to an atom at [0,0,0]. If your/pyscal's calculations and Yuji's disagree for HCP and the input is the same I guess this warrants looking into. By democracy I guess it's probably Yuji's code that's wrong^1, but in case he's right it's important to fix pyscal!

oh ok - indeed for if it's cube I get the same values for bcc. BUT then there's a potentially larger problem: pyscal takes 14 neighbors for bcc while Yuji takes 8, which of course deliver different values. I'm not really sure whether this is something that the user should be aware of.

pyiron_atomistics/atomistics/structure/neighbors.py
```diff
@@ -524,6 +526,58 @@ def _check_width(self, width, pbc):
...
+            np.linalg.norm(vecs[within_cutoff,:2], axis=-1), vecs[within_cutoff,2]
+        )
+        return np.sum(
+            sph_harm(m, l, theta, phi)*within_cutoff, axis=-1
+        )/np.sum(within_cutoff, axis=-1)
+
+    def get_steinhardt_parameter(self, l, cutoff_radius=np.inf):
+        """
+        Args:
+            l (int/numpy.array): Order of Steinhardt parameter
```
pmrv, line 571: Even if we don't expect `cutoff_radius` to be set, it should appear the documented argument list.

liamhuber:
> oh ok - indeed for if it's cube I get the same values for bcc. BUT then there's a potentially larger problem: pyscal takes 14 neighbors for bcc while Yuji takes 8, which of course deliver different values. 

Ok, good. As long as for the same input the routines are giving the same output, I'm happy.

> I'm not really sure whether this is something that the user should be aware of.

I might need to think about it more, but shouldn't the user be able to specify either a number of neighbours or a distance cutoff? It's been a while since I looked at the Steinhardt math, but I've used them before at disorderly sites before so something like this should be possible. Of course you ought to get (slightly?) different results  when you include more or fewer neighbours.

liamhuber:
>  might need to think about it more, but shouldn't the user be able to specify either a number of neighbours or a distance cutoff?

Yeah, you have this right in your code. I think everything is fine then. Getting different results for 8 and 14 neighbours is expected, and I don't think any special user alert is needed.

pyiron_atomistics/atomistics/structure/neighbors.py
```diff
@@ -524,6 +526,58 @@ def _check_width(self, width, pbc):
                 return True
         return False
 
+    def get_spherical_harmonics(self, l, m, cutoff_radius=np.inf, rotation=None):
+        """
+        Args:
+            l (int/numpy.array): Degree of the harmonic (int); must have ``l >= 0``.
+            m (int/numpy.array): Order of the harmonic (int); must have ``|m| <= l``.
+            rotation (numpy.array/list): Rotation to make sure phi does not become nan
```
pmrv: Should document shape of `rotation`.  It makes sense as a rotation matrix, but it should be mentioned, otherwise I see myself already checking source in a few months. ;P

pyiron_atomistics/atomistics/structure/neighbors.py
```diff
@@ -524,6 +526,58 @@ def _check_width(self, width, pbc):
                 return True
         return False
 
+    def get_spherical_harmonics(self, l, m, cutoff_radius=np.inf, rotation=None):
+        """
```
pmrv, line 530: Would be cleaner if it mentions that it operates on `self.vecs`.  This makes sense, but less mental burden on users is always better.

srmnitc:
@liamhuber @samwaseda My thoughts about the bcc issue. They do indeed give drastically different values. If you sort of look at the q4-q6 maps that I normally show in my presentations, the whole bcc distribution will shift depending on if you include just the first shell or the next one too. In pyscal, you can explicitly change it with the neighbor method. The default behaviour is inspired by high-temperature calculations. Close to the melting temperature, the first and second shells become more or less indistinguishable, thus it makes sense to include both.  

You can see (not explicitly) the differences if include two shells [here](https://pyscal.org/part3/03_steinhardt_parameters.html).

tests/atomistics/structure/test_neighbors.py
```diff
@@ -351,6 +351,20 @@ def test_chemical_symbols(self):
...
+        self.assertAlmostEqual(0, neigh.get_steinhardt_parameter(2)[0])
+        self.assertAlmostEqual(0.19094065395649323, neigh.get_steinhardt_parameter(4)[0])
+        self.assertAlmostEqual(0.5745242597140696, neigh.get_steinhardt_parameter(6)[0])
+        neigh = StructureFactory().ase_bulk('Ti').get_neighbors(num_neighbors=12)
+        self.assertAlmostEqual(0.0845126986439901, neigh.get_steinhardt_parameter(4)[0])
+        self.assertAlmostEqual(0.4812230183483194, neigh.get_steinhardt_parameter(6)[0])
+        neigh = StructureFactory().ase_bulk('Fe').get_neighbors(num_neighbors=14)
+        self.assertAlmostEqual(0.03636964837266537, neigh.get_steinhardt_parameter(4)[0])
+        self.assertAlmostEqual(0.5106882308569508, neigh.get_steinhardt_parameter(6)[0])
+        with self.assertRaises(ValueError):
```
liamhuber: NIT: why not `self.assertRaises(ValueError, neigh.get_steinhardt_parameter, 2, 2)`?

pyiron_atomistics/atomistics/structure/neighbors.py
```diff
@@ -524,6 +526,58 @@ def _check_width(self, width, pbc):
...
+        Args:
+            l (int/numpy.array): Degree of the harmonic (int); must have ``l >= 0``.
+            m (int/numpy.array): Order of the harmonic (int); must have ``|m| <= l``.
+            rotation (numpy.array/list): Rotation to make sure phi does not become nan
+
+        Returns:
+            (numpy.array) spherical harmonic values
```
pmrv: For the same reasons as the other comments, this should document the shape (same as `self.distances` I guess).

liamhuber (approved):
lgtm

pmrv (approved):
Looks nice! 

For the tests, I think it's ok to just compare to pyscal, but I'll follow whatever @liamhuber  is suggesting.

liamhuber:
> @liamhuber @samwaseda My thoughts about the bcc issue. They do indeed give drastically different values. If you sort of look at the q4-q6 maps that I normally show in my presentations, the whole bcc distribution will shift depending on if you include just the first shell or the next one too. In pyscal, you can explicitly change it with the neighbor method. The default behaviour is inspired by high-temperature calculations. Close to the melting temperature, the first and second shells become more or less indistinguishable, thus it makes sense to include both.

👍 Sounds very reasonable to me. In this case I'm content with the `for automated neighbours blah blah see pyscal` warning and just insisting that the user specify a cutoff if they want something other than the neighbour count coming with the `Neighbors` object already.

pyiron_atomistics/atomistics/structure/neighbors.py
liamhuber: > This makes sense, but less mental burden on users is always better.

💯

srmnitc (approved):
everything seems okay.

srmnitc:
> 
> 
> > Sorry I said something super misleading - I compared fcc and hcp (because bcc was missing in Yuji's data) and hcp differed.
> 
> I guess BCC is 'cube' in Yuji's notation. All his positions seem to be relative to an atom at [0,0,0]. If your/pyscal's calculations and Yuji's disagree for HCP and the input is the same I guess this warrants looking into. By democracy I guess it's probably Yuji's code that's wrong^1, but in case he's right it's important to fix pyscal!
> 
> ^1 still way less wrong than my implementation he was correcting, for the record.

I just took a look at the hcp difference. This is due to the different c/a ratio as @liamhuber said. Yuji's data uses a perfect one, while Ti in the tests is slightly different. So Yuji's code, pyscal, and this implementation are all correct. If you are interested, you can see some detailed tests with pyscal [here](https://github.com/srmnitc/pyscal_boo_test/blob/main/notebooks/test.ipynb).

Changing the tests to use Mg for the hcp structure should give agreement between all codes.

tests/atomistics/structure/test_neighbors.py
```diff
@@ -351,6 +351,20 @@ def test_chemical_symbols(self):
         neigh = vacancy.get_neighbors(num_neighbors=None, cutoff_radius=3)
         self.assertEqual(neigh.chemical_symbols[0,-1], 'v')
 
+    def test_steinhardt_parameters(self):
+        neigh = StructureFactory().ase_bulk('Al').get_neighbors(num_neighbors=12)
+        # values obtained with pyscal
+        self.assertAlmostEqual(0, neigh.get_steinhardt_parameter(2)[0])
+        self.assertAlmostEqual(0.19094065395649323, neigh.get_steinhardt_parameter(4)[0])
+        self.assertAlmostEqual(0.5745242597140696, neigh.get_steinhardt_parameter(6)[0])
+        neigh = StructureFactory().ase_bulk('Ti').get_neighbors(num_neighbors=12)
```
srmnitc: Changing the hcp structure to Mg should make comparison to reported publications such as [this](https://doi.org/10.1063/1.4774084) easier. The slightly lower c/a ratio of Ti changes the value of Steinhardt parameters.

samwaseda: Ah super nice! Thanks!

pyiron_atomistics/atomistics/structure/neighbors.py
samwaseda: What's the common notation? Do I say `3x3 numpy.array/list` or should it come after the colon?

pyiron_atomistics/atomistics/structure/neighbors.py
samwaseda: It's true that it's what reviewing process is for, but I really admire your ability to spot important details XD

pyiron_atomistics/atomistics/structure/neighbors.py
pmrv: Numpy does it this way
```
a : (..., M, M) array_like
    Input array to compute determinants for.
```
(from numpy.linalg.det). I'd just follow like this
```suggestion
            rotation ( (3,3) numpy.array/list): Rotation to make sure phi does not become nan
```

pyiron_atomistics/atomistics/structure/neighbors.py
samwaseda: It should be `(n,)`. And again, I don't know what's the expected notation XD `((natoms,) numpy.array)`?

tests/atomistics/structure/test_neighbors.py
samwaseda: Because I didn't know that such a possibility existed :D

pyiron_atomistics/atomistics/structure/neighbors.py
pmrv: Ja, I think that would work.  It's just important, that using it you know which axes mean what.

PR #189, opened 2021-05-03
Make get_structure() faster: Avoid tolist()+array()

pmrv:
Similar to #46, affords ~25% speed up.

See also #190 that is merged into this PR.

### DIFF
--- a/pyiron_atomistics/atomistics/job/atomistic.py
+++ b/pyiron_atomistics/atomistics/job/atomistic.py
@@ -778,19 +778,19 @@ class GenericOutput(object):
 
     @property
     def cells(self):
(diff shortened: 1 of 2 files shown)


pmrv:
From the profiling I also find that our HDF5 layer has at least ~300% overhead as compared `h5io` directly.  I won't dig deeper now, but we should have that on the watch list.

pmrv:
We discussed on Friday that changing the lookup order is the cleaner solution, so I implemented that [here](https://github.com/pyiron/pyiron_base/pull/309).  I'll close this PR and open a new one with the remaining changes of this PR, so the discussion stays clean.

PR #196, opened 2021-05-13
Make ´delete_existing_job´ available to the ´AtomisticGenericJob´ class

raynol-dsouza:
The `AtomisticGenericJob` class inherits from the `GenericJobCore`/`GenericJob` class. While the `GenericJobCore.copy_to()` has the argument `delete_existing_job`, `AtomisticGenericJob.copy_to()` doesn't. Since Protocol jobs create child jobs, the workflows of some of the them rely on an existing child job being deleted, while using `copy_to()`. Hence, the PR.

In addition, PyCharm kept complaining that the `copy_to()` of `AtomisticGenericJob` did not have the same signature as the `copy_to()` of `GenericJobCore`.

### DIFF
--- a/pyiron_atomistics/atomistics/job/atomistic.py
+++ b/pyiron_atomistics/atomistics/job/atomistic.py
@@ -158,25 +158,30 @@ class AtomisticGenericJob(GenericJobCore, HasStructure):
         """
         self._generic_input.read_only = True
 
-    def copy_to(
-        self, project=None, new_job_name=None, input_only=False, new_database_entry=True
-    ):
+    def copy_to(self, project=None, new_job_name=None, input_only=False, new_database_entry=True,
+                delete_existing_job=False):
         """
+        Copy the content of the job including the HDF5 file to a new location.
 
         Args:
-            destination:
-            new_job_name:
-            input_only:
-            new_database_entry:
+            project (JobCore/ProjectHDFio/Project/None): The project to copy the job to.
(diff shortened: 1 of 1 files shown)


liamhuber:
Will not get to the review until Monday, but nice catch!

liamhuber (approved):
The functional code looks perfect; I streamlined the docstring part over in #200 so please review and merge that first.

Not related to your PR, but while we're touching this bit of code: I was unhappy to see that the extra functionality in this child class (a) used `not` instead of an explicit `is not None` check, and (b) wasn't covered in the unit tests. So once we get this signature fix in, I'll go write a new unit test.

pmrv (approved):
Looks good to me, but I had a itch to make it a bit cleaner in pyiron_base and open [this](https://github.com/pyiron/pyiron_base/pull/320) and [this](https://github.com/pyiron/pyiron_atomistics/pull/201).

This would make this PR not necessary, but I feel the new PR needs a bit more architecture discussion, so I'd suggest we merge this so that @raynol-dsouza gets this work done and then work on the tests in the new PR.

liamhuber:
> This would make this PR not necessary, but I feel the new PR needs a bit more architecture discussion, so I'd suggest we merge this so that @raynol-dsouza gets this work done and then work on the tests in the new PR.

Yeah, I like this. I'm a giant fan of solving the problem as close to the source as possible, but for the sake of immediacy there's no reason to delay merging this more local fix!

PR #200, opened 2021-05-17
No duplicated docstring text

liamhuber:
Exploit `functools.wraps` to avoid re-writing any docstring text.


raynol-dsouza (approved):
Aha. I did initially use `copy_to.__doc__` but couldn’t figure out how to replace `GenericJob` with `AtomisticGenericJob`.

liamhuber:
> Aha. I did initially use `copy_to.__doc__` but couldn’t figure out how to replace `GenericJob` with `AtomisticGenericJob`.

@raynol-dsouza yeah, actually I just learned about `__qualname__` through doing this, so it was a useful exercise for me! The problem was referencing the name of the object (class in this case) from *inside* that object, and that's what this is explicitly for.

PR #201, opened 2021-05-17
Use new copy hook

pmrv:
Pending [this](https://github.com/pyiron/pyiron_base/pull/320) PR in pyiron_base.   Alternative to https://github.com/pyiron/pyiron_atomistics/pull/196.


pyiron_atomistics/atomistics/job/atomistic.py
```diff
@@ -158,29 +158,9 @@ def set_input_to_read_only(self):
...
-            project=project,
-            new_job_name=new_job_name,
-            input_only=input_only,
-            new_database_entry=new_database_entry,
-        )
-        if not new_generic_job._structure:
-            new_generic_job._structure = copy.copy(self._structure)
-        return new_generic_job
+    def _after_generic_copy_to(self, reloaded):
+        if not self._structure:
```
niklassiemer:
```suggestion
        if self._structure is None:
```

pmrv: Good catch!

PR #203, opened 2021-05-19
remove tree-internal periodic boundary detection

samwaseda:
The internal algorithm of `cKDTree` for the periodic boundary conditions was removed, because:

- It was not faster (even actually slower)
- It was applied in very few cases (neighbour search width smaller than half the smallest box length, orthogonal box etc.)
- `cKDTree` raises error if there's an atom which lies outside of the box, which happens every now and then due to python precision.

The third point was the main one for me to remove this feature, but anyway there was no obvious advantage from both the practical point of view (meaning computation time) and the programmer friendliness (meaning it was applied only in a small number of cases).


pmrv (approved):
I don't entirely understand the issue, but it makes the code shorter.  If you benchmarked it for large cells too, I'm fine with this.

PR #205, opened 2021-05-20
Steinhardt interface

liamhuber:
The pyscal Steinhardt docs were incomplete in the return values. In the process of updating the docs I also made it return numpy arrays, stopped recasting the clustering values as a boolean (as this casting negates most of the benefit from `n_clusters>2`), and streamlined the input by deprecating the boolean `clustering` arg (just set `n_clusters=None`) instead. Tests were updated alongside this.

The one thing I'm still really unhappy with is that the docstring is manually duplicated. We should be able to get around this with `functools.wraps`, but I want to do that separately -- I think we should be able to do this for all the pyscal routines that basically just pops off the `atoms` argument line. Easier to do once we subdivide the `Analyze` object to have a `Pyscal` attribute, IMO.

### DIFF
--- a/pyiron_atomistics/atomistics/structure/analyse.py
+++ b/pyiron_atomistics/atomistics/structure/analyse.py
@@ -67,7 +67,7 @@ class Analyse:
                 description: sklearn.cluster.AgglomerativeClustering
             id_list (list/numpy.ndarray): List of atoms for which the layers
                 should be considered.
-            wrap_atoms: TODO.
+            wrap_atoms (bool): Whether to consider periodic boundary conditions according to the box definition or not. If set to `False`, atoms lying on box boundaries are considered to belong to different layers, regardless of whether the box itself has the periodic boundary condition in this direction or not. If `planes` is not `None` and `wrap_atoms` is `True`, this tag has the same effect as calling `get_layers()` after calling `center_coordinates_in_unit_cell()`
             planes (list/numpy.ndarray): Planes along which the layers are calculated. Planes are
                 given in vectors, i.e. [1, 0, 0] gives the layers along the x-axis. Default planes
                 are orthogonal unit vectors: [[1, 0, 0], [0, 1, 0], [0, 0, 1]]. If you have a
--- a/pyiron_atomistics/atomistics/structure/analyse.py
+++ b/pyiron_atomistics/atomistics/structure/analyse.py
@@ -4,7 +4,7 @@
 
 import numpy as np
 from pyiron_base import Settings
-from sklearn.cluster import AgglomerativeClustering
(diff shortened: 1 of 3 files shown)


pyiron_atomistics/atomistics/structure/analyse.py
```diff
@@ -63,6 +67,7 @@ def get_layers(self, distance_threshold=0.01, id_list=None, wrap_atoms=True, pla
                 description: sklearn.cluster.AgglomerativeClustering
             id_list (list/numpy.ndarray): List of atoms for which the layers
                 should be considered.
+            wrap_atoms: TODO.
```
liamhuber: @samwaseda can you fill this in please?

samwaseda:
```suggestion
            wrap_atoms (bool): Whether to consider periodic boundary conditions according to the box definition or not. If set to `False`, atoms lying on box boundaries are considered to belong to different layers, regardless of whether the box itself has the periodic boundary condition in this direction or not. If `planes` is not `None` and `wrap_atoms` is `True`, this tag has the same effect as calling `get_layers()` after calling `center_coordinates_in_unit_cell()`
```

samwaseda: This might be too long for this section - maybe the second part (starting from *If set to*) should be added as a comment at the end of the Doc String.

samwaseda (approved):
If it's mostly about formatting, I might actually not be the right person to do the review...

pyiron_atomistics/atomistics/structure/analyse.py
liamhuber: Perfect, thanks. I wouldn't worry about it too much; because of the indentation formatting people should be able to skip over this parameter while reading pretty easily if they're not interested.

liamhuber:
> If it's mostly about formatting, I might actually not be the right person to do the review...

Fair, but you're listed as maintainer on the file so you're stuck with it 😉 Anyhow, I thought that since you're maintainer you might also be someone who actually uses this and I did slightly change the behaviour without backwards compatibility (the cluster indices are indices now instead of a boolean array), so I wanted to make sure it got brought to your attention 👍

pyiron_atomistics/atomistics/structure/analyse.py
samwaseda: Sorry one last thing: Instead of *atoms lying on box boundaries*, maybe better *atoms lying on **opposite** box boundaries*.

liamhuber: 👍 Done

srmnitc (approved):
Everything looks good.

PR #206, opened 2021-05-22
Move create pipeline

liamhuber:
It used to live on `pyiron_base.GenericJob` but Ii references `self.project.create_pipeline`, and `pyiron_base.Project` has no such method.


pmrv (approved):
lgtm

PR #208, opened 2021-05-25
Re-add check for building the doc

niklassiemer:
It seems we lost the check if the docs are building here and some changes here cause the docs test to fail [here](https://github.com/pyiron/pyiron/pull/1340#).

### DIFF
--- a/.ci_support/environment-docs.yml
+++ b/.ci_support/environment-docs.yml
@@ -9,3 +9,4 @@ dependencies:
   - Graphviz
   - sphinxcontrib-applehelp
   - seaborn
+  - jinja2 <3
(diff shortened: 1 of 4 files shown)


niklassiemer:
@pmrv I do not really get what is going wrong on building the docs... could you have a quick glance at it?

pmrv:
> @pmrv I do not really get what is going wrong on building the docs... could you have a quick glance at it?

Will check tomorrow.

pmrv:
For some reason the output of the CI and the output on my local machine are completely different, so I'm not quite sure where to look for things.

niklassiemer:
I will also again try to reproduce the error... The docs are building, right?...

pmrv:
No, they are also failing on readthedocs with the same [error](https://readthedocs.org/projects/pyiron-atomistics/builds/13860539/) as on the CI.

niklassiemer:
Sorry for e8112de6dc7b5b29c29d84cfe8c1959a329ff8e1 this should have gone in here and not directly to the master! However, lucky me, this is only used for building the docs to the best of my knowledge.

PR #209, opened 2021-05-25
Add optional method to allow non-integer frames

pmrv:
Adds an optional method to `HasStructure`.

Some sub classes may not only have integer ids associated to structures,
but also (string) names.  They can override _translate_frame to enable
such behavior.

I want this for the new structure list we're working on [here](https://github.com/pyiron/pyiron_contrib/pull/148) and for the `TrainingContainer`.

### DIFF
--- a/pyiron_atomistics/atomistics/structure/has_structure.py
+++ b/pyiron_atomistics/atomistics/structure/has_structure.py
@@ -37,6 +37,9 @@ class HasStructure(ABC):
     Sub classes that wish to document special behavior of their implementation of :method:`.get_structure` may do so by
     adding documention to it in the "Methods:" sub section of their class docstring.
 
+    Sub classes may support custom data types as indices for `frame` in :method:`.get_structure()` by overriding
+    :method:`._translate_frame()`.
+
     The example below shows how to implement this mixin and how to check whether an object derives from it
 
     >>> from pyiron_atomistics.atomistics.structure.atoms import Atoms
@@ -71,7 +74,8 @@ class HasStructure(ABC):
         calculation has been run on it, see :property:`.number_of_structures`.
 
         Args:
-            frame (int): index of the structure requested, if negative count from the back
+            frame (int, object): index of the structure requested, if negative count from the back; if
+            :method:`_translate_frame()` is overridden, frame will pass through it
             iteration_step (int): deprecated alias for frame
             wrap_atoms (bool): True if the atoms are to be wrapped back into the unit cell
 
@@ -83,6 +87,12 @@ class HasStructure(ABC):
         """
         if iteration_step is not None:
             frame = iteration_step
+        if not isinstance(frame, int):
+            try:
+                frame = self._translate_frame(frame)
+            except NotImplementedError:
+                raise KeyError(f"argument frame {frame} is not an integer and _translate_frame() not implemented!") \
+                        from None
         num_structures = self.number_of_structures
         if frame < 0:
             frame += num_structures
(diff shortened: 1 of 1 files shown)


pyiron_atomistics/atomistics/structure/has_structure.py
```diff
@@ -83,6 +87,12 @@ def get_structure(self, frame=-1, wrap_atoms=True, iteration_step=None):
         """
         if iteration_step is not None:
             frame = iteration_step
+        if not isinstance(frame, int):
```
liamhuber: I always run into headaches comparing for `int`yness. Can you make an array of numpy ints and check if this works as expected with those? If not, did numpy ints work OK before? In the case that traditional and numpy ints either both work now or that numpy ints never worked then I don't mind this, but we should avoid breaking numpy int interpretation.

liamhuber (commented):
Good idea 👍  One concern about the type check, but otherwise looks nice.

Leimeroth (approved):
I like it. Did you test it already with the StructureContainer?

pmrv:
> I like it. Did you test it already with the StructureContainer?

Working on the PR right now, but it'll have some more stuff too.  I hope to unify this + StructureContainer + TrainingContainer in one swoop.

pyiron_atomistics/atomistics/structure/has_structure.py
samwaseda: @liamhuber's comment should be part of (currently non-existing) unit tests

pmrv: @liamhuber I found [this gem](https://docs.python.org/3/library/numbers.html#numbers.Integral) in stdlib and checked that it works with numpy ints, so I'll use it.

@samwaseda Since it's an abstract mixin testing it directly is a bit tricky.  I could either add a dummy implementation or test the implementors in the current code base.  Thoughts?  I think it's orthogonal to this issue though and would make it in a separate PR.

samwaseda: my point is that if you're anyway checking the behaviour with both `int` and `np.int` (maybe also `np.int64`?), you can simply leave the lines in the unit test as well.

samwaseda: And indeed:

```
i = np.int32(5)
a = np.arange(10)
if isinstance(i, int):
    print(a[i])
else:
    print(f'{i} is not an integer')
```

This prints `5 is not an integer`, even though `a[i]` very well returns 5.

pmrv: With the numbers module from above this works
```python
i = np.int32(5)
a = np.arange(10)
if isinstance(i, numbers.Integral):
    print(a[i])
else:
    print(f'{i} is not an integer')
```

samwaseda: ok but we can come up with several scenarios, like `numbers` for some reason gets deprecated, or there's a major overhaul for this code or whatsoever, where someone thinks of rewriting these lines while believing they are maintaining the idea (i.e. `frame` is an `int`), even though in reality they might by mistake replace `numbers.Integral` by `int`, in which case the presence of a unit test containing `np.int32` etc. might be a reminder that `int` is not sufficient for this `if`-statement.

samwaseda: I'm not 100% if I understood your comment, but if you're saying that for the testing you need an underlying structure, I guess the easiest solution is to overload `_get_structure` for the testing purpose.

pmrv: > I'm not 100% if I understood your comment, but if you're saying that for the testing you need an underlying structure, I guess the easiest solution is to overload `_get_structure` for the testing purpose.

Jep, that's what I meant.  It's possible to just sub classes for the test and I can do it, but as I said I'd rather leave it for another day.

liamhuber (approved):

PR #213, opened 2021-05-27
Phonopy options

liamhuber:
`PhonopyJob` now accepts as input `number_of_snapshots`, which is an argument of the same name for the underlying `phonopy` object's `generate_displacement` method. Instead of using symmetry arguments for building up a displacement set, it just makes a collection of random displacements.

I ran into trouble with large, low-symmetry defect structures wanting to make thousands of displacements. Since I only need a rough approximation of the harmonic behaviour as a reference point for a thermodynamic integration run, I want to be able to choose how many displacement frames I take -- thankfully phonopy has exactly a way to do this and it just needed to be exposed.

Currently it relies on the `alm` package. This is available on conda forge already, but only for OSX and Linux. @pmrv, could you please briefly review to see if I'm using `ImportAlarm` appropriately and whether you have any clever ideas about including this as an optional dependency? I think we talked about that topic at a recent pyiron meeting.

@raynol-dsouza if you could take a look at the actual phonopy stuff itself that would be great. I hope it will be useful for you too.

### DIFF
--- a/pyiron_atomistics/atomistics/master/phonopy.py
+++ b/pyiron_atomistics/atomistics/master/phonopy.py
@@ -16,7 +16,7 @@ from phonopy.file_IO import write_FORCE_CONSTANTS
 from pyiron_atomistics.atomistics.structure.atoms import Atoms
 from pyiron_atomistics.atomistics.master.parallel import AtomisticParallelMaster
 from pyiron_atomistics.atomistics.structure.phonopy import publication as phonopy_publication
-from pyiron_base import JobGenerator, Settings
+from pyiron_base import JobGenerator, Settings, ImportAlarm
 
 __author__ = "Jan Janssen, Yury Lysogorskiy"
 __copyright__ = (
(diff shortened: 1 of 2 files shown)


pyiron_atomistics/atomistics/master/phonopy.py
```diff
@@ -510,4 +522,11 @@ def validate_ready_to_run(self):
             raise ValueError("Phonopy reference jobs should be static calculations, but got {}".format(
                 self.ref_job._generic_input["calc_mode"]
             ))
+        if self.input["number_of_snapshots"] is not None:
+            with ImportAlarm(
+                    "Phonopy with random snapshot displacement relies on the alm module but this unavailable. Please "
+                    "ensure your python environment contains alm, e.g. by running `conda install -c conda-forge alm`."
+                    "Note: at time of writing alm is only available for Linux and OSX systems."
+            ) as import_alarm:
```
pmrv, line 530: That doesn't quite work.  `import_alarm` is designed to be a decorator after the `with` statement.  I made a quick [patch](https://github.com/pyiron/pyiron_base/pull/328) to trigger the warning manually.

pmrv (changes requested):

pyiron_atomistics/atomistics/master/phonopy.py
liamhuber: Pushed an update -- like this?

pmrv: No, the message is updated after the code exits the with statement.  I'll update the docstring in the other PR.

pyiron_atomistics/atomistics/master/phonopy.py
```diff
@@ -510,4 +522,12 @@ def validate_ready_to_run(self):
...
+                import alm
+                import_alarm.warn_if_failed()
```
pmrv:
```suggestion
            import_alarm.warn_if_failed()
```

PR #217, opened 2021-05-31
Lammps units

sudarsan-surendralal:
Implemented proper conversion of LAMMPS output into pyiron units on parsing. Made the following changes

1. Conversion factors defined in a new file `units.py`
2. Avoiding unnecessary conversion of numpy array to list before storing

### DIFF
--- a/pyiron_atomistics/lammps/base.py
+++ b/pyiron_atomistics/lammps/base.py
@@ -569,12 +569,12 @@ class LammpsBase(AtomisticGenericJob):
                 h5_dict[key] = key.replace('f_', '')
 
             df = df.rename(index=str, columns=h5_dict)
+            pressure_dict = dict()
             if all([x in df.columns.values for x in ["Pxx", "Pxy", "Pxz", "Pxy", "Pyy", "Pyz", "Pxz", "Pyz", "Pzz"]]):
                 pressures = np.stack(
                     (df.Pxx, df.Pxy, df.Pxz, df.Pxy, df.Pyy, df.Pyz, df.Pxz, df.Pyz, df.Pzz),
                     axis=-1,
                 ).reshape(-1, 3, 3).astype('float64')
-                pressures *= 0.0001  # bar -> GPa
 
                 # Rotate pressures from Lammps frame to pyiron frame if necessary
                 rotation_matrix = self._prism.R.T
@@ -586,7 +586,7 @@ class LammpsBase(AtomisticGenericJob):
                         ((df.columns.str.len() == 3) & df.columns.str.startswith("P"))
                     ]
                 )
-                df["pressures"] = pressures.tolist()
+                pressure_dict["pressure"] = pressures
             else:
                 warnings.warn("LAMMPS warning: log.lammps does not contain the required pressure values.")
             if 'mean_pressure[1]' in df.columns:
@@ -596,7 +596,6 @@ class LammpsBase(AtomisticGenericJob):
                      df['mean_pressure[5]'], df['mean_pressure[6]'], df['mean_pressure[3]']),
                     axis=-1,
                 ).reshape(-1, 3, 3).astype('float64')
-                pressures *= 0.0001  # bar -> GPa
                 if np.matrix.trace(rotation_matrix) != 3:
                     pressures = rotation_matrix.T @ pressures @ rotation_matrix
                 df = df.drop(
@@ -604,11 +603,9 @@ class LammpsBase(AtomisticGenericJob):
                         (df.columns.str.startswith("mean_pressure") & df.columns.str.endswith(']'))
                     ]
                 )
-                # Conversion to list for some reason?
-                df["mean_pressures"] = pressures.tolist()
+                pressure_dict["mean_pressures"] = pressures
 
-            generic_keys_lst = list(h5_dict.values()) + ["pressures", "mean_pressures", ]
-
+            generic_keys_lst = list(h5_dict.values())
             # For unit conversion
             units_key_list = np.array(list(LAMMPS_UNIT_CONVERSIONS[self.input.control["units"]].keys()))
             with self.project_hdf5.open("output/generic") as hdf_output:
@@ -619,13 +616,13 @@ class LammpsBase(AtomisticGenericJob):
                         array_quantity = np.array(v)
                         key_check = np.array([val in k for val in units_key_list])
                         if any(key_check):
-                            # Convert list element to numpy array
-                            if isinstance(array_quantity[0], list):
-                                array_quantity = np.array([np.array(val) for val in array_quantity])
                             hdf_output[k] = array_quantity / LAMMPS_UNIT_CONVERSIONS[
                                 self.input.control["units"]][units_key_list[key_check == True][0]]
                         else:
                             hdf_output[k] = array_quantity
+                # Store pressures as numpy arrays
+                for key, val in pressure_dict.items():
+                    hdf_output[key] = val / LAMMPS_UNIT_CONVERSIONS[self.input.control["units"]]["pressure"]
 
             with self.project_hdf5.open("output/lammps") as hdf_output:
                 # This is a hack for backward comparability
--- a/pyiron_atomistics/lammps/base.py
+++ b/pyiron_atomistics/lammps/base.py
@@ -19,6 +19,8 @@ from pyiron_base import Settings, extract_data_from_file, deprecate
 from pyiron_atomistics.lammps.control import LammpsControl
 from pyiron_atomistics.lammps.potential import LammpsPotential
 from pyiron_atomistics.lammps.structure import LammpsStructure, UnfoldingPrism
+from pyiron_atomistics.lammps.units import LAMMPS_UNIT_CONVERSIONS
+
 
 __author__ = "Joerg Neugebauer, Sudarsan Surendralal, Jan Janssen"
 __copyright__ = (
@@ -464,7 +466,8 @@ class LammpsBase(AtomisticGenericJob):
             ]
             indices = [indices_i.tolist() for indices_i in h5md["/particles/all/indices/value"]]
         with self.project_hdf5.open("output/generic") as h5_file:
-            h5_file["forces"] = np.array(forces)
+            # Store after converting to pyiron units
+            h5_file["forces"] = np.array(forces) / LAMMPS_UNIT_CONVERSIONS[self.input.control["units"]]["force"]
             h5_file["positions"] = np.array(positions)
             h5_file["steps"] = np.array(steps)
             h5_file["cells"] = cell
@@ -601,14 +604,28 @@ class LammpsBase(AtomisticGenericJob):
                         (df.columns.str.startswith("mean_pressure") & df.columns.str.endswith(']'))
                     ]
                 )
+                # Conversion to list for some reason?
                 df["mean_pressures"] = pressures.tolist()
 
             generic_keys_lst = list(h5_dict.values()) + ["pressures", "mean_pressures", ]
+
+            # For unit conversion
+            units_key_list = np.array(list(LAMMPS_UNIT_CONVERSIONS[self.input.control["units"]].keys()))
             with self.project_hdf5.open("output/generic") as hdf_output:
                 # This is a hack for backward comparability
                 for k, v in df.items():
                     if k in generic_keys_lst:
-                        hdf_output[k] = np.array(v)
+                        # Convert to pyiron units wherever necessary
+                        array_quantity = np.array(v)
+                        key_check = np.array([val in k for val in units_key_list])
+                        if any(key_check):
+                            # Convert list element to numpy array
+                            if isinstance(array_quantity[0], list):
+                                array_quantity = np.array([np.array(val) for val in array_quantity])
+                            hdf_output[k] = array_quantity / LAMMPS_UNIT_CONVERSIONS[
+                                self.input.control["units"]][units_key_list[key_check == True][0]]
+                        else:
+                            hdf_output[k] = array_quantity
 
             with self.project_hdf5.open("output/lammps") as hdf_output:
                 # This is a hack for backward comparability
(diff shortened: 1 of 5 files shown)


pyiron_atomistics/lammps/base.py
```diff
@@ -583,7 +586,7 @@ def collect_output_log(self, file_name="log.lammps", cwd=None):
                         ((df.columns.str.len() == 3) & df.columns.str.startswith("P"))
                     ]
                 )
-                df["pressures"] = pressures.tolist()
+                pressure_dict["pressure"] = pressures
```
samwaseda: `pressures` to `pressure`?

This being said, I personally prefer `pressure` to `pressures`, because except for `cells` and `pressures`, per-atom quantities are always in plural and global quantities are in singular in pyiron.

sudarsan-surendralal: I too prefer `pressure` since this is a per box quantity but I now realize that all codes use `pressures`. Better to make the transition now?

samwaseda: ok then it's a topic that we should raise at the next pyiron meeting. But I don't think it has to be part of this PR -> for now I'd probably rather stick to `pressures`.

sudarsan-surendralal:
I notice that the tests fail because the pressure values are modified  because a rotation transformation is performed:

https://github.com/pyiron/pyiron_atomistics/blob/4b7223fd705db3f9e239fa4694e93860f19d4477/pyiron_atomistics/lammps/base.py#L581-L582

The trace of the rotation matrix in this test case is 2.999989 and it doesn't exactly match the value of 3. @samwaseda are we doing this correctly. It doesn't seem that for such a box a rotation transformation is necessary

samwaseda:
> The trace of the rotation matrix in this test case is 2.999989 and it doesn't exactly match the value of 3. @samwaseda are we doing this correctly. It doesn't seem that for such a box a rotation transformation is necessary

I agree that such the operation itself is fairly meaningless, but the output pressure should be nearly identical to the input pressure, so I guess the problem comes from elsewhere.

sudarsan-surendralal:
> I agree that such the operation itself is fairly meaningless, but the output pressure should be nearly identical to the input pressure, so I guess the problem comes from elsewhere.

The values are nearly identical but not identical since the pressure values have a large absolute value. So I guess we have to live with a small round off error?

pyiron_atomistics/lammps/base.py
sudarsan-surendralal: Ok done!

pyiron_atomistics/lammps/base.py
```diff
@@ -593,22 +595,32 @@ def collect_output_log(self, file_name="log.lammps", cwd=None):
...
+            units_key_list = np.array(list(LAMMPS_UNIT_CONVERSIONS[self.input.control["units"]].keys()))
             with self.project_hdf5.open("output/generic") as hdf_output:
                 # This is a hack for backward comparability
                 for k, v in df.items():
                     if k in generic_keys_lst:
-                        hdf_output[k] = np.array(v)
+                        # Convert to pyiron units wherever necessary
+                        array_quantity = np.array(v)
+                        key_check = np.array([val in k for val in units_key_list])
+                        if any(key_check):
```
liamhuber: if `key_check.sum() > 1` aren't we in trouble anyhow?

pyiron_atomistics/lammps/base.py
```diff
@@ -593,22 +595,32 @@ def collect_output_log(self, file_name="log.lammps", cwd=None):
...
+                            hdf_output[k] = array_quantity / LAMMPS_UNIT_CONVERSIONS[
+                                self.input.control["units"]][units_key_list[key_check == True][0]]
```
liamhuber: `LAMMPS_UNIT_CONVERSIONS[self.input.control["units"]][units_key_list[key_check == True][0]]` is too hard to read.

pyiron_atomistics/lammps/base.py
```diff
@@ -1002,10 +1014,18 @@ def collect_dump_file(self, file_name="dump.out", cwd=None):
...
             for kk in keys[keys.str.startswith('c_')]:
                 output[kk.replace('c_', '')] = np.array([cc[kk] for cc in content], dtype=float)
-
+            units_key_list = np.array(list(LAMMPS_UNIT_CONVERSIONS[self.input.control["units"]].keys()))
             with self.project_hdf5.open("output/generic") as hdf_output:
                 for k, v in output.items():
-                    hdf_output[k] = v
+
+                    # Convert to pyiron units wherever necessary
+                    key_check = np.array([val in k for val in units_key_list])
```
liamhuber: Duplicate code!

pyiron_atomistics/lammps/base.py
```diff
@@ -464,7 +466,8 @@ def collect_h5md_file(self, file_name="dump.h5", cwd=None):
             ]
             indices = [indices_i.tolist() for indices_i in h5md["/particles/all/indices/value"]]
         with self.project_hdf5.open("output/generic") as h5_file:
-            h5_file["forces"] = np.array(forces)
+            # Store after converting to pyiron units
+            h5_file["forces"] = np.array(forces) / LAMMPS_UNIT_CONVERSIONS[self.input.control["units"]]["force"]
             h5_file["positions"] = np.array(positions)
```
liamhuber: Good catch on converting forces! I believe we must also convert positions and cells with `LAMMPS_UNIT_CONVERSIONS[self.input.control["units"]]["distance"]`

pyiron_atomistics/lammps/units.py
```diff
@@ -0,0 +1,103 @@
...
+GPA_TO_ATM = spc.giga / spc.atm
+GPA_TO_BAR = spc.giga / spc.bar
+GPA_TO_BARYE = spc.giga / (spc.micro * spc.bar)  # "barye" = 1e-6 bar
+GPA_TO_PA = spc.giga
+
+# Conversions for most of the Lammps units to Pyiron units
+# Lammps units source doc: https://lammps.sandia.gov/doc/units.html
+# Pyrion units source doc: https://pyiron.readthedocs.io/en/latest/source/faq.html
+# At time of writing, not all these conversion factors are used, but may be helpful later.
+
```
liamhuber, line 47:
```suggestion


def convert_array_to_pyiron_units(array, name, units):
    """
    Convert a numpy array from Lammps output units to pyiron units.
    
    Args:
        array (np.ndarray): The array to convert.
        name (str): The type of data it is, e.g. `energy`, `mean_force` (which contains `force`), etc.
        units (str): The unit types Lammps is using.
        
    Returns:
        np.ndarray: The same array, rescaled by the appropriate unit conversion if necessary.
        
    Warning: 
        If `name` contains multiple unit types, e.g. `charge_temperature`, or inadvertently contains a unit, e.g. `forced_foos`, you will get some weird conversion you weren't expecting.
    """
    try:
        unit_map = LAMMPS_UNIT_CONVERSIONS[units]
        unit_type = [key for key in unit_map.keys() if key in name][0]  # Still doesn't check for double occurance...
        return array / unit_map[unit_type]
    except IndexError:  # Name does not contain a known unit type
        return array

```

liamhuber (changes requested):
I really like how you further centralized the unit conversions. I'd like to push this even a bit further though. In particular I suggest extracting the conversion right over to the new (and excellent!) `units` module and `from pyiron_atomistics.lammps.units import convert_array_to_pyiron_units` instead of `LAMMPS_UNIT_CONVERSIONS`. (Cf. review for an untested solution of what that function could look like)

Then throughout the conversions can just follow this pattern:

```python
with self.project_hdf5.open("output/generic") as hdf_output:
    for k, v in output.items():
        hdf_output[k] = convert_array_to_pyiron_units(v, k, self.input.control["units"])
```

If/when this works, the same pattern applies equally well to the regular output, pressures, and the h5md parsing.

So far I basically just refined your existing pattern; I am still concerned about how certain output names can inadvertently trigger a unit conversion, but I don't see an elegant solution atm.

sudarsan-surendralal:
Thanks @liamhuber for the review. Based on your comments, I now do the unit conversions using a class to make the code more readable and flexible. If you think this is okay, I can also add docs later!

liamhuber (commented):
Yeah, this is the perfect direction! I guess this is really going in the direction of building a little ontology for the units. Can we refactor the globals into ontologically oriented classes as well? @Ahmed-Aslam @muh-hassani @niklassiemer and whoever else was in the ontology hackathons, do you have any feedback or suggestions to give on this new `units` module? I don't want to prematurely optimize, but I do think we have the opportunity here to abstract a little bit so this can later be quite easily moved from `pyiron_atomistics.lammps.units` to `pyiron_atomistics.units`.

sudarsan-surendralal:
> 
> 
> Yeah, this is the perfect direction! I guess this is really going in the direction of building a little ontology for the units. Can we refactor the globals into ontologically oriented classes as well? @Ahmed-Aslam @muh-hassani @niklassiemer and whoever else was in the ontology hackathons, do you have any feedback or suggestions to give on this new `units` module? I don't want to prematurely optimize, but I do think we have the opportunity here to abstract a little bit so this can later be quite easily moved from `pyiron_atomistics.lammps.units` to `pyiron_atomistics.units`.

I think a general ontology for the units would be great! For the time being though, I'd prioritize optimizing and merging this module soon to get the correct outputs of LAMMPS calculations with non-metal units (for eg. our internal electrochemistry hackathon depends on this 😅). We can revisit this an genralize anytime!

liamhuber:
> > I agree that such the operation itself is fairly meaningless, but the output pressure should be nearly identical to the input pressure, so I guess the problem comes from elsewhere.
> 
> The values are nearly identical but not identical since the pressure values have a large absolute value. So I guess we have to live with a small round off error?

Is it not better to replace `if np.matrix.trace(rotation_matrix) != 3:` with `if not np.isclose(np.matrix.trace(rotation_matrix), 3):`? Genuine question -- I'm worried I'm missing an unintended side effect here.

pyiron_atomistics/lammps/base.py
```diff
@@ -1002,10 +1010,10 @@ def collect_dump_file(self, file_name="dump.out", cwd=None):
             keys = content[0].keys()
             for kk in keys[keys.str.startswith('c_')]:
                 output[kk.replace('c_', '')] = np.array([cc[kk] for cc in content], dtype=float)
-
             with self.project_hdf5.open("output/generic") as hdf_output:
                 for k, v in output.items():
-                    hdf_output[k] = v
+                    # Convert to pyiron units
```
liamhuber: redundant comment now that we have a nicely named function 😉

pyiron_atomistics/lammps/base.py
```diff
@@ -464,11 +467,12 @@ def collect_h5md_file(self, file_name="dump.h5", cwd=None):
             ]
             indices = [indices_i.tolist() for indices_i in h5md["/particles/all/indices/value"]]
         with self.project_hdf5.open("output/generic") as h5_file:
-            h5_file["forces"] = np.array(forces)
-            h5_file["positions"] = np.array(positions)
-            h5_file["steps"] = np.array(steps)
-            h5_file["cells"] = cell
-            h5_file["indices"] = self.remap_indices(indices)
+            # Store after converting to pyiron units
```
liamhuber: Also a redundant comment now that we have a nicely named function

pyiron_atomistics/lammps/base.py
```diff
@@ -593,28 +597,31 @@ def collect_output_log(self, file_name="log.lammps", cwd=None):
...
-
-            generic_keys_lst = list(h5_dict.values()) + ["pressures", "mean_pressures", ]
+                pressure_dict["mean_pressures"] = pressures
+            generic_keys_lst = list(h5_dict.values())
             with self.project_hdf5.open("output/generic") as hdf_output:
                 # This is a hack for backward comparability
                 for k, v in df.items():
                     if k in generic_keys_lst:
-                        hdf_output[k] = np.array(v)
+                        # Convert to pyiron units
```
liamhuber: Ok, final nit 😉 again this comment is now redundant, and if pycharm doesn't complain about the line length I'd try in-lining the array conversion as you do elsewhere.

liamhuber (approved):
> For the time being though, I'd prioritize optimizing and merging this module soon to get the correct outputs of LAMMPS calculations with non-metal units (for eg. our internal electrochemistry hackathon depends on this 😅). We can revisit this an genralize anytime!

I am overall uncomfortable with the idea of leaving knwon improvements to a later date when you're motivated to get this corner of the codebase cleaned up right now, **but** I am totally sensitive to the fact you guys need this, and honestly what is here is already a great improvement and is only lacking in that there is room to be even better not that it is bad to begin with. So merge away 😉

samwaseda:
> Is it not better to replace `if np.matrix.trace(rotation_matrix) != 3:` with `if not np.isclose(np.matrix.trace(rotation_matrix), 3):`? Genuine question -- I'm worried I'm missing an unintended side effect here.

I partly agree, but in my opinion we shouldn't totally depend on the numpy definition of what's close and what's not close, so we should also put the tolerance value, which ultimately should be an input value that the user should be able to set. I'm saying this mostly because whether the rotation matrix is exactly the identity matrix or close to the identity matrix should anyway not affect the overall results (i.e. it should only affect the computation time), so I find it slightly problematic that we're silently switching `!=` to `np.isclose` and potentially only forcing some users to compromise simulation precision.

liamhuber:
> so I find it slightly problematic that we're silently switching `!=` to `np.isclose` and potentially only forcing some users to compromise simulation precision.

I was mostly concerned that we will literally never trigger the condition with a strict `!=`. This concern stemmed from that fact that on conversion to/from Lammps we're anyhow picking up machine-precision level error in our cells:

```python
pr = Project('scratch')
job = pr.create.job.Lammps('tmp', delete_existing_job=True)
job.structure = pr.create.structure.bulk('Al', cubic=True)
job.run()
print(job.get_structure().cell.array)
>>> [[4.05000000e+00 2.47990977e-16 2.47990977e-16]
     [0.00000000e+00 4.05000000e+00 2.47990977e-16]
     [0.00000000e+00 0.00000000e+00 4.05000000e+00]]
```

However, now I see that this doesn't actually add any error to the *trace* of the rotation matrix (at least in this simple case...)

```python
print(job._get_rotation_matrix(pressure=0))
>>> [[1.000000e+00 0.000000e+00 0.000000e+00]
     [6.123234e-17 1.000000e+00 0.000000e+00]
     [6.123234e-17 6.123234e-17 1.000000e+00]]
```

It is still a float-to-int comparison with an int-like comparator though, which makes me mildly uncomfortable. I don't grok all the transformations involved though, so I'm very happy to take your word that this simple example of it working fine even with the `!=` is enough.

>>> so we should also put the tolerance value, which ultimately should be an input value that the user should be able to set. 

If/when we make a switch to `isclose` I don't object to explicitly setting the tolerance level, but my only intent here was really to get rid of machine noise and at least for that we don't need any user input.

samwaseda:
> I was mostly concerned that we will literally never trigger the condition with a strict `!=`. This concern stemmed from that fact that on conversion to/from Lammps we're anyhow picking up machine-precision level error in our cells:

Yeah I share the same concern, but the problem is @jan-janssen and I have already seen that the numpy precision is simply not enough, which is the reason that [this line](https://github.com/pyiron/pyiron_atomistics/blob/06d54acd2f8db0e4aef4ba3f07ff84923082a4f4/pyiron_atomistics/atomistics/job/interactive.py#L165) (among other) contains the explicit precision.

sudarsan-surendralal:
I will be merging this PR. However, some of the feature requests born out of the discussion here are now documented as separate issues (https://github.com/pyiron/pyiron_atomistics/issues/229 and https://github.com/pyiron/pyiron_atomistics/issues/230)

PR #219, opened 2021-05-31
Get interstitials

samwaseda:
This function internally does the following steps:

            1. Initialize grid points (or Voronoi vertices) which are considered as
                interstitial site candidates.
            2. Eliminate points within a distance from the nearest neighboring atoms as
                given by `min_distance`
            3. Initialize neighbor environment using `get_neighbors`
            4. Shift interstitial candidates to the nearest symmetric points with respect to the
                neighboring atom sites/vertices.
            5. Kick out points with large neighbor distance variances
            6. Cluster interstitial candidates to avoir point overlapping.

        In complex structures (i.e. grain boundary, dislocation etc.), the default parameters
        should be chosen properly. In order to see other quantities, which potentially
        characterize interstitial sites, see the following functions from
        `structure.analyse.interstitials`:

            - `get_variance()`
            - `get_distance()`
            - `get_steinhardt_parameter()`
            - `get_volume()`
            - `get_area()`

### DIFF
--- a/pyiron_atomistics/atomistics/structure/analyse.py
+++ b/pyiron_atomistics/atomistics/structure/analyse.py
@@ -95,12 +95,11 @@ class Interstitials:
             shaped interstitials
         6. Cluster interstitial candidates to avoid point overlapping.
 
-    The interstitial sites can be obtained through `get_positions`
+    The interstitial sites can be obtained via `positions`
 
     In complex structures (i.e. grain boundary, dislocation etc.), the default parameters
     should be chosen properly. In order to see other quantities, which potentially
-    characterize interstitial sites, see the following functions from
-    `structure.analyse.interstitials`:
+    characterize interstitial sites, see the following class methods:
 
         - `get_variances()`
         - `get_distances()`
@@ -108,33 +107,79 @@ class Interstitials:
         - `get_volumes()`
         - `get_areas()`
     """
-    def __init__(self, structure, n_gridpoints_per_angstrom=5, min_distance=1, use_voronoi=False):
+    def __init__(self,
+        structure,
+        num_neighbors,
+        n_gridpoints_per_angstrom=5,
+        min_distance=1,
+        use_voronoi=False,
+        variance_buffer=0.01,
+        n_iterations=2,
+        eps=0.1,
+    ):
         """
+
         Args:
-            structure (pyiron_atomistics.atomistics.structure.atoms.Atoms): Reference structure
+            num_neighbors (int): Number of neighbors/vertices to consider for the interstitial
+                sites. By definition, tetrahedral sites should have 4 vertices and octahedral
+                sites 6.
             n_gridpoints_per_angstrom (int): Number of grid points per angstrom for the
                 initialization of the interstitial candidates. The finer the mesh (i.e. the larger
                 the value), the likelier it is to find the correct sites but then also it becomes
                 computationally more expensive. Ignored if `use_voronoi` is set to `True`
             min_distance (float): Minimum distance from the nearest neighboring atoms to the
-                positions for them to be considered as interstitial site candidates.
+                positions for them to be considered as interstitial site candidates. Set
+                `min_distance` to 0 if no point should be removed.
             use_voronoi (bool): Use Voronoi vertices for the initial interstitial candidate
                 positions instead of grid points.
+            variance_buffer (bool): Maximum permitted variance value (in distance unit) of the
+                neighbor distance values with respect to the minimum value found for each point.
+                It should be close to 0 for perfect crystals and slightly higher values for
+                structures containing defects. Set `variance_buffer` to `numpy.inf` if no selection
+                by variance value should take place.
+            n_iterations (int): Number of iterations for the shifting of the candidate positions
+                to the nearest symmetric positions with respect to `num_neighbors`. In most of the
+                cases, 1 is enough. In some rare cases (notably tetrahedral sites in bcc), it
+                should be at least 2. It is unlikely that it has to be larger than 2. Set 
+                `n_iterations` to 0 if no shifting should take place.
+            eps (float): Distance below which two interstitial candidate sites to be considered as
+                one site after the symmetrization of the points. Set `eps` to 0 if clustering should
+                not be done.
         """
-        self._num_neighbors = None
         self._hull = None
         self._neigh = None
         self._positions = None
-        self.min_distance = min_distance
+        self.num_neighbors = num_neighbors
         self.structure = structure
+        self._initialize(
+            n_gridpoints_per_angstrom=n_gridpoints_per_angstrom,
+            min_distance=min_distance,
+            use_voronoi=use_voronoi,
+            variance_buffer=variance_buffer,
+            n_iterations=n_iterations,
+            eps=eps
+        )
+
+    def _initialize(
+        self,
+        n_gridpoints_per_angstrom=5,
+        min_distance=1,
+        use_voronoi=False,
+        variance_buffer=0.01,
+        n_iterations=2,
+        eps=0.1
+    ):
         if use_voronoi:
             self.positions = structure.analyse.get_voronoi_vertices()
         else:
             self.positions = self._create_gridpoints(
                 n_gridpoints_per_angstrom=n_gridpoints_per_angstrom
             )
-        if min_distance > 0:
-            self._remove_too_close()
+        self._remove_too_close(min_distance=min_distance)
+        for _ in range(n_iterations):
+            self._set_interstitials_to_high_symmetry_points()
+        self._kick_out_points(variance_buffer=variance_buffer)
+        self._cluster_points(eps=eps)
 
     @property
     def num_neighbors(self):
@@ -215,9 +260,9 @@ class Interstitials:
             'ji,nj->ni', self.structure.cell, np.stack(positions, axis=-1).reshape(-1, 3)
         )
 
-    def _remove_too_close(self):
+    def _remove_too_close(self, min_distance=1):
         neigh = self.structure.get_neighborhood(self.positions, num_neighbors=1)
-        self.positions = self.positions[neigh.distances.flatten() > self.min_distance]
+        self.positions = self.positions[neigh.distances.flatten() > min_distance]
 
     def _set_interstitials_to_high_symmetry_points(self):
         self.positions = self.positions+np.mean(self.neigh.vecs, axis=-2)
@@ -229,6 +274,8 @@ class Interstitials:
         self.positions = self.positions[variance < min_var+variance_buffer]
 
     def _cluster_points(self, eps=0.1):
+        if eps == 0:
+            return
         extended_positions, indices = self.structure.get_extended_positions(
             eps, return_indices=True, positions=self.positions
         )
@@ -238,65 +285,6 @@ class Interstitials:
         self.positions = get_mean_positions(
             self.positions, self.structure.cell, self.structure.pbc, labels
         )
-
-    def get_positions(
-        self,
-        num_neighbors,
-        variance_buffer=0.01,
-        n_iterations=2,
-        eps=0.1,
-    ):
-        """
-        Get potential interstitial positions
-
-        Args:
-            num_neighbors (int): Number of neighbors/vertices to consider for the interstitial
-                sites. By definition, tetrahedral sites should have 4 vertices and octahedral
-                sites 6.
-            variance_buffer (bool): Maximum permitted variance value (in distance unit) of the
-                neighbor distance values with respect to the minimum value found for each point.
-                It should be close to 0 for perfect crystals and slightly higher values for
-                structures containing defects.
-            n_iterations (int): Number of iterations for the shifting of the candidate positions
-                to the nearest symmetric positions with respect to `num_neighbors`. In most of the
-                cases, 1 is enough. In some rare cases (notably tetrahedral sites in bcc), it
-                should be at least 2. It is unlikely that it has to be larger than 2.
-            eps (float): Distance below which two interstitial candidate sites to be considered as
-                one site after the symmetrization of the points.
-
-        Returns:
-            ( (n, 3) numpy.ndarray) Interstitial candidate positions.
-
-        This function internally does the following steps:
-
-            1. Initialize grid points (or Voronoi vertices) which are considered as
-                interstitial site candidates.
-            2. Eliminate points within a distance from the nearest neighboring atoms as
-                given by `min_distance`
-            3. Initialize neighbor environment using `get_neighbors`
-            4. Shift interstitial candidates to the nearest symmetric points with respect to the
-                neighboring atom sites/vertices.
-            5. Kick out points with large neighbor distance variances; this eliminates "irregular"
-                shaped interstitials
-            6. Cluster interstitial candidates to avoid point overlapping.
-
-        In complex structures (i.e. grain boundary, dislocation etc.), the default parameters
-        should be chosen properly. In order to see other quantities, which potentially
-        characterize interstitial sites, see the following functions from
-        `structure.analyse.interstitials`:
-
-            - `get_variances()`
-            - `get_distances()`
-            - `get_steinhardt_parameters()`
-            - `get_volumes()`
-            - `get_areas()`
-        """
-        self.num_neighbors = num_neighbors
-        for _ in range(n_iterations):
-            self._set_interstitials_to_high_symmetry_points()
-        self._kick_out_points(variance_buffer=variance_buffer)
-        self._cluster_points(eps=eps)
-        return self.positions
 
     def get_variances(self):
         """
@@ -355,13 +343,6 @@ class Analyse:
             structure (:class:`pyiron.atomistics.structure.atoms.Atoms`): reference Atom structure.
         """
         self._structure = structure
-        self._interstitials = None
-
-    @property
-    def interstitials(self):
-        if self._interstitials is None:
-            raise AssertionError('Initialize interstitials via get_interstitials()')
-        return self._interstitials
 
     def get_interstitials(
         self,
@@ -373,71 +354,17 @@ class Analyse:
         n_iterations=2,
         eps=0.1,
     ):
-        """
-        Get potential interstitial positions
-
-        Args:
-            num_neighbors (int): Number of neighbors/vertices to consider for the interstitial
-                sites. By definition, tetrahedral sites should have 4 vertices and octahedral
-                sites 6.
-            n_gridpoints_per_angstrom (int): Number of grid points per angstrom for the
-                initialization of the interstitial candidates. The finer the mesh (i.e. the larger
-                the value), the likelier it is to find the correct sites but then also it becomes
-                computationally more expensive. Ignored if `use_voronoi` is set to `True`
-            min_distance (float): Minimum distance from the nearest neighboring atoms to the
-                positions for them to be considered as interstitial site candidates.
-            use_voronoi (bool): Use Voronoi vertices for the initial interstitial candidate
-                positions instead of grid points.
-            variance_buffer (bool): Maximum permitted variance value (in distance unit) of the
-                neighbor distance values with respect to the minimum value found for each point.
-                It should be close to 0 for perfect crystals and slightly higher values for
-                structures containing defects.
-            n_iterations (int): Number of iterations for the shifting of the candidate positions
-                to the nearest symmetric positions with respect to `num_neighbors`. In most of the
-                cases, 1 is enough. In some rare cases (notably tetrahedral sites in bcc), it
-                should be at least 2. It is unlikely that it has to be larger than 2.
-            eps (float): Distance below which two interstitial candidate sites to be considered as
-                one site after the symmetrization of the points.
-
-        Returns:
-            ( (n, 3) numpy.ndarray) Interstitial candidate positions.
-
-        This function internally does the following steps:
-
-            1. Initialize grid points (or Voronoi vertices) which are considered as
-                interstitial site candidates.
-            2. Eliminate points within a distance from the nearest neighboring atoms as
-                given by `min_distance`
-            3. Initialize neighbor environment using `get_neighbors`
-            4. Shift interstitial candidates to the nearest symmetric points with respect to the
-                neighboring atom sites/vertices.
-            5. Kick out points with large neighbor distance variances; this eliminates "irregular"
-                shaped interstitials
-            6. Cluster interstitial candidates to avoid point overlapping.
-
-        In complex structures (i.e. grain boundary, dislocation etc.), the default parameters
-        should be chosen properly. In order to see other quantities, which potentially
-        characterize interstitial sites, see the following functions from
-        `structure.analyse.interstitials`:
-
-            - `get_variances()`
-            - `get_distances()`
-            - `get_steinhardt_parameters()`
-            - `get_volumes()`
-            - `get_areas()`
-        """
-        self._interstitials = Interstitials(
+        return Interstitials(
             structure=self._structure,
+            num_neighbors=num_neighbors,
             n_gridpoints_per_angstrom=n_gridpoints_per_angstrom,
             min_distance=min_distance,
             use_voronoi=use_voronoi,
-        )
-        return self._interstitials.get_positions(
-            num_neighbors=num_neighbors,
             variance_buffer=variance_buffer,
             n_iterations=n_iterations,
             eps=eps,
         )
+    get_interstitials.__doc__ = Interstitials.__doc__.replace('Class', 'Function') + Interstitials.__init__.__doc__
 
     def get_layers(self, distance_threshold=0.01, id_list=None, wrap_atoms=True, planes=None):
         """
--- a/pyiron_atomistics/atomistics/structure/analyse.py
+++ b/pyiron_atomistics/atomistics/structure/analyse.py
@@ -28,23 +28,20 @@ __date__ = "Sep 1, 2017"
 s = Settings()
 
 
-def get_mean_positions(positions, cell, pbc, labels=None):
+def get_mean_positions(positions, cell, pbc, labels):
     """
-    This function calculates the average position(-s) across periodic boundary conditions. If the
-    labels are specified, the positions are grouped accordingly.
+    This function calculates the average position(-s) across periodic boundary conditions according
+    to the labels
 
     Args:
         positions (numpy.ndarray (n, 3)): Coordinates to be averaged
         cell (numpy.ndarray (3, 3)): Cell dimensions
         pbc (numpy.ndarray (3,)): Periodic boundary conditions (in boolean)
-        labels (numpy.ndarray (n,)): (optional) labels according to which the atoms are
-            grouped
+        labels (numpy.ndarray (n,)): labels according to which the atoms are grouped
 
     Returns:
         (numpy.ndarray): mean positions
     """
-    if labels is None:
-        labels = np.zeros(len(positions)).astype(int)
     _, labels, counts = np.unique(labels, return_inverse=True, return_counts=True)
     mean_positions = positions[np.unique(labels, return_index=True)[1]]
     all_positions = positions-mean_positions[labels]
--- a/pyiron_atomistics/atomistics/structure/analyse.py
+++ b/pyiron_atomistics/atomistics/structure/analyse.py
@@ -91,7 +91,7 @@ class Interstitials:
             neighboring atom sites/vertices.
         5. Kick out points with large neighbor distance variances; this eliminates "irregular"
             shaped interstitials
-        6. Cluster interstitial candidates to avoir point overlapping.
+        6. Cluster interstitial candidates to avoid point overlapping.
 
     The interstitial sites can be obtained through `get_positions`
 
@@ -274,7 +274,7 @@ class Interstitials:
                 neighboring atom sites/vertices.
             5. Kick out points with large neighbor distance variances; this eliminates "irregular"
                 shaped interstitials
-            6. Cluster interstitial candidates to avoir point overlapping.
+            6. Cluster interstitial candidates to avoid point overlapping.
 
         In complex structures (i.e. grain boundary, dislocation etc.), the default parameters
         should be chosen properly. In order to see other quantities, which potentially
@@ -298,7 +298,7 @@ class Interstitials:
         """
         Get variance of neighboring distances. Since interstitial sites are mostly in symmetric
         sites, the variance values tend to be small. In the case of fcc, both tetrahedral and
-        octahedral sites as well as tetrahedral sites in bcc should hvae the value of 0.
+        octahedral sites as well as tetrahedral sites in bcc should have the value of 0.
 
         Returns:
             (numpy.array (n,)) Variance values
@@ -412,7 +412,7 @@ class Analyse:
                 neighboring atom sites/vertices.
             5. Kick out points with large neighbor distance variances; this eliminates "irregular"
                 shaped interstitials
-            6. Cluster interstitial candidates to avoir point overlapping.
+            6. Cluster interstitial candidates to avoid point overlapping.
 
         In complex structures (i.e. grain boundary, dislocation etc.), the default parameters
         should be chosen properly. In order to see other quantities, which potentially
@@ -558,8 +558,8 @@ class Analyse:
                 - total : return number of atoms belonging to each structure
                 - numeric : return a per atom list of numbers- 0 for unknown,
                     1 fcc, 2 hcp, 3 bcc and 4 icosa
-                - str : return a per atom string of sructures
-            ovito_compatibility(bool): use ovito compatiblity mode
+                - str : return a per atom string of structures
+            ovito_compatibility(bool): use ovito compatibility mode
 
         Returns:
             (depends on `mode`)
@@ -588,8 +588,8 @@ class Analyse:
                 - total : return number of atoms belonging to each structure
                 - numeric : return a per atom list of numbers- 0 for unknown,
                     1 fcc, 2 hcp, 3 bcc and 4 icosa
-                - str : return a per atom string of sructures
-            ovito_compatibility(bool): use ovito compatiblity mode
+                - str : return a per atom string of structures
+            ovito_compatibility(bool): use ovito compatibility mode
 
         Returns:
             (depends on `mode`)
@@ -607,7 +607,7 @@ class Analyse:
         Args:
             epsilon (float): displacement to add to avoid wrapping of atoms at borders
             distance_threshold (float): distance below which two vertices are considered as one.
-                Agglomerative clustering algorith (sklearn) is employed. Final positions are given
+                Agglomerative clustering algorithm (sklearn) is employed. Final positions are given
                 as the average positions of clusters.
             width_buffer (float): width of the layer to be added to account for pbc.
 
--- a/pyiron_atomistics/atomistics/structure/analyse.py
+++ b/pyiron_atomistics/atomistics/structure/analyse.py
@@ -42,12 +42,17 @@ def get_mean_positions(positions, cell, pbc, labels):
     Returns:
         (numpy.ndarray): mean positions
     """
+    # Translate labels to integer enumeration (0, 1, 2, ... etc.) and get their counts
     _, labels, counts = np.unique(labels, return_inverse=True, return_counts=True)
+    # Get reference point for each unique label
     mean_positions = positions[np.unique(labels, return_index=True)[1]]
+    # Get displacement vectors from reference points to all other points for the same labels
     all_positions = positions-mean_positions[labels]
+    # Account for pbc
     all_positions = np.einsum('ji,nj->ni', np.linalg.inv(cell), all_positions)
     all_positions -= np.rint(all_positions)[:,pbc]
     all_positions = np.einsum('ji,nj->ni', cell, all_positions)
+    # Add average displacement vector of each label to the reference point
     np.add.at(mean_positions, labels, (all_positions.T/counts[labels]).T)
     return mean_positions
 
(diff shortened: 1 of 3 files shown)


pyiron_atomistics/atomistics/structure/analyse.py
```diff
@@ -26,6 +28,33 @@
...
+        cell (numpy.ndarray (3, 3)): Cell dimensions
+        pbc (numpy.ndarray (3,)): Periodic boundary conditions (in boolean)
+        labels (numpy.ndarray (n,)): (optional) labels according to which the atoms are
+            grouped
+
+    Returns:
+        (numpy.ndarray): mean positions
+
+    This function calculates the average position(-s) across periodic boundary conditions. If the
+    labels are specified, the positions are grouped accordingly.
```
pmrv: Should be above "Args".

pyiron_atomistics/atomistics/structure/analyse.py
```diff
@@ -47,6 +76,234 @@ def get_average_of_unique_labels(labels, values):
...
+    This class internally does the following steps:
+
+        1. Initialize grid points (or Voronoi vertices) which are considered as
+            interstitial site candidates.
+        2. Eliminate points within a distance from the nearest neighboring atoms as
+            given by `min_distance`
+        3. Initialize neighbor environment using `get_neighbors`
+        4. Shift interstitial candidates to the nearest symmetric points with respect to the
+            neighboring atom sites/vertices.
+        5. Kick out points with large neighbor distance variances
```
pmrv: Should explain a bit more our intent on why large distance variances are bad, maybe

```suggestion
        5. Kick out points with large neighbor distance variances; this eliminates "irregular" shaped interstitials
```

pyiron_atomistics/atomistics/structure/analyse.py
```diff
@@ -47,6 +76,234 @@ def get_average_of_unique_labels(labels, values):
...
+    def num_neighbors(self, n):
+        self.reset()
+        self._num_neighbors = n
+
+    def reset(self):
+        self._hull = None
+        self._neigh = None
+
+    @property
+    def neigh(self):
```
pmrv, line 201: Short docstring for the properties would be helpful, like "neigh" is the neighborhood of interstitial candidate site, "positions" are their positions, not atomic ones, etc.  It makes sense once you've seen/used the code, but probably not the first time round.

pyiron_atomistics/atomistics/structure/analyse.py
```diff
@@ -47,6 +76,234 @@ def get_average_of_unique_labels(labels, values):
...
+        if min_distance > 0:
+            self._remove_too_close()
+
+    @property
+    def num_neighbors(self):
+        return self._num_neighbors
+
+    @num_neighbors.setter
+    def num_neighbors(self, n):
+        self.reset()
```
pmrv, line 193: Not defined?

pyiron_atomistics/atomistics/structure/analyse.py
```diff
@@ -47,6 +76,234 @@ def get_average_of_unique_labels(labels, values):
...
+    @property
+    def hull(self):
+        if self._hull is None:
+            self._hull = [ConvexHull(v) for v in self.neigh.vecs]
+        return self._hull
+
+    def _create_gridpoints(self, n_gridpoints_per_angstrom=5):
+        cell = self.structure.cell.diagonal()
+        gridpoints = (n_gridpoints_per_angstrom*cell).astype(int)
+        positions = [np.linspace(0, cell[i], gridpoints[i], endpoint=False) for i in range(3)]
```
pmrv: Would that break for non-cubic cells?  What about

```python
i = np.arange(grid_points)
positions = np.einsum('ij,j... -> i...', self.structure.cell.array, np.meshgrid(i, i, i))
```

(I hope this is how einsum works)

pyiron_atomistics/atomistics/structure/analyse.py
```diff
@@ -47,6 +76,234 @@ def get_average_of_unique_labels(labels, values):
...
+            eps, return_indices=True, positions=self.positions
+        )
+        labels = DBSCAN(eps=eps, min_samples=1).fit_predict(extended_positions)
+        coo = coo_matrix((labels, (np.arange(len(labels)), indices)))
+        labels = coo.max(axis=0).toarray().flatten()
+        self.positions = get_mean_positions(
+            self.positions, self.structure.cell, self.structure.pbc, labels
+        )
+
+    def get_interstitials(
```
pmrv: UI-wise `interstitials.get_interstitials()` feels a bit clunky, what about

```python
... = structure.analyse.iterstitials.get_positions()
```

or 

```python
structure.analyse.interstitials.calculate()
... = structure.analyse.interstitials.positions
```

pmrv: Purely cosmetic, though, I won't insist.

pyiron_atomistics/atomistics/structure/analyse.py
```diff
@@ -56,6 +313,93 @@ def __init__(self, structure):
             structure (:class:`pyiron.atomistics.structure.atoms.Atoms`): reference Atom structure.
         """
         self._structure = structure
+        self._interstitials = None
+
+    @property
+    def interstitials(self):
+        if self._interstitials is None:
+            raise AssertionError('Initialize interstitials via get_interstitials()')
+        return self._interstitials
```
pmrv: In line with my cosmetic comment above
```suggestion
        if self._interstitials is None:
            self._interstitials = Interstitials(…)
        return self._interstitials
```

I wouldn't really mind dropping this and forcing people to make an instance first and then use, like we do with the neighborhood.

pyiron_atomistics/atomistics/structure/atoms.py
```diff
@@ -1185,7 +1185,7 @@ def scaled_pos_xyz(self):
         xyz = self.get_scaled_positions(wrap=False)
         return xyz[:, 0], xyz[:, 1], xyz[:, 2]
 
-    def get_extended_positions(self, width, return_indices=False, norm_order=2):
+    def get_extended_positions(self, width, return_indices=False, norm_order=2, positions=None):
```
pmrv, line 1208: This feels more and more like a general utility for our analyse functions.  Do we expect people to call this manually?  If not I'd move it away from `Atoms` and into a non-public helper function in the analyse module.  Tell-tale sign for me is that the only thing this method needs from its instance is `positions` and `cells` and one of them can even be passed over by an argument.

tests/atomistics/structure/test_analyse.py
```diff
@@ -67,6 +68,46 @@ def test_get_voronoi_vertices(self):
...
+        x_octa_ref = fcc.get_wrapped_coordinates(x_octa_ref)
+        x_octa = fcc.analyse.get_interstitials(num_neighbors=6)
+        self.assertEqual(len(x_octa), len(x_octa_ref))
+        self.assertAlmostEqual(
+            np.linalg.norm(x_octa_ref[:,None,:]-x_octa[None,:,:], axis=-1).min(axis=0).sum(), 0
+        )
+        self.assertTrue(np.all(fcc.analyse.interstitials.get_area()>0))
+        self.assertTrue(np.all(fcc.analyse.interstitials.get_volume()>0))
+        self.assertTrue(np.all(fcc.analyse.interstitials.get_distance()>0))
+        self.assertTrue(np.all(fcc.analyse.interstitials.get_steinhardt_parameter(4)>0))
```
pmrv: Does this test anything or is it just for coverage.  If it is not too difficult I would compare to the real values, like the octahedron volume is $4/3 a_0^3$ (if I didn't make a mistake just now).

pmrv (changes requested):
Only thing left is the user interface, I'd prefer something more closer to `get_neighbors()`, everything else looks great!

pyiron_atomistics/atomistics/structure/analyse.py
samwaseda: uh that's a super good point

pyiron_atomistics/atomistics/structure/analyse.py
samwaseda: I like the suggestion

pyiron_atomistics/atomistics/structure/analyse.py
samwaseda: Here in this case we have to give a number for `num_neighbors`. I'm not really sure if we should assume a number...

pyiron_atomistics/atomistics/structure/atoms.py
samwaseda: Actually we need it the most in `get_neighbors`. I guess the most elegant solution would be to derive `Interstitials` from `Atoms`, which I also thought about. The biggest problem however is that we don't have a proper definition for vacancies. Maybe it's time to raise this issue because we've talked about it some years ago with @liamhuber @jan-janssen @sudarsan-surendralal etc.

pyiron_atomistics/atomistics/structure/analyse.py
```diff
@@ -4,12 +4,14 @@
 
 import numpy as np
 from pyiron_base import Settings
-from sklearn.cluster import AgglomerativeClustering  # TODO: Handle sklearn in dependencies or wrap in import warning
+# TODO: Handle sklearn in dependencies or wrap in import warning
```
liamhuber: Please TODONE before merging. Should  be a straightforward application of the import alarm.

pyiron_atomistics/atomistics/structure/analyse.py
```diff
@@ -26,6 +28,33 @@
...
+        positions (numpy.ndarray (n, 3)): Coordinates to be averaged
+        cell (numpy.ndarray (3, 3)): Cell dimensions
+        pbc (numpy.ndarray (3,)): Periodic boundary conditions (in boolean)
+        labels (numpy.ndarray (n,)): (optional) labels according to which the atoms are
+            grouped
+
+    Returns:
+        (numpy.ndarray): mean positions
+    """
+    if labels is None:
```
liamhuber:
```suggestion
    labels = labels if labels is not None else np.zeros(len(positions)).astype(int)
```

liamhuber: I've taken a liking to one-lining these "if None set default" conditions. How do you feel about this?

pyiron_atomistics/atomistics/structure/analyse.py
```diff
@@ -26,6 +28,33 @@
 s = Settings()
 
 
+def get_mean_positions(positions, cell, pbc, labels=None):
+    """
+    This function calculates the average position(-s) across periodic boundary conditions. If the
```
liamhuber: It's not clear to me what this means. Is this supposed to be a description of the center of mass position, but where all atoms are assumed to have the same mass? If we're averaging "across" PBCs and the PBCs are a 3-tuple of booleans, does that mean our (n,3) positions are getting transformed to something (n,)-shaped?

pyiron_atomistics/atomistics/structure/analyse.py
```diff
@@ -26,6 +28,33 @@
...
+    Returns:
+        (numpy.ndarray): mean positions
+    """
+    if labels is None:
+        labels = np.zeros(len(positions)).astype(int)
+    _, labels, counts = np.unique(labels, return_inverse=True, return_counts=True)
+    mean_positions = positions[np.unique(labels, return_index=True)[1]]
+    all_positions = positions-mean_positions[labels]
+    all_positions = np.einsum('ji,nj->ni', np.linalg.inv(cell), all_positions)
+    all_positions -= np.rint(all_positions)[:,pbc]
```
liamhuber: If I understand these couple of lines you transform to relative coordinates, subtract off the integer bits to center all the positions around (0, 0, 0), and then transform back to absolute space. Does this still do what you intend if one or more positions lies more than one unit cell outside the unit cell?

pyiron_atomistics/atomistics/structure/analyse.py
```diff
@@ -26,6 +28,33 @@
 s = Settings()
 
 
+def get_mean_positions(positions, cell, pbc, labels=None):
```
liamhuber: The only place this is used is in the same module hidden behind a private method; I lean towards making functions private unless they are already expected to be used elsewhere -- is it fair to make this private?

pyiron_atomistics/atomistics/structure/analyse.py
```diff
@@ -47,6 +76,272 @@ def get_average_of_unique_labels(labels, values):
         return mean_values.flatten()
     return mean_values
 
+class Interstitials:
+    """
+    Class to analyse interstitial sites
```
liamhuber:
```suggestion
    Class to locate interstitial sites
```

pyiron_atomistics/atomistics/structure/analyse.py
```diff
@@ -47,6 +76,272 @@ def get_average_of_unique_labels(labels, values):
...
+        """
+        Number of atoms (vertices) to consider for each interstitial atom. By definition,
+        tetrahedral sites should have 4 and octahedral sites 6.
+        """
+        return self._num_neighbors
+
+    @num_neighbors.setter
+    def num_neighbors(self, n):
+        self.reset()
+        self._num_neighbors = n
```
liamhuber, line 194: pycharm nit: define `_num_neighbors` in init

pyiron_atomistics/atomistics/structure/analyse.py
```diff
@@ -47,6 +76,272 @@ def get_average_of_unique_labels(labels, values):
...
+
+    def reset(self):
+        self._hull = None
```
liamhuber, line 197: pycharm nit: define `_hull` in init

pyiron_atomistics/atomistics/structure/analyse.py
```diff
@@ -47,6 +76,272 @@ def get_average_of_unique_labels(labels, values):
...
+        self._neigh = None
```
liamhuber, line 198: pycharm nit: define `_neigh` in init

pyiron_atomistics/atomistics/structure/analyse.py
```diff
@@ -47,6 +76,272 @@ def get_average_of_unique_labels(labels, values):
...
+        ```
+        >>> positions = neigh.positions
+        >>> positions[0][0] = x
+        >>> neigh.positions = positions
+        ```
+
+        This is because in the first case related properties (most importantly the neighborhood
+        information) is not updated, which might lead to inconsistencies.
+        """
+        return self._positions
```
liamhuber, line 236: Not a pycharm nit, but I don't understand why not: define `_positions` in init

pyiron_atomistics/atomistics/structure/analyse.py
```diff
@@ -47,6 +76,272 @@ def get_average_of_unique_labels(labels, values):
...
+        gridpoints = (n_gridpoints_per_angstrom*cell).astype(int)
+        positions = [np.linspace(0, 1, gridpoints[i], endpoint=False) for i in range(3)]
+        positions = np.meshgrid(*positions)
+        return np.einsum(
+            'ji,nj->ni', self.structure.cell, np.stack(positions, axis=-1).reshape(-1, 3)
+        )
+
+    def _remove_too_close(self):
+        neigh = self.structure.get_neighborhood(self.positions, num_neighbors=1)
+        self.positions = self.positions[neigh.distances.flatten()>self.min_distance]
```
liamhuber:
```suggestion
        self.positions = self.positions[neigh.distances.flatten() > self.min_distance]
```

liamhuber: PEP8. I won't mention other PEP8 violations from here on, but the PR could use a linting commit.

pyiron_atomistics/atomistics/structure/analyse.py
```diff
@@ -47,6 +76,272 @@ def get_average_of_unique_labels(labels, values):
...
+        ```
+
+        GOOD:
+        ```
+        >>> positions = neigh.positions
+        >>> positions[0][0] = x
+        >>> neigh.positions = positions
+        ```
+
+        This is because in the first case related properties (most importantly the neighborhood
```
liamhuber, line 233: I have no objection to this in the short run, and I find the documentation very clear.

In the long run, I bet we can get a hook onto the setattr/setitem of positions to trigger the update when a component is modified. I'm thinking through a new array-like class, but there might be a more clever way to do it.

pyiron_atomistics/atomistics/structure/analyse.py
```diff
@@ -47,6 +76,272 @@ def get_average_of_unique_labels(labels, values):
...
+        """
+        if self._hull is None:
+            self._hull = [ConvexHull(v) for v in self.neigh.vecs]
+        return self._hull
+
+    def _create_gridpoints(self, n_gridpoints_per_angstrom=5):
+        cell = self.structure.get_vertical_length()
+        gridpoints = (n_gridpoints_per_angstrom*cell).astype(int)
+        positions = [np.linspace(0, 1, gridpoints[i], endpoint=False) for i in range(3)]
+        positions = np.meshgrid(*positions)
```
liamhuber: I'd just one-line the entire definition of `positions` here by combining this and above. (Separate definition of `gridpoints` does help with readability though, I like that)

pyiron_atomistics/atomistics/structure/analyse.py
```diff
@@ -47,6 +76,272 @@ def get_average_of_unique_labels(labels, values):
...
+        """
+        Args:
+            l (int/numpy.array): Order of Steinhardt parameter
+
+        Returns:
+            (numpy.array (n,)) Steinhardt parameter values
+        """
+        return self.neigh.get_steinhardt_parameter(l=l)
+
+    def get_volume(self):
```
liamhuber: For `get` methods without any arguments, I would rather see a property

pyiron_atomistics/atomistics/structure/analyse.py
```diff
@@ -47,6 +76,272 @@ def get_average_of_unique_labels(labels, values):
...
+        """
+        Returns:
+            (numpy.array (n,)): Convex hull volume of each site.
+        """
+        return np.array([h.volume for h in self.hull])
+
+    def get_area(self):
```
liamhuber: Also `@property`able

pyiron_atomistics/atomistics/structure/analyse.py
```diff
@@ -47,6 +76,272 @@ def get_average_of_unique_labels(labels, values):
...
+            - `get_area()`
+        """
+        self.num_neighbors = num_neighbors
+        for _ in range(n_iterations):
+            self._set_interstitials_to_high_symmetry_points()
+        self._kick_out_points(variance_buffer=variance_buffer)
+        self._cluster_points(eps=eps)
+        return self.positions
+
+    def get_variance(self):
```
liamhuber: Aha, now that I look back this could also be a property

pyiron_atomistics/atomistics/structure/analyse.py
liamhuber: One solution would be to make `num_neighbors` an initialization argument for the class. On this line you could also make `variance_buffer` etc class variables and let `positions` also be a property.

Looking back over the `positions` setter/getter alongside `get_positions`, I think I would be more comfortable if `positions` was a read-only property that did the full jazz found in `get_positions` when called -- i.e. make the current `get_positions` arguments class attributes, make that method private, cache it for faster return, and have `positions` call `_get_positions` under the hood passing in the class attributes (so the cache works?).

Take that with a serious grain of salt though, since I don't currently understand why you would want to give people write access to `positions` -- but you really explicitly *do* do that, so I suspect I'm still missing something pretty fundamental here.

pyiron_atomistics/atomistics/structure/atoms.py
```diff
@@ -1185,7 +1185,24 @@ def scaled_pos_xyz(self):
         xyz = self.get_scaled_positions(wrap=False)
         return xyz[:, 0], xyz[:, 1], xyz[:, 2]
 
-    def get_extended_positions(self, width, return_indices=False, norm_order=2):
+    def get_vertical_length(self, norm_order=2):
+        """
+        Return the length of the cell in each direction projected on the vector vertical to the
+        plane.
+
```
liamhuber, line 1192: Needs `Args:`

pyiron_atomistics/atomistics/structure/atoms.py
liamhuber: I don't object to some `Interstitials(Atoms)`, but we would just need to make sure we're super explicit that the class is only sensible for crystalline `Atoms` and not molecular/liquid/gassy/glassy materials.

pyiron_atomistics/atomistics/structure/analyse.py
```diff
@@ -56,6 +351,94 @@ def __init__(self, structure):
...
+        n_gridpoints_per_angstrom=5,
+        min_distance=1,
+        use_voronoi=False,
+        variance_buffer=0.01,
+        n_iterations=2,
+        eps=0.1,
+        initialize_only=False
+    ):
+        """
+        Get potential interstitial positions
```
liamhuber: (Nearly) duplicate docstring. Please play some `__doc__` and/or `wraps`  fanciness so we only need to keep it up to date in one place.

tests/atomistics/structure/test_analyse.py
```diff
@@ -67,6 +68,46 @@ def test_get_voronoi_vertices(self):
         self.assertEqual(len(basis.analyse.get_voronoi_vertices()), 12)
         self.assertEqual(len(basis.analyse.get_voronoi_vertices(distance_threshold=2)), 1)
 
+    def test_get_interstitials_bcc(self):
```
liamhuber, line 71: Please PEP8 lint, and some `msg=` in the tests would be helpful for skimming through the tests suite/keeping things up-to-date going forward.

tests/atomistics/structure/test_analyse.py
```diff
@@ -67,6 +68,46 @@ def test_get_voronoi_vertices(self):
...
+        bcc = StructureFactory().ase_bulk('Fe', cubic=True)
```
liamhuber:
```suggestion
        bcc = StructureFactory().ase.bulk('Fe', cubic=True)
```

tests/atomistics/structure/test_analyse.py
```diff
@@ -67,6 +68,46 @@ def test_get_voronoi_vertices(self):
...
+        )
+        x_tetra = bcc.analyse.get_interstitials(num_neighbors=4)
+        x_tetra_ref = bcc.get_wrapped_coordinates(bcc.analyse.get_voronoi_vertices())
+        self.assertEqual(len(x_tetra), len(x_tetra_ref))
+        self.assertAlmostEqual(
+            np.linalg.norm(x_tetra_ref[:,None,:]-x_tetra[None,:,:], axis=-1).min(axis=0).sum(), 0
+        )
+
+    def test_get_interstitials_fcc(self):
+        fcc = StructureFactory().ase_bulk('Al', cubic=True)
```
liamhuber:
```suggestion
        fcc = StructureFactory().ase.bulk('Al', cubic=True)
```

pyiron_atomistics/atomistics/structure/analyse.py
```diff
@@ -47,6 +76,272 @@ def get_average_of_unique_labels(labels, values):
...
+    @property
+    def positions(self):
+        """
+        Positions of the interstitial candidates (and not those of the atoms).
+
+        IMPORTANT: Do not set positions via numpy setter, i.e.
+
+        BAD:
+        ```
+        >>> neigh.positions[0][0] = x
```
liamhuber: Wait, why is it `neigh` in this example? Isn't the root object supposed to be an instance of `Interstitials`?

pyiron_atomistics/atomistics/structure/analyse.py
liamhuber: Ok, I had a couple false starts to commenting on this issue, but I think I understand now what's happening. I agree with Marvin.

Things I don't like right now:
- The discrepancy between `Atoms.get_neighbors` and `Analyze.get_interstitials`
- `Analyze.get_interstitials` returns two different class types (depending on a boolean flag) 
  - and by default it's *not* the type `Interstitials`(?!?!?!)
  - And then after it's been called `Analyze.get_interstitials` and `Analyze.interstitials` return different things
- I feel like this compounds on top of the already confusing interplay of `Interstitials.positions` vs `Interstitials.get_positions`

pyiron_atomistics/atomistics/structure/analyse.py
```diff
@@ -47,6 +76,272 @@ def get_average_of_unique_labels(labels, values):
...
+            eps, return_indices=True, positions=self.positions
+        )
+        labels = DBSCAN(eps=eps, min_samples=1).fit_predict(extended_positions)
+        coo = coo_matrix((labels, (np.arange(len(labels)), indices)))
+        labels = coo.max(axis=0).toarray().flatten()
+        self.positions = get_mean_positions(
+            self.positions, self.structure.cell, self.structure.pbc, labels
+        )
+
+    def get_positions(
```
liamhuber: Looking over the `positions` setter/getter alongside `get_positions`, I think I would be more comfortable if `positions` was a read-only property that did the full jazz found in `get_positions` when called -- i.e. make the current `get_positions` arguments class attributes, make that method private, cache it for faster return, and have `positions` call `_get_positions` under the hood passing in the class attributes (so the cache works?).

Take that with a serious grain of salt though, since I don't currently understand why you would want to give people write access to `positions` -- but you really explicitly *do* do that, so I suspect I'm still missing something pretty fundamental here.

liamhuber (changes requested):
Super duper functionality!

At the highest level, I would also like to suggest that we split `Interstitials` off into its own module, e.g. `.../analyze/analyze.py` and `.../analyze/interstitials.py`. First, `analyze.py` is getting quite long IMO, second I think we will wind up needing even more submodules for analysis in the future so why not get started now, and third there is a very real possibility that we will later rebase `Interstitials(object)` to `Interstitials(Atoms)` and then having it separated out from the analysis module itself will be helpful.

At a slightly lower level I share Marvin's concern about the interface via `Analyze` and at a much lower level have a number of requests and nits.

liamhuber:
> At the highest level, I would also like to suggest that we split `Interstitials` off into its own module

I forgot to say: I think we can do this in a separate PR *after* this gets merged. I just wanted to bring it up already since it's a pretty straightforward refactor once this PR is in.

pyiron_atomistics/atomistics/structure/analyse.py
samwaseda: Actually I didn't really understand what this TODO means... (I didn't write it :D)

pyiron_atomistics/atomistics/structure/analyse.py
samwaseda: Hm in this case I might prefer the current version. If you absolutely prefer your version I can also change it.

pyiron_atomistics/atomistics/structure/analyse.py
samwaseda: Actually it's true that this function makes very little sense with `labels=None`. I'm gonna change it.

pyiron_atomistics/atomistics/structure/analyse.py
samwaseda: So following the comment below, `labels=None` is anyway now removed, so I close this thread.

pyiron_atomistics/atomistics/structure/analyse.py
samwaseda: I added strings explaining what these lines do. I hope now it's easy to understand.

pyiron_atomistics/atomistics/structure/analyse.py
samwaseda: So I made it public intentionally, because whenever you have a routine of cluster -> get cluster centres, you'll need this function. For example, imagine a system with precipitates distributed in a box. If you want to have the average position of each precipitate you can label the atoms accordingly and apply this function. This being said, maybe it's not a particularly good idea to make it neither private nor a member function of any class.

pyiron_atomistics/atomistics/structure/analyse.py
samwaseda: Yeah that's the thing. I'm not really sure if we can claim that this class is really able to **locate** the interstitial sites. But ok I'm gonna change it anyway :)

pyiron_atomistics/atomistics/structure/analyse.py
samwaseda: Yeah since we had the same problem in `Atoms` it might be worth spending some time to figure it out.

pyiron_atomistics/atomistics/structure/analyse.py
samwaseda: I thought about it as well, but then I decided to use `get_` anyway, because all properties related to characterising interstitial sites are obtained via `get_`. And in the current implementation it's difficult to live without them. So I guess I'd rather keep the current one.

In light of `Interstitials(Atoms)`, I'm gonna change this maybe to plural to avoid conflicts...

pyiron_atomistics/atomistics/structure/analyse.py
liamhuber: Ok, I think I understand better. It seems like something that might nicely live directly on analyze. Like 'strict.analyze.get_mean_positions(labels=my_id_subset)'

With that in mind what I would really want is an option to weight by mass so we can readily get com

pyiron_atomistics/atomistics/structure/analyse.py
liamhuber: Aha, indeed you didn't! I looked at the blame and it turns out that I slapped it on last time I went through and linted this file. But there's a "blame previous to this commit" button on github and I saw that you added the original import so I'm still going to stick you with the TODO 😝 

cf. [ImportAlarm](https://github.com/pyiron/pyiron_base/blob/b6984ef894e7ed3b59517eacf5697e8b0f9cc710/pyiron_base/generic/util.py#L238). @pmrv wrote some great documentation for it, but if anything remains unclear ping him.

pyiron_atomistics/atomistics/structure/analyse.py
liamhuber: Ugh, I 'unresolved' and it deleted my comment draft >.<

I see now we're coming from exactly the same perspective though, and I totally agree "locate" is too strong. This was also my concern with "analyze" which I felt really implied that we had already found them and they were real/correct.

What about `attempt to locate` or `search for`?

pyiron_atomistics/atomistics/structure/analyse.py
samwaseda: Because I didn't pay attention :D

pyiron_atomistics/atomistics/structure/analyse.py
samwaseda: I'll leave com for now because I don't have a good test case and adding com will not change the current functionality anyway.

pyiron_atomistics/atomistics/structure/analyse.py
samwaseda: I don't really understand what you guys mean with the discrepancy between `Atoms.get_neighbors` and `Analyse.get_interstitials`, because `get_neighbors` gives the information on the neighbourhood for each atom (i.e. `n_atoms x n_neighbors`), while `get_interstitials` is about giving interstitial positions. In total, I would rather expect `Interstitials` to be like `Atoms` but not like `Neighbors`.

samwaseda: Oh you mean like `get_neighbors` returns `Neighbors` and `get_interstitials` doesn't return `Interstitials`? If that's the case I agree that it's maybe something that has to be changed.

liamhuber: `Atoms.get_neighbors` returns type `Neighbors`^. `Analyze.get_interstitials` returns type `np.ndarray`^^. This type discrepancy is my primary objection. I understand this to be Marvin's objection as well; that use should be 

```python
inter = some_struct.analyze.get_interstitials(**kwargs)
my_pos = inter.get_positions()
stein4 = inter.get_steinhardt_parameters(4)
```

etc. And not:

```python
my_pos  = some_struct.analyze.get_insterstitials(**kwargs)
stein4 = some_struct.analyze.interstitials.get_steinhardt_parameters(4)
```

> In total, I would rather expect Interstitials to be like Atoms but not like Neighbors.

I am OK with this sentiment, and I suspect Marvin is too. The trouble I see is that right now, with access via `Analyse` (sorry, just realized I've been applying wrong spelling), `Interstitials` behaves like `Atoms.positions` not `Atoms`...

^ By default; there is also a flag to return some `Tree` object?
^^ Also by default, we do here have a flag to force the return of 
I'm not really thrilled about returning totally different types based on an argument flag, but (a) I recognize there is room on that front for disagreement, and (b) I think it's tangential to this PR and we don't need to resolve it.

pyiron_atomistics/atomistics/structure/analyse.py
liamhuber: Yes, the comment lines definitely make it easier to follow.

My question stands though about whether `rint` is a robust solution? In the case that the positions are not too far outside the home cell I see that it is a nice solution, but it seems to me like it breaks if the positions are too spread out. I'm not 100% sure though, nor do I know if you care. On solution might be to quickly decompose it into an extra line and variable and then simply throw an error if `np.any(np.abs(rint_array) >= 1)`.

pyiron_atomistics/atomistics/structure/analyse.py
liamhuber: right, and actually we can't currently slap it as an `Analyse` method because it is needed to create the `Interstitials` and these don't (currently) inherit from `Atoms` so they don't have their own `analyse` property via which to access it 😝

pyiron_atomistics/atomistics/structure/analyse.py
samwaseda: I see your point. Indeed if the point cloud is larger than half the simulation box, it might break down. I share the problem in the figure below. In particular, the average position might depend on the reference point (from which all arrows are drawn). I'm not really sure if there's a good way to prevent this from happening from the beginning, except for calculating all possible average positions by taking all possible reference points. And I'm not really sure whether we should raise an error if the largest distance is close to half a box, because I don't know how to define "close".

Other than that, it `rint` should be a solid method whether certain positions are far outside the box or not, as far as I can imagine at least.

Screenshot 2021-06-08 at 10 25 16

pyiron_atomistics/atomistics/structure/analyse.py
samwaseda: @pmrv help me I don't know how to use it XD

pyiron_atomistics/atomistics/structure/analyse.py
liamhuber: That is a super good sketch of the issue.

Is this not a case where you want the MIC atom relative to your reference point? In this case ASE's `find_mic` might do the trick more robustly instead of what you have here. I might still misunderstand the problem though. Geometry is always surprisingly difficult 😭

samwaseda: Yeah ASE will have essentially the same problem. Depending on what you consider as the reference point, certain positions could be lying across different boundaries. So in the example above, I took a reference point of a position on the right hand side and calculate the average position, but if I take one of the positions on the left side (close to the border) as the reference point, I'll get a different average position.

liamhuber: > So in the example above, I took a reference point of a position on the right hand side and calculate the average position, but if I take one of the positions on the left side (close to the border) as the reference point, I'll get a different average position.

Yes, but not if you wrap the result back into the home cell. Using `find_mic` with the reference point as the atom shown, I get the three atoms in at the right side of the cell, and three ghost atoms outside it -- breaking that long arrow to the left. Alternatively, suppose I choose that very atom at the head of the long arrow as my reference site. Then applying `find_mic` to the other atoms I will get the three atoms on the left side of the cell, plus three more ghost atoms to the left outside it. The centres of these two clusters will differ by Lx. I think that final cell length offset is unavoidable and maybe you want to wrap back into the home cell for the final result.

pyiron_atomistics/atomistics/structure/analyse.py
liamhuber: > Oh you mean like get_neighbors returns Neighbors and get_interstitials doesn't return Interstitials? If that's the case I agree that it's maybe something that has to be changed.

Sorry, missed this second comment during my initial reply. Yep! Cool, I think we're on a sufficiently similar page then.

pyiron_atomistics/atomistics/structure/analyse.py
samwaseda: Okay I have to admit that the figure above was quite useless XD. Here's my second try: Imagine a cluster of spheres below. If I take the green sphere as the reference point, I'd take the PBC into account (I put opacity to represent PBC) and the cluster center will be more or less where the green atom is, right? But if I take the red sphere as the reference point, the displacement vectors will be different, and the cluster center will probably be more or less where the red sphere is.

But it's true that regardless of this discussion indeed I should anyway use `find_mic` because the lines above do exactly the same thing.

Screenshot 2021-06-08 at 16 29 23

liamhuber: Ok, yeah, I see your point now. I guess what you really need is a loop over available image sites and finally choose the first moment that comes from a set of sites that minimize the second moment?

samwaseda: Yes, but I guess for now I'm gonna keep the current version because if I have to minimise the second moment it's going to be computationally expensive.

pyiron_atomistics/atomistics/structure/analyse.py
pmrv: Let's just directly import it and add it to deps.

PR #220, opened 2021-06-01
VASP - change NSW if interactive

raynol-dsouza:
Modify the `NSW` flag to `1000` if VASP is run interactively.

### DIFF
--- a/pyiron_atomistics/vasp/interactive.py
+++ b/pyiron_atomistics/vasp/interactive.py
@@ -295,7 +295,7 @@ class VaspInteractive(VaspBase, GenericInteractive):
             self._check_incar_parameter(parameter="INTERACTIVE", value=True)
             self._check_incar_parameter(parameter="IBRION", value=-1)
             self._check_incar_parameter(parameter="POTIM", value=0.0)
-            self._check_incar_parameter(parameter="NSW", value=1000)
+            self._check_incar_parameter(parameter="NSW", value=np.iinfo(np.int32).max - 1)
             self._check_incar_parameter(parameter="ISYM", value=0)
         super(VaspInteractive, self)._run_if_new(debug=debug)
 
--- a/pyiron_atomistics/vasp/interactive.py
+++ b/pyiron_atomistics/vasp/interactive.py
@@ -277,6 +277,9 @@ class VaspInteractive(VaspBase, GenericInteractive):
     def _check_incar_parameter(self, parameter, value):
         if parameter not in self.input.incar._dataset["Parameter"]:
             self.input.incar[parameter] = value
+        if parameter == "NSW":
+            self._logger.debug("setting NSW to {}".format(value))
+            self.input.incar["NSW"] = value
 
     def _interactive_check_output(self):
         while self._interactive_library.poll() is None:


liamhuber:
I don't see how the commit does what the title suggests?

raynol-dsouza:
@liamhuber the check for interactive is done in `_run_if_new()`, which in turn calls `_check_incar_parameters()` where the change is made.

liamhuber:
> @liamhuber the check for interactive is done in `_run_if_new()`, which in turn calls `_check_incar_parameters()` where the change is made.

Doesn't the number 1000 still need to appear somewhere, or is this just setting the interactive NSW value to the default (calc static? calc md?) NSW value?

raynol-dsouza:
```
def _run_if_new(self, debug=False):
        if (
            self.server.run_mode.interactive
            or self.server.run_mode.interactive_non_modal
        ):
            self._check_incar_parameter(parameter="INTERACTIVE", value=True)
            self._check_incar_parameter(parameter="IBRION", value=-1)
            self._check_incar_parameter(parameter="POTIM", value=0.0)
            self._check_incar_parameter(parameter="NSW", value=1000)
            self._check_incar_parameter(parameter="ISYM", value=0)
        super(VaspInteractive, self)._run_if_new(debug=debug)
```

I believe it is a hard coded value.

liamhuber:
Hmmm, I have multiple issues here.

If I understand correctly, when the NSW value is not large enough the VASP process terminates before `interactive_close` can be called and the interactive job is not able to collect output data correctly. Currently, the interactive job sets NSW to 1000 *if it is not yet set*. Here, your goal is to *always override* this value with 1000 whether it's set or not.

What I don't like about your solution is that the correction is buried as deep as possible, all the way down in `_check_incar_parameter`. Definitely this can get bumped up to `_run_if_new` by replacing `self._check_incar_parameter(parameter="NSW", value=1000)` with `self.input.incar['NSW'] = 1000`. This has the same effect, but is less opaque. Since a large NSW value is a prerequisite for running VASP interactively, I would like it even better if this line was in `validate_ready_to_run`, since validation of readiness is exactly what this is, but I'm not 100% sure that putting it there won't introduce some bug.

What I do like about your solution is it really moves towards forcing the interactive run to work, regardless of what silly input the user has provided. But then I'm not such a fan of the original code -- why `1000`? What if the user wants to do more than 1000 iterations? An Al unit cell takes only 2s to `calc_static` on the head node, so this is not inconcievable. The current code fixes the problem only in heuristic way, and exposes the user to a surprising bug if (a) they override the NSW value (which you now fix) or (b) if they run for longer than the devs expected (which still remains after this PR).

If jacking up the NSW value to some large integer is really ok, why not go all the way? `self.input.incar['NSW'] = np.iinfo(np.int32).max - 1` just sets it to one less the biggest possible (signed) int32. Putting that as high up the interactive food chain as possible solves the problem completely and with less ambiguity of purpose.

(Note: I originally tried this manually with int64, but that breaks things. The error message was unclear to me, but it obviously must have been some overflow. Dropping to int32 worked fine. I don't think we need to worry anymore about people running pyiron on a 16 bit system, nor about people running more than 2147483646 iterations since that works out to 136 years even with the 2s iteration for an Al one-atom unit cell...)

liamhuber:
This might feel like overkill, but if we're going to squash this bug let's squash it forever! 😈 

only good bug

raynol-dsouza:
Although I do agree with you, @liamhuber, since this is my first time trying to get some jobs run using VASP, I would rather have people who use VASP more frequently take a call on this.

liamhuber:
After looking at your issue page, I remembered that both Dominik and Hari ran into exactly this problem with QM/MM, so I'm glad you took the time to figure out where it needs to be addressed!

raynol-dsouza:
@liamhuber I only ran into the bug! It was @jan-janssen who isolated where the bug was. :)

liamhuber:
> @liamhuber I only ran into the bug! It was @jan-janssen who isolated where the bug was. :)

No no Raynol, you're supposed to take credit for your excellent management work in delegating the problem 😆 

But seriously, I'm glad you guys tracked it down, regardless who did the bug finding and who did the line finding 👍

raynol-dsouza:
@liamhuber should this PR be merged? It has already been merged with `vasp_int_conv_check`, which has been merged with master.

liamhuber:
> @liamhuber should this PR be merged? It has already been merged with `vasp_int_conv_check`, which has been merged with master.

The `convergence_check` code is definitely conflicting with #222, which should be merged in here. Learning from that PR, my current understand is that you also need a `if self.server.run_mode.interactive:` catch around your NSW modification here, or? At the same time, instead of 1000 let's really use a properly big number (`np.iinfo(np.int32).max - 1`) so we don't need to come re-fix it later.

Ultimately you should probably also get a review from someone who knows the VASP code. The PR itself is only a couple of lines, it's just a matter of being familiar with other pieces it might interact with. @sudarsan-surendralal?

pyiron_atomistics/vasp/interactive.py
```diff
@@ -296,6 +299,16 @@ def _run_if_new(self, debug=False):
             self._check_incar_parameter(parameter="ISYM", value=0)
         super(VaspInteractive, self)._run_if_new(debug=debug)
 
+    def convergence_check(self):
```
sudarsan-surendralal: Merge the master branch into this which updates this function!

sudarsan-surendralal (changes requested):
I agree with @liamhuber, it's better to use something like `np.iinfo(np.int32).max - 1` in `_run_if_new` rather than limiting it to something like 1000.

pyiron_atomistics/vasp/interactive.py
raynol-dsouza: Done.

sudarsan-surendralal (approved):
Looks good to me! Feel free to merge this @raynol-dsouza

PR #222, opened 2021-06-01
Fix to stop convergence checks for interactive jobs!

### LINKED ISSUE #218: Running VASP interactively throws a BrokenPipeError
The following code:

```
job = pr.create.job.Vasp('vasp_job')
job.structure = pr.create.structure.ase.bulk('Al', cubic=True).repeat(2)
job.set_kpoints(mesh=[3,3,3])
job.set_encut(encut=250.)
job.calc_static()
job.server.cores = 3
job.interactive_open()
for i in np.arange(3):
    print(i)
    job.run()
job.interactive_close()
```

throws the following error:

```
BrokenPipeError                           Traceback (most recent call last)
<ipython-input-9-3c3d382797c1> in <module>
      7 for i in np.arange(3):
      8     print(i)
----> 9     job.run()
     10 job.interactive_close()

~/pyiron_repos/pyiron_base/pyiron_base/generic/util.py in decorated(*args, **kwargs)
    211                         stacklevel=2
    212                     )
--> 213             return function(*args, **kwargs)
    214         return decorated
    215 

~/pyiron_repos/pyiron_base/pyiron_base/job/generic.py in run(self, delete_existing_job, repair, debug, run_mode, run_again)
    665                 self._run_if_submitted()
    666             elif status == "running":
--> 667                 self._run_if_running()
    668             elif status == "collect":
    669                 self._run_if_collect()

~/pyiron_repos/pyiron_base/pyiron_base/job/interactive.py in _run_if_running(self)
    163         """
    164         if self.server.run_mode.interactive:
--> 165             self.run_if_interactive()
    166         elif self.server.run_mode.interactive_non_modal:
    167             self.run_if_interactive_non_modal()

~/pyiron_repos/pyiron_atomistics/pyiron_atomistics/vasp/interactive.py in run_if_interactive(self)
    255 
    256     def run_if_interactive(self):
--> 257         self.run_if_interactive_non_modal()
    258         self._interactive_check_output()
    259         self._interactive_vasprun = Outcar()

~/pyiron_repos/pyiron_atomistics/pyiron_atomistics/vasp/interactive.py in run_if_interactive_non_modal(self)
    251                     self._logger.debug("Vasp library: " + text)
    252                     self._interactive_library.stdin.write(text + "\n")
--> 253             self._interactive_library.stdin.flush()
    254         self._interactive_fetch_completed = False
    255 

BrokenPipeError: [Errno 32] Broken pipe
```

Has anyone come across this error before? Is this related to #216?

sudarsan-surendralal:
A temporary fix! Please feel free to modify this if necessary

### DIFF
--- a/pyiron_atomistics/vasp/interactive.py
+++ b/pyiron_atomistics/vasp/interactive.py
@@ -306,6 +306,8 @@ class VaspInteractive(VaspBase, GenericInteractive):
         """
         if self.server.run_mode.interactive:
             return True
+        else:
+            return super(VaspBase).convergence_check()
 
 
 class Output(OutputBase):
--- a/pyiron_atomistics/vasp/interactive.py
+++ b/pyiron_atomistics/vasp/interactive.py
@@ -307,7 +307,7 @@ class VaspInteractive(VaspBase, GenericInteractive):
         if self.server.run_mode.interactive:
             return True
         else:
-            return super(VaspBase).convergence_check()
+            return super().convergence_check()
 
 
 class Output(OutputBase):
--- a/pyiron_atomistics/vasp/interactive.py
+++ b/pyiron_atomistics/vasp/interactive.py
@@ -304,7 +304,8 @@ class VaspInteractive(VaspBase, GenericInteractive):
             bool: Always True
 
         """
-        return True
+        if self.server.run_mode.interactive:
+            return True
 
 
 class Output(OutputBase):
--- a/pyiron_atomistics/vasp/interactive.py
+++ b/pyiron_atomistics/vasp/interactive.py
@@ -296,6 +296,16 @@ class VaspInteractive(VaspBase, GenericInteractive):
             self._check_incar_parameter(parameter="ISYM", value=0)
         super(VaspInteractive, self)._run_if_new(debug=debug)
 
+    def convergence_check(self):
+        """
+        We don't care about convergence for interactive jobs! Always returns True
+
+        Returns:
+            bool: Always True
+
+        """
+        return True
+
 
 class Output(OutputBase):
     """


pmrv:
We could also put this into `GenericInteractive`, no?

pyiron_atomistics/vasp/interactive.py
```diff
@@ -296,6 +296,16 @@ def _run_if_new(self, debug=False):
...
 
+    def convergence_check(self):
+        """
+        We don't care about convergence for interactive jobs! Always returns True
+
+        Returns:
+            bool: Always True
+
+        """
+        return True
```
pmrv: (untested)

```suggestion
        if self.server.run_mode.interactive:
            return True
```

We only want to skip that in interactive jobs, but sadly all our jobs always derive from `Interactive...`

liamhuber: Unfortunately true

sudarsan-surendralal: Done!

liamhuber:
> We could also put this into `GenericInteractive`, no?

If I understand, then we rely entirely on the MRO and if anyone switched `class VaspInteractive(VaspBase, GenericInteractive)` to `class VaspInteractive(GenericInteractive, VaspBase)` we would break the functionality because of all the funky convergence check jazz in `VaspBase`. (Or it starts backwards and we would need to do this now? Honestly I can never remember the order of resolution.) I think it's unlikely that someone would go around messing with the inheritance structure without knowing what they're doing/having a good reason, but it still seems like it adds unnecessary fragility.

liamhuber (approved):
In the long run we need to adjust the inheritance tree so this sort of awkwardness is not necessary, but in the short run I don't see any problems or have any better suggestions!

pmrv:
> > We could also put this into `GenericInteractive`, no?
> 
> If I understand, then we rely entirely on the MRO and if anyone switched `class VaspInteractive(VaspBase, GenericInteractive)` to `class VaspInteractive(GenericInteractive, VaspBase)` we would break the functionality because of all the funky convergence check jazz in `VaspBase`. (Or it starts backwards and we would need to do this now? Honestly I can never remember the order of resolution.) I think it's unlikely that someone would go around messing with the inheritance structure without knowing what they're doing/having a good reason, but it still seems like it adds unnecessary fragility.

Works for me, just wanted to raise the point.

liamhuber:
> Works for me, just wanted to raise the point.

It was a good one. I'm 100% behind the principle of pushing the method as far upstream as it still makes sense (and it really makes sense in `GenericInteractive`, it's just that here the streams get too crossed 😝

PR #235, opened 2021-06-14
interactive_prepare

raynol-dsouza:
The modification/addition of tags to the INCAR file in `VaspInteractive` when `run_mode.interactive = True` is only done when the `run` method is called. This creates an issue when I define a `VaspInteractive` reference job, and then make a copy of it WITHOUT running the reference job -- the copy gets the original, unmodified INCAR file. This PR is a workaround. The user calls a new method, `interactive_prepare()` (could change the name to `interactive_initialize()`), should the user need to copy the reference job without having to run it, so the copy gets the correct INCAR file.

### DIFF
--- a/pyiron_atomistics/vasp/interactive.py
+++ b/pyiron_atomistics/vasp/interactive.py
@@ -292,12 +292,18 @@ class VaspInteractive(VaspBase, GenericInteractive):
             self.server.run_mode.interactive
             or self.server.run_mode.interactive_non_modal
         ):
-            self._check_incar_parameter(parameter="INTERACTIVE", value=True)
-            self._check_incar_parameter(parameter="IBRION", value=-1)
-            self._check_incar_parameter(parameter="POTIM", value=0.0)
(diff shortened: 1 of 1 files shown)


sudarsan-surendralal (approved):
LGTM! Maybe change the function to `prepare_for_interactive()` (optional)

liamhuber (approved):
Now that it's factored out and doesn't require actually running the job (which would take a license) it might also be nice to add a quick check, e.g. on the NSW tag, that it actually does what we want (it does, but with the test it will continue doing).

raynol-dsouza:
> it might also be nice to add a quick check

@liamhuber I went through the `tests` directory in `pyiron_atomistics`, and could not find any tests written for `VaspInteractive` (or for `LammpsInteractive`). I do not know where to add the check.

liamhuber:
> I went through the tests directory in pyiron_atomistics, and could not find any tests written for VaspInteractive (or for LammpsInteractive).

Issue #197, opened 2021-05-16
Elastic tensor and child jobs when submitting


Leimeroth:
Also see updates 1 and 2 below, issues seems to be the storage of pressure in vasp jobs
**Summary**

ElasticTensor jobs fail for me when I submit them to our remote cluster.
I am not really sure but I think this could be related to some other weird behavior which I noticed that affects all AtomisticParallel jobs. When I submitted Murnaghan jobs with old pyiron versions (not sure which ones) each child job was given an id and stored in the local database, then each job was submitted to the cluster. Now only the master job is stored in the local database and submitted to the cluster. This has the side effect that the child_ids attribute is an empty list, maybe this causes the issue?
Murnaghan jobs still work, even if there child_ids is empty, so maybe something else causes the problem.

**pyiron Version and Platform**

Newest pyiron versions from git.

**Steps to Reproduce**

Submit a ElasticTensor job to a remote cluster.

**Error message**
```
  File "/home/nl77pupy/miniconda3/lib/python3.8/runpy.py", line 194, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/home/nl77pupy/miniconda3/lib/python3.8/runpy.py", line 87, in _run_code
    exec(code, run_globals)
  File "/home/nl77pupy/git_projects/pyiron_base/pyiron_base/cli/__main__.py", line 2, in <module>
    main()
  File "/home/nl77pupy/git_projects/pyiron_base/pyiron_base/cli/__init__.py", line 61, in main
    args.cli(args)
  File "/home/nl77pupy/git_projects/pyiron_base/pyiron_base/cli/wrapper.py", line 34, in main
    job_wrapper_function(
  File "/home/nl77pupy/git_projects/pyiron_base/pyiron_base/job/wrapper.py", line 149, in job_wrapper_function
    job.run()
  File "/home/nl77pupy/git_projects/pyiron_base/pyiron_base/job/wrapper.py", line 115, in run
    self.job.run_static()
  File "/home/nl77pupy/git_projects/pyiron_base/pyiron_base/master/parallel.py", line 678, in run_static
    self._run_if_master_modal_child_non_modal(job=job)
  File "/home/nl77pupy/git_projects/pyiron_base/pyiron_base/master/parallel.py", line 657, in _run_if_master_modal_child_non_modal
    self.run()  # self.run_if_collect()
[...]

PR #237, opened 2021-06-15
Enable `use_pressure = False`

samwaseda:
By setting `use_pressure = False`, it ignores the pressures regardless of whether it's available or not.

Problem stated in [this issue](https://github.com/pyiron/pyiron_atomistics/issues/197)


Leimeroth (approved):
Works with vasp now, thanks

PR #239, opened 2021-06-17
Get neighbors

### LINKED ISSUE #232: New mode for `get_neighbors`
Here I want to talk about a small modification in `get_neighbors` and I would like to know what you guys think.

Currently, `get_neighbors` gives different results depending on `allow_ragged = True` or `False`. This essentially concerns when the cutoff radius is specified and the number of neighbours for each atom is not the same (e.g. first atom has 4 neighbours within a radius `r` and the second atom 5 etc.). The two modes handle such a case as follows:

- `allow_ragged = True`: Create a list of numpy arrays with different lengths
- `allow_ragged = False`: Fill missing entries with impossible elements (`np.inf` in `distances`, `vecs`; `natoms+1` for `indices`; and 0 for `shells`).

As it turned out for me, they were both not particularly useful (and also not elegant). Now I'm wondering what you guys think of doing the distinction as follows:

- `mode = 'flattened'`: Return flattened arrays for `distances`, `vecs` and `shells` (where the filled values are removed) and `(N, 2)`-array for `indices` containing the pairs of indices (first is that of the atom in focus and second is that of the neighbour)
- `mode = 'filled'`: Same as `allow_ragged = False`
- `mode = 'ragged'` Same as `allow_ragged = True`

I could have opened a PR directly but I wanted to know your opinions on this before doing so.

samwaseda:
Create a new mode `flattened`.

- 'filled': Fill variables for the missing entries are filled as follows: `np.inf` in `distances`, `numpy.array([np.inf, np.inf, np.inf])` in `vecs`, `n_atoms+1` (or a larger value) in `indices` and -1 in `shells`.
- 'ragged': Create lists of different lengths.
- 'flattened': Return flattened arrays for distances, vecs and shells. The indices corresponding to the row numbers in 'filled' and 'ragged' are in `atom_numbers`

Furthermore, the variables are now stored only in the `filled`-mode. So the conversion takes place whenever you call the variables (which are properties). For `ragged`, it might become somewhat slow (which I don't think would cause a problem though).

The same functionality can be obtained by specifying the mode between the class and the values, e.g. `neigh.filled.distances`, `neigh.ragged.vecs`, `neigh.flattened.indices` etc.

Other main changes:

- `get_numbers_of_neighbors` is removed from `Atoms` (because it's not well modularized). Instead, now you can initialize the `Neighbors`-class and get it via `number_of_neighbors` (which is a property).
- `get_neighbors_by_distance` is now deprecated, because the only differences to `get_neighbors` are the default values (sorry @sudarsan-surendralal that I made so much fuss last time).
- I erased `get_distances`, `get_vectors` and `get_indices`, because it was extremely rare that they were used and I don't think anyone is affected.
- `vecs` is initialised when it's called. No `t_vec` to specify anymore.

Closes #232


pyiron_atomistics/atomistics/structure/neighbors.py
```diff
@@ -182,26 +235,25 @@ def _get_wrapped_positions(self, positions, distance_buffer=1.0e-12):
...
         cell = self._ref_structure.cell
         x_scale = np.dot(x, np.linalg.inv(cell))+distance_buffer
-        x[...,self._ref_structure.pbc] -= np.dot(np.floor(x_scale),
-                                                 cell)[...,self._ref_structure.pbc]
+        x[...,self._ref_structure.pbc] -= np.dot(
+            np.floor(x_scale), cell
+        )[...,self._ref_structure.pbc]
         return x
 
+    @deprecate(allow_ragged="use mode instead.")
```
niklassiemer: Do we really need to deprecate private functions? I lean towards just removing the `allow_ragged` here!

pyiron_atomistics/atomistics/structure/neighbors.py
```diff
@@ -182,26 +235,25 @@ def _get_wrapped_positions(self, positions, distance_buffer=1.0e-12):
...
     def _get_distances_and_indices(
         self,
         positions=None,
+        mode='filled',
         allow_ragged=None,
         num_neighbors=None,
         cutoff_radius=np.inf,
         width_buffer=1.2,
     ):
+        if allow_ragged is not None:
```
niklassiemer: I would then also shift this check to `get_distances` and `get_indices` accordingly.

pyiron_atomistics/atomistics/structure/neighbors.py
```diff
@@ -230,19 +282,18 @@ def _get_distances_and_indices(
...
+    @deprecate(allow_ragged="use mode instead.")
     def get_indices(
         self,
         positions=None,
         allow_ragged=None,
+        mode='filled',
         num_neighbors=None,
         cutoff_radius=np.inf,
         width_buffer=1.2,
+        flatten=False,
```
niklassiemer: flatten is not used here!

pyiron_atomistics/atomistics/structure/neighbors.py
```diff
@@ -338,53 +410,57 @@ def get_vectors(
...
         return self._get_vectors(
             positions=positions,
             allow_ragged=allow_ragged,
+            mode=mode,
             num_neighbors=num_neighbors,
             cutoff_radius=cutoff_radius,
             width_buffer=width_buffer,
         )
 
+    @deprecate(allow_ragged="use mode instead.")
```
niklassiemer: Same here, let's shift the `if allow_ragged` check to `get_vectors` and just update the private function.

niklassiemer (changes requested):
Thanks a lot for the idea! I think this a great idea to `reshape` on the `@property` to switch representation of variables depending on the `mode`! 
One minor thing I would change is how we deprecate/update the functions. I would like to deprecate only functions visible to the user and move the input parameter conversion into these such that we do not deprecate private functions.

pyiron_atomistics/atomistics/structure/neighbors.py
```diff
@@ -70,6 +71,36 @@ def __init__(self, ref_structure):
...
+            if v:
+                return k
+
+    @mode.setter
+    def mode(self, new_mode):
+        if new_mode not in self._mode.keys():
+            raise KeyError(
+                new_mode + " not found. Available modes: 'filled', 'ragged', 'flattened'"
+            )
+        self._mode = dict.fromkeys(list(self._mode.keys()), False)
```
sudarsan-surendralal: Can be simplified as below no?

```suggestion
        self._mode = {key: False for key in self._mode.keys()}
```

pyiron_atomistics/atomistics/structure/neighbors.py
```diff
@@ -99,6 +129,42 @@ def copy(self):
...
+    @property
+    def numbers_of_neighbors(self):
+        """
+        Numbers of neighbors for each atom. If `cutoff` was not specified for the initialization,
+        the same number is shown for all atoms.
+        """
+        return np.sum(self._distances < np.inf, axis=-1)
+
+    @property
+    def distances(self):
```
sudarsan-surendralal, line 141: Needs a one line docstring

pyiron_atomistics/atomistics/structure/neighbors.py
```diff
@@ -99,6 +129,42 @@ def copy(self):
...
+        return self._reshape(self._distances)
+
+    @property
+    def vecs(self):
```
sudarsan-surendralal, line 156: Needs a one line docstring

pyiron_atomistics/atomistics/structure/neighbors.py
```diff
@@ -99,6 +129,42 @@ def copy(self):
...
+        return self._reshape(self._vecs)
+
+    @property
+    def indices(self):
```
sudarsan-surendralal, line 161: Needs a one line docstring

pyiron_atomistics/atomistics/structure/neighbors.py
```diff
@@ -99,6 +129,42 @@ def copy(self):
...
+        return self._reshape(self._indices)
+
+    @property
+    def atom_numbers(self):
```
sudarsan-surendralal, line 166: Needs a one line docstring

sudarsan-surendralal (commented):
LGTM! Just some cosmetic changes suggested.

pyiron_atomistics/atomistics/structure/neighbors.py
```diff
@@ -6,11 +6,11 @@
 from sklearn.cluster import AgglomerativeClustering
 from scipy.sparse import coo_matrix
 from scipy.special import gamma
-from pyiron_base import Settings
 from pyiron_atomistics.atomistics.structure.analyse import get_average_of_unique_labels
 from scipy.spatial.transform import Rotation
 from scipy.special import sph_harm
 import warnings
+from pyiron_base import Settings, deprecate, deprecate_soon
```
sudarsan-surendralal: You don't use `deprecate_soon` anywhere

pyiron_atomistics/atomistics/structure/neighbors.py
```diff
@@ -253,31 +304,45 @@ def get_indices(
...
             positions=positions,
             allow_ragged=allow_ragged,
+            mode=mode,
             num_neighbors=num_neighbors,
             cutoff_radius=cutoff_radius,
             width_buffer=width_buffer,
         )[1]
 
+    @property
+    def numbers_of_neighbors(self):
```
sudarsan-surendralal, line 289: You've defined this twice

sudarsan-surendralal (changes requested):
Oops just noticed some things that need to be changed

pyiron_atomistics/atomistics/structure/neighbors.py
samwaseda: The reason why `allow_ragged` is still here is because `_get_distances_and_indices` is used by multiple functions -> by leaving `allow_ragged` here, the conversion of `allow_ragged` to `mode` has to be done only once.

But actually you might be right that maybe it's not necessary to have `@deprecate` here, since the user should be able to see the warning in an earlier function.

pyiron_atomistics/atomistics/structure/neighbors.py
samwaseda: Oh ok then forget about my comment above :D

pyiron_atomistics/atomistics/structure/neighbors.py
samwaseda: Good job! Before I introduced `mode` I was thinking about using `flatten`. I simply forgot to remove it.

pyiron_atomistics/atomistics/structure/neighbors.py
samwaseda: oops

pyiron_atomistics/atomistics/structure/neighbors.py
```diff
@@ -37,16 +37,12 @@ class Tree:
...
-        only if the cutoff length was specified, in which case the number of atoms for each
-        position could differ. `allow_ragged = False` is computationally more efficient in most of the
-        methods. When `allow_ragged = False`, the variables are filled as follows:
-        `np.inf` in `distances`, `numpy.array([np.inf, np.inf, np.inf])` in `vecs`, `n_atoms+1` (or
-        a larger value) in `indices` and -1 in `shells`
     - wrap_positions (bool): Whether to wrap back the positions entered by user in get_neighborhood
         etc. Since the information outside the original box is limited to a few layer,
         wrap_positions=False might miss some points without issuing an error.
 
+    Change representation mode via `mode` (cf. its DocString)
```
pmrv:
```suggestion
    Change representation mode via :attribute:`.Neighbors.mode` (cf. its DocString)
```

should generate the appropriate link in readthedocs.

pmrv (approved):

sudarsan-surendralal (approved):

samwaseda:
As the option of `t_vec = False` was extremely special, I simply removed it instead of making a deprecation warning.

pyiron_atomistics/atomistics/structure/atoms.py
```diff
@@ -1269,60 +1270,38 @@ def get_numbers_of_neighbors_in_sphere(
...
-                    num_neighbors = 2 * num_neighbors
-        return num_neighbors_per_atom
+        return self.get_neighbors(
+            cutoff_radius=cutoff_radius,
+            num_neighbors=num_neighbors,
+            id_list=id_list,
+            width_buffer=width_buffer,
+        ).numbers_of_neighbors
 
+    @deprecate(allow_ragged="use mode instead.")
```
liamhuber:
```suggestion
    @deprecate(allow_ragged="use `mode='ragged'` instead.")
```

pyiron_atomistics/atomistics/structure/atoms.py
```diff
@@ -1269,60 +1270,38 @@ def get_numbers_of_neighbors_in_sphere(
...
         num_neighbors=None,
-        t_vec=True,
         tolerance=2,
         id_list=None,
         width_buffer=1.2,
         allow_ragged=True,
+        mode='ragged',
         norm_order=2,
     ):
         """
```
liamhuber, line 1292: Nit: As far as I can see here, `get_neighbors_by_distance` has an identical docstring as `get_neighbors`. Please remove the code duplication with `wraps` or `get_neighbors_by_distance.__doc__ = get_neighbors.__doc__`.

pyiron_atomistics/atomistics/structure/atoms.py
```diff
@@ -1337,37 +1316,38 @@ def get_neighbors_by_distance(
...
-            t_vec=t_vec,
             tolerance=tolerance,
             id_list=id_list,
             width_buffer=width_buffer,
             allow_ragged=allow_ragged,
+            mode=mode,
             norm_order=norm_order,
         )
 
+    @deprecate(allow_ragged="use mode instead.")
```
liamhuber:
```suggestion
    @deprecate(allow_ragged="use `mode='ragged'` instead.")
```

pyiron_atomistics/atomistics/structure/atoms.py
```diff
@@ -1337,37 +1316,38 @@ def get_neighbors_by_distance(
...
-                        (pbc are automatically taken into account)
             tolerance (int): tolerance (round decimal points) used for computing neighbor shells
             id_list (list): list of atoms the neighbors are to be looked for
             cutoff_radius (float): Upper bound of the distance to which the search must be done
             width_buffer (float): width of the layer to be added to account for pbc.
-            allow_ragged (bool): Whether to allow ragged list of arrays or rectangular
-                numpy.ndarray filled with np.inf for values outside cutoff_radius
+            allow_ragged (bool): (Deprecated; use mode) Whether to allow ragged list of arrays or
+                rectangular numpy.ndarray filled with np.inf for values outside cutoff_radius
+            mode (str): Representation of per-atom quantities (distances etc.). Choose from
```
liamhuber: The behaviour of the different modes should also be documented. Here at the variable definition is fine, but since it's longer you could also have something like `"Neighbor data is formatted to be returned in one of three ways based on the mode argument. The first and default format is 'filled'..."`

pyiron_atomistics/atomistics/structure/neighbors.py
```diff
@@ -70,6 +68,36 @@ def __init__(self, ref_structure):
...
+        """
+        for k,v in self._mode.items():
+            if v:
+                return k
+
+    @mode.setter
+    def mode(self, new_mode):
+        if new_mode not in self._mode.keys():
+            raise KeyError(
+                new_mode + " not found. Available modes: 'filled', 'ragged', 'flattened'"
```
liamhuber:
```suggestion
                f"{new_mode} not found. Available modes: {', '.join(self._mode.keys())}"
```

improved resilience by removing hard-coded values.

pyiron_atomistics/atomistics/structure/neighbors.py
```diff
@@ -79,26 +107,63 @@ def __repr__(self):
...
         new_neigh.wrap_positions = self.wrap_positions
         new_neigh._tree = self._tree
         new_neigh.num_neighbors = self.num_neighbors
         new_neigh.cutoff_radius = self.cutoff_radius
         new_neigh._norm_order = self._norm_order
+        new_neigh._positions = self._positions.copy()
         return new_neigh
 
+    def _reshape(self, value, key=None, ref_vector=None):
+        if key is None:
```
liamhuber, line 131: This is absolutely a style thing, so ignore it if you don't like it! I've taken to one-lining this operation most of the time it pops up, i.e. `key = self.mode if key is None else key` (or the inverse `key = key if key is not None else self.mode`, personally I find them equally easy to read.) Ignore if you don't like.

pyiron_atomistics/atomistics/structure/neighbors.py
```diff
@@ -79,26 +107,63 @@ def __repr__(self):
...
+            )
+        return self._reshape(self._vecs)
+
+    @property
+    def indices(self):
+        """Indices of neighboring atoms"""
+        return self._reshape(self._indices)
+
+    @property
+    def atom_numbers(self):
```
liamhuber, line 166: I didn't get a chance to mention it in the meeting, but I find this variable name unintuitive. I feel like it should somehow also utilize the root "index" to make its relation to the `indices` attribute clear, but I struggle to think of something I really like.

In the same way that we make `volume` and `energy_X` singular in MD output, what do you think about making this singular to contrast it to the plural `indices` where there's more than one value in each row? Something like `target_index`?

pyiron_atomistics/atomistics/structure/neighbors.py
```diff
@@ -118,53 +183,40 @@ def norm_order(self, value):
...
     def allow_ragged(self):
         """
         Whether to allow ragged list of distancs/vectors/indices or fill empty slots with numpy.inf
         to get rectangular arrays
         """
-        return self._allow_ragged
+        return self._mode['ragged']
 
     @allow_ragged.setter
+    @deprecate("Use mode", version="1.0.0")
```
liamhuber:
```suggestion
    @deprecate("Use `mode='ragged'`", version="1.0.0")
```

pyiron_atomistics/atomistics/structure/neighbors.py
```diff
@@ -230,16 +277,15 @@ def _get_distances_and_indices(
...
         indices[distances<np.inf] = self._get_wrapped_indices()[indices[distances<np.inf]]
-        if allow_ragged is None:
-            allow_ragged = self.allow_ragged
-        if allow_ragged:
-            return self._contract(distances, distances), self._contract(indices, distances)
-        return distances, indices
+        indices[distances==np.inf] = np.iinfo(np.int32).max
+        return self._reshape(distances, mode, distances), self._reshape(indices, mode, distances)
 
+    @deprecate(allow_ragged="use mode instead.")
```
liamhuber:
```suggestion
    @deprecate(allow_ragged="use `mode='ragged'` instead.")
```

pyiron_atomistics/atomistics/structure/neighbors.py
```diff
@@ -253,68 +299,47 @@ def get_indices(
...
-        return self._get_distances_and_indices(
-            positions=positions,
-            allow_ragged=allow_ragged,
-            num_neighbors=num_neighbors,
-            cutoff_radius=cutoff_radius,
-            width_buffer=width_buffer,
-        )[0]
+        return np.sum(self._distances < np.inf, axis=-1)
 
+    @deprecate(allow_ragged="use mode instead.")
```
liamhuber:
```suggestion
    @deprecate(allow_ragged="use `mode='ragged'` instead.")
```

pyiron_atomistics/atomistics/structure/neighbors.py
```diff
@@ -363,28 +394,25 @@ def _get_vectors(
...
         else:
-            raise AssertionError(
-                'vectors not created yet -> put positions or reinitialize with t_vec=True'
-            )
-        if allow_ragged==self.allow_ragged:
-            return vectors
-        if allow_ragged:
-            return self._contract(vectors, distances)
-        return self._fill(vectors)
+            if self._vecs is None:
```
liamhuber: I don't understand, what is the point if assigning a variable to `_` and then never grabbing `_` later? Is this some sort of garbage collecting trick? As far as I can see these two lines can be removed and the functionality is the same, but if I'm wrong I will learn something new 😄

pyiron_atomistics/atomistics/structure/neighbors.py
```diff
@@ -673,19 +665,18 @@ def get_local_shells(self, tolerance=None, cluster_by_distances=False, cluster_b
         Returns:
             shells (numpy.ndarray): shell indices
         """
-        allow_ragged = self.allow_ragged
-        if allow_ragged:
-            self.allow_ragged = False
         if tolerance is None:
             tolerance = self._tolerance
+        if mode is None:
```
liamhuber, line 633: Again, only do this if you see it and like it more, but I've been enjoying `mode = self.mode if mode is None else mode`

pyiron_atomistics/atomistics/structure/neighbors.py
```diff
@@ -730,20 +723,19 @@ def get_global_shells(self, tolerance=None, cluster_by_distances=False, cluster_
         Returns:
             shells (numpy.ndarray): shell indices (cf. shells)
         """
-        allow_ragged = self.allow_ragged
-        if allow_ragged:
-            self.allow_ragged = False
         if tolerance is None:
             tolerance = self._tolerance
-        if self.distances is None:
+        if mode is None:
```
liamhuber, line 691: can be one-lined if that's what you're into

tests/atomistics/structure/test_neighbors.py
```diff
@@ -254,12 +254,54 @@ def test_get_local_shells_ragged(self):
...
         self.assertEqual(np.sum([len(s)==11 for s in neigh.get_local_shells(cluster_by_vecs=True)]), 12)
         self.assertEqual(np.sum([len(s)==11 for s in neigh.get_local_shells(cluster_by_distances=True, cluster_by_vecs=True)]), 12)
 
+    def test_get_shells_flattened(self):
+        structure = StructureFactory().ase.bulk('Al', cubic=True).repeat(2)
+        del structure[0]
+        r = structure.cell[0,0]*0.49
+        neigh = structure.get_neighbors(cutoff_radius=r, num_neighbors=None, mode='flattened')
+        self.assertEqual(len(np.unique(neigh.shells)), 1)
+        self.assertEqual(len(neigh.shells), 360)
```
liamhuber, line 249: I saw failing tests on the discussion page, and I bet this style of test is the main culprit now that you're using `np.iinfo(np.int32).max`

liamhuber (approved):
Very nice! I lazily procrastinated on reviewing this, and I'm glad since after the pyiron meeting and other reviews it's already good-to-go 😝 

I made a bunch of suggestions that I hope you will like, but I will not cry myself to sleep even if it's merged as-is. Nice improvements!!!

samwaseda:
Aaaaaah sorry I should have changed it to draft right away after the pyiron meeting - I'm changing a lot of things right now so maybe it's not worth spending too much time looking at the current version

pyiron_atomistics/atomistics/structure/neighbors.py
samwaseda: Yeah I've been having the same problem for years and I never came up with a good name. So far I'm glad that @sudarsan-surendralal suggested `atom_numbers`, but it's true that the relation to `indices` might still be somewhat unclear.

Since we tend to use plural for per-atom quantities, I might actually prefer to keep plural here as well. I like `target_index`, but it still might be difficult to know whether the atoms themselves are meant or the neighbours. What do you think of `self_indices`?

pyiron_atomistics/atomistics/structure/neighbors.py
samwaseda: That's because inside `def vecs(self)`, `self._vecs` is initialized (no idea if this is a particularly elegant way to do that). Anyway following last week's pyiron meeting this functionality is going to be removed.

samwaseda:
I'm going to merge it as soon as the tests pass.

PR #241, opened 2021-06-21
Create new Atoms in get_structure if size changes

pmrv:
pbc and species taken from job.structure, because they're not saved in
the HDF5.

This enables to use `get_structure` on an interactive jobs, when you put in differently sized structures.  It does not help sadly with the problem of needing all species present in the initial structure.


sudarsan-surendralal:
> 
> 
> pbc and species taken from job.structure, because they're not saved in
> the HDF5.
> 
> This enables to use `get_structure` on an interactive jobs, when you put in differently sized structures. It does not help sadly with the problem of needing all species present in the initial structure.

I guess we're talking about a grand canonical situation where the species and the count of each species changes? Shouldn't there be something like `self.output.species` for such jubs then? Wouldn't this solve the problem that all species should be present at first?

pmrv:
In this case it's more about calculating energies for arbitrary structures via an interactive job, but grand-canonical would be another application.

I had wanted to use `output.species` but in my quick test it was not populated, i.e. it was an empty list.  Maybe I missed something?

liamhuber (approved):
Not a corner of the code base I'm super familiar with, but it looks good to me.

pyiron_atomistics/atomistics/job/atomistic.py
```diff
@@ -612,19 +612,23 @@ def get_final_structure(self):
...
-                snapshot.indices = self.output.indices[frame]
+                indices = self.output.indices[frame]
             except IndexError:
-                pass
+                indices = None
+        if indices is not None and len(indices) != len(self.structure):
+            snapshot = Atoms(species=self.structure.species, indices=indices,
+                             positions=np.zeros(indices.shape + (3,)), cell=cell, pbc=self.structure.pbc)
+        else:
+            snapshot = self.structure.copy()
```
sudarsan-surendralal, line 631: Shouldn't you set the cell of the snapshot here? 
```suggestion
            snapshot = self.structure.copy()
            if cell is not None:
                snapshot.cell = cell
```

sudarsan-surendralal (changes requested):

sudarsan-surendralal:
> I had wanted to use `output.species` but in my quick test it was not populated, i.e. it was an empty list. Maybe I missed something?

I realize that the interactive codes (for now LAMMPS and SPHINX) don't store the changing species info in the output (which is bad IMO). Would it be difficult to implement this (perhaps even in a separate PR)?

pyiron_atomistics/atomistics/job/atomistic.py
pmrv: Jup, added this and also the indices.

pmrv:
> > I had wanted to use `output.species` but in my quick test it was not populated, i.e. it was an empty list. Maybe I missed something?
> 
> I realize that the interactive codes (for now LAMMPS and SPHINX) don't store the changing species info in the output (which is bad IMO). Would it be difficult to implement this (perhaps even in a separate PR)?

I've just checked that this is also the case for non-interactive run jobs, e.g. Lammps & calc_minimize.  I've used `job['output/structure/species']` for now which should be the same as `job.structure.species`, but I agree that this should be addressed in another PR.

pyiron_atomistics/atomistics/job/atomistic.py
```diff
@@ -612,19 +612,30 @@ def get_final_structure(self):
...
+            snapshot = Atoms(species=species, indices=indices,
+                             positions=np.zeros(indices.shape + (3,)), cell=cell, pbc=self.structure.pbc)
+        else:
+            snapshot = self.structure.copy()
+            if cell is not None:
+                snapshot.cell = cell
+            if indices is not None:
+                snapshot.indices = indices
+            if species is not None:
+                snapshot.species = species
```
sudarsan-surendralal: You can't set the species directly and have to use `set_species(species)` (This should fix the tests)

sudarsan-surendralal: Sorry, I realize that this doesn't work too since the species list should be a list of `ChemicalElement` instances and not symbols. IMO since this line basically does nothing since we use the original species info, it's probably best to remove this line.

sudarsan-surendralal:
Everything seems to work fine now! But still, I wonder if it's possible to add some tests where we check get structure returns the exact thing we (cell, species, and positions). For instance , we fail to capture the fact that we didn't update the cell (based on the initial commits where the tests passed). Sorry for insisting on this but this is a function that pretty much every pyiron workflow uses!

pmrv:
I've added a simple test case, but it required me to add a packed project and some tango around how `project.unpack` works right now.

tests/atomistics/job/test_atomistic.py
```diff
@@ -69,5 +69,26 @@ def test_get_displacements(self):
...
 
+    def test_get_structure(self):
+        """get_structure() should return structures with the exact values from the HDF files even if the size of
+        structures changes."""
+
+        # tested here with lammps as a concrete instantiation of AtomisticGenericJob
+        # have to do extra tango because Project.unpack is weird right now
+        cwd = os.curdir
+        to_dir = os.path.join(self.project.path, "..")
+        shutil.copy("tests/static/lammps_test_files/get_structure_test.tar.gz", to_dir)
```
sudarsan-surendralal, line 81: You haven't imported `shutil` in this file!

pmrv: Ah, the curse of copy-paste, fixed.

sudarsan-surendralal (approved):
Looks great now!

niklassiemer:
Depending on the start directory of the tests, one may get a `FileNotFoundError` due to "tests/static/lammps_test_files/get_structure_test.tar.gz".

PR #243, opened 2021-06-22
Update to 0.2.17

niklassiemer:


pmrv:
I added some lines to catch only vasp related warnings, so that this will hopefully not happen again.

niklassiemer:
Once `pyiron_base` is on conda-forge, I will also update the `environment/setup` and then it will hopefully run through 😄

niklassiemer:
> I added some lines to catch only vasp related warnings, so that this will hopefully not happen again.

Are all of the remaining warnings from VASP? We'll see on the CI later.

pmrv:
> > I added some lines to catch only vasp related warnings, so that this will hopefully not happen again.
> 
> Are all of the remaining warnings from VASP? We'll see on the CI later.

At least when I checked this locally there where three warnings from the Vasprun parser.

tests/vasp/test_vasp_import.py
```diff
@@ -46,9 +47,10 @@ def test_import(self):
             self.file_location, "../static/vasp_test_files/full_job_minor_glitch"
         )
         with warnings.catch_warnings(record=True) as w:
-            warnings.simplefilter("always")
+            warnings.simplefilter("ignore")
+            warnings.simplefilter("always", category=VasprunWarning)
             self.project.import_from_path(path=folder_path, recursive=False)
-            self.assertEqual(len(w), 4)
+            self.assertEqual(len(w), 3)
```
sudarsan-surendralal, line 53: This was bad test design on my part! Perhaps a better way to check the errors is:
```suggestion
            self.assertTrue(all([isinstance(warn, VasprunWarning) for warn in w])) 
```

niklassiemer: With the patch from Marvin, this would always be `True`. Thus, checking for exactly 3 `VasprunWarnings` sounds reasonable to me.
I will now merge this PR and release a new `pyiron_atomistics`. Feel free to adapt the test in another PR, though.

PR #249, opened 2021-06-24
Allow overriding of cluster method in get_layers()

pmrv:
I promised this when `get_layers()` was originally written by @samwaseda, so here it is: a flag to change the cluster algorithm, merely the next step in my global plan to DBSCAN everything.  (Agglomerative clustering is just a bit to slow for my current application)

Since `distance_threshold` is specific to `AgglomerativeClustering` I would suggest we deprecate it, but for now I've left it in.

### DIFF
--- a/pyiron_atomistics/atomistics/structure/analyse.py
+++ b/pyiron_atomistics/atomistics/structure/analyse.py
@@ -427,14 +427,14 @@ class Analyse:
         if planes is not None:
             mat = np.asarray(planes).reshape(-1, 3)
             positions = np.einsum('ij,i,nj->ni', mat, 1/np.linalg.norm(mat, axis=-1), positions)
+        if cluster_method is None:
+            cluster_method = AgglomerativeClustering(
+                linkage='complete',
+                n_clusters=None,
+                distance_threshold=distance_threshold
+            )
         layers = []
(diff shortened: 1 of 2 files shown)


pyiron_atomistics/atomistics/structure/analyse.py
```diff
@@ -422,11 +429,13 @@ def get_layers(self, distance_threshold=0.01, id_list=None, wrap_atoms=True, pla
             positions = np.einsum('ij,i,nj->ni', mat, 1/np.linalg.norm(mat, axis=-1), positions)
         layers = []
         for ii, x in enumerate(positions.T):
-            cluster = AgglomerativeClustering(
-                linkage='complete',
-                n_clusters=None,
-                distance_threshold=distance_threshold
-            ).fit(x.reshape(-1, 1))
+            if cluster_method is None:
```
samwaseda: Does `cluster_method` have to be defined within the `for`-loop?

pmrv: No, should be fine outside, too.  But in any case creating a new `AgglomerativeClustering`, etc. should be cheap. I'll just move it though, for symmetry.

samwaseda (approved):
Okay it looks perfect to me :)

PR #251, opened 2021-06-27
Neighbour analysis for the entire trajectory

sudarsan-surendralal:
The idea is to easily perform a neighbor analysis over the trajectory of an atomistic simulation. This is something I've been doing in my notebooks for some time which could be useful to others as well. There are also ways to make this more efficient. So please feel free to suggest improvements!


pyiron_atomistics/atomistics/job/atomistic.py
```diff
@@ -889,3 +896,25 @@ def __dir__(self):
             return hdf5_path.list_nodes()
         else:
             return []
+
+def get_neighbors_traj(struct, positions, cells=None, num_neighbors=20):
+    [n_steps, n_atoms, _] = positions.shape
+    indices, distances = [np.zeros((n_steps, n_atoms, num_neighbors)) for _ in range(2)]
+    vecs = np.zeros((n_steps, n_atoms, num_neighbors, 3))
+    struct = struct.copy()
+    if cells is not None:
```
samwaseda: Wouldn’t it be better to do the `if`-statement inside the `for`-loop in order to reduce lines?

samwaseda:
I see that the results are stored in private variables. How are they supposed to be used?

pyiron_atomistics/atomistics/job/atomistic.py
sudarsan-surendralal: > 
> 
> Wouldn’t it be better to do the `if`-statement inside the `for`-loop in order to reduce lines?

Yes, but won't it make the code slower since we make the same check at every time step?

samwaseda: It’s about whether to perform `if cells is not None` or not, correct? I don’t think that this operation could ever take noticeable amount of time.

sudarsan-surendralal:
> 
> 
> I see that the results are stored in private variables. How are they supposed to be used?

I've updated the structure of the class! This can now be accessed by
 ```
neighbors = job.get_neighbors()
neighbors.neighbor_vectors
```

**Comment by samwaseda** [comment]
Currently it's mostly equivalent to `[job.get_structure(i).get_neighbors() for i in range(given_range)]`, but I presume that this class is going to be extended to do more physics in the future, right? Can you tell (roughly, no detail needed) how you use it (i.e. what kind of physical quantities do you usually extract in your work)? Depending on the purpose I might have different feedback.

**Comment by sudarsan-surendralal** [comment]
> 
> 
> Currently it's mostly equivalent to `[job.get_structure(i).get_neighbors() for i in range(given_range)]`, 

In principle, yes. Such code is trivial but I have to repeat this in pretty much all my notebooks. Options to do this only for certain parts of the trajectory is also pretty useful for me


> but I presume that this class is going to be extended to do more physics in the future, right? Can you tell (roughly, no detail needed) how you use it (i.e. what kind of physical quantities do you usually extract in your work)? Depending on the purpose I might have different feedback.

I'll mostly use this to analyze bond breaking/formation for molecules (reactivity), species sensitive radial distribution functions and so on!

File: pyiron_atomistics/atomistics/job/atomistic.py [hunk 2]
```diff
@@ -598,6 +598,43 @@ def write_traj(
             **kwargs
         )
 
+    def get_neighbors_snapshots(self, snapshot_indices=None, num_neighbors=12, **kwargs):
```
pmrv, line 601: Would it make sense to move this to the `Trajectory` class somehow?

pmrv: Ah, it's already on there anyway.  Then I'd want to think about whether this is really necessary to have it duplicated here.  It's just about typing

```python
job.trajectory().get_neighbors()
```
vs.
```python
job.get_neighbors()
```
?

pyiron_atomistics/atomistics/structure/neighbors.py
```diff
@@ -1047,3 +1047,147 @@ def get_cluster(dist_vec, ind_vec, prec=prec):
...
+class NeighborsTraj:
+
+    """
+    This class generates the neighbors for a given atomistic trajectory. The assumption here is that the trajectory is
+    canonical (no change in the number and types of species throughout the trajectory). The resulting indices,
+    distances, and vectors are stored as numpy arrays.
+
+    """
+
+    def __init__(self, init_structure=None, positions=None, cells=None, num_neighbors=12, **kwargs):
```
pmrv: I think it'd be extremely useful for this to take any object that implements `HasStructure` instead of passing all of these things separately.  Calling `get_structure` every step might be a bit slower, but since you're doing essentially what `AtomisticGenericJob.get_structure` does anyway, it shouldn't be that much.

pmrv: If it turns out much slower, then I'd say we add a method roughly like this
```python
@classmethod
def from_structures(cls, has_structure):
  structs = list(has_structure.iter_structures())
  return cls(structs[0], [s.positions for s in structs], [s.cell for s in structs])
```
to `NeighborTraj` to achieve the same generality, but offer a small escape hatch for cases where we know `get_structure` is too slow.

pyiron_atomistics/atomistics/job/atomistic.py
sudarsan-surendralal: > 
> 
> Ah, it's already on there anyway. Then I'd want to think about whether this is really necessary to have it duplicated here. It's just about typing
> 
> ```python
> job.trajectory().get_neighbors()
> ```
> 
> vs.
> 
> ```python
> job.get_neighbors()
> ```
> 
> ?

Yes, I prefer the second one over the first mainly because it's available via auto-complete and the first isn't. Plus it is naturally equivalent to `structure.get_neighbors()`

pyiron_atomistics/atomistics/structure/neighbors.py
sudarsan-surendralal: > If it turns out much slower, then I'd say we add a method roughly like this
> 
> ```python
> @classmethod
> def from_structures(cls, has_structure):
>   structs = list(has_structure.iter_structures())
>   return cls(structs[0], [s.positions for s in structs], [s.cell for s in structs])
> ```
> 
> to `NeighborTraj` to achieve the same generality, but offer a small escape hatch for cases where we know `get_structure` is too slow.

I think this is a great way to generalize this for any derivative of `HasStructure`. However, I don't think building the positions and cells from the individual structures is not that efficient.

So how about something like this:

```
def __init__(self, has_structure=None, init_struct=None, positions=None, cells=None):
    if has_structure is None:
        if positions is None or init_struct is None:
            raise ValueError()

```

and then

```
def _get_neighbors_hs(has_struct, num_neighbors=20, **kwargs):
    [struct.get_neighbors() for struct in has_struct.iter_structures()]
```
which gets called instead of `_get_neighbors`

pyiron_atomistics/atomistics/job/atomistic.py
pmrv: auto-complete is a fair argument.

sudarsan-surendralal:
@samwaseda & @pmrv any further changes you'd like in this PR?

pyiron_atomistics/atomistics/structure/neighbors.py
pmrv: I think we should simply give it a try to use `get_structure()` everywhere and see if it really is slower.  Checking `Trajectory._get_structure` and your code in `_get_neighbors()` looks essentially the same to me, so I don't expect a big hit. 

If you don't want to do it, I can give it a go tonight.

pmrv: Also `__init__` needs to accept an `init` argument as the first argument and pass it to `super().__init__()`, otherwise recursive loading from HDF will not work.

pyiron_atomistics/atomistics/structure/neighbors.py
```diff
@@ -1047,3 +1047,120 @@ def get_cluster(dist_vec, ind_vec, prec=prec):
...
+        """
+        Compute the neighbors across the trajectory
+        """
+        self._neighbor_indices, self._neighbor_distances, self._neighbor_vectors = \
+            _get_neighbors(self._init_structure,
+                           positions=self._positions,
+                           cells=self._cells,
+                           num_neighbors=self._num_neighbors, **self._get_neighbors_kwargs)
+
+    def to_hdf(self, hdf, group_name="neighbors_traj"):
```
pmrv: If you want to set a default "group_name" to be used when writing to HDF, pass `table_name` to `super().__init__`, then you don't need to overload this.

pyiron_atomistics/atomistics/structure/neighbors.py
```diff
@@ -1047,3 +1047,120 @@ def get_cluster(dist_vec, ind_vec, prec=prec):
...
+        """
+        The maximum number of neighbors to be computed
+
+        Returns:
+
+            int: The max number of neighbors
+        """
+        return self._num_neighbors
+
+    def compute_neighbors(self):
```
pmrv, line 1088: Since the only purpose of this class is to call this method, can we just call it in `__init__` automatically (when it is not loaded from HDF)?

pyiron_atomistics/atomistics/structure/neighbors.py
```diff
@@ -1047,3 +1047,120 @@ def get_cluster(dist_vec, ind_vec, prec=prec):
...
+
+        Args:
+            hdf (pyiron_base.generic.hdfio.FileHDFio): HDF5 group or file
+            group_name (str): Name of the HDF5 group
+
+        """
+        super().to_hdf(hdf=hdf, group_name=group_name)
+
+
+def _get_neighbors(struct, positions, cells=None, num_neighbors=20, **kwargs):
```
pmrv:
```suggestion
def _get_neighbors(struct, positions, cells=None, num_neighbors=12, **kwargs):
```
Consistent defaults make it a bit easier to read.  I have no opinions on 12 vs. 20, but we should pick one.

pyiron_atomistics/atomistics/structure/neighbors.py
sudarsan-surendralal: > 
> 
> I think we should simply give it a try to use `get_structure()` everywhere and see if it really is slower. Checking `Trajectory._get_structure` and your code in `_get_neighbors()` looks essentially the same to me, so I don't expect a big hit.
> 
> If you don't want to do it, I can give it a go tonight.

Sure I'll merge your PR into this one and see what happens!

sudarsan-surendralal: > 
> 
> Also `__init__` needs to accept an `init` argument as the first argument and pass it to `super().__init__()`, otherwise recursive loading from HDF will not work.

Not sure what you mean by this. Could you clarify?

pyiron_atomistics/atomistics/structure/neighbors.py
sudarsan-surendralal: > 
> 
> If you want to set a default "group_name" to be used when writing to HDF, pass `table_name` to `super().__init__`, then you don't need to overload this.

Cool, thanks! I was wondering how to do this.

pyiron_atomistics/atomistics/structure/neighbors.py
pmrv: > > Also `__init__` needs to accept an `init` argument as the first argument and pass it to `super().__init__()`, otherwise recursive loading from HDF will not work.
> 
> Not sure what you mean by this. Could you clarify?

When you derive from `DataContainer` the new `__init__` method must be compatible to the original one. One of the ways `DataContainer` is instantiated is by passing a collection to initialize it, like so
```python
d = DataContainer({'a': 1, 'b': 3, 2: 42})
assert d[1] == 3
```
Because this is used by the `DataContainer` internally when loading from hdf, subclasses also need to offer this.  There's a little bit more explanation [here in the attention box](https://pyiron-base.readthedocs.io/en/latest/apidoc/pyiron_base.generic.datacontainer.html?highlight=DataContainer).

pyiron_atomistics/atomistics/structure/neighbors.py
```diff
@@ -1047,3 +1047,147 @@ def get_cluster(dist_vec, ind_vec, prec=prec):
...
+                             (eg. cutoff_radius, norm_order , etc.)
+        """
+        self._init_structure = init_structure
+        self._neighbor_indices = None
+        self._neighbor_distances = None
+        self._neighbor_vectors = None
+        self._positions = positions
+        self._cells = cells
+        self._num_neighbors = num_neighbors
+        self._get_neighbors_kwargs = kwargs
```
pmrv: For `__init__` to be compatible a change like this would be necessary.  When `init` is given the other parameters are not set, because presumably they are set in the `init` read from HDF5.  If you want to be complete, you can check that `init` really contains all the attributes that `__init__` normally sets and set those manually that are not in `init.

```suggestion
    def __init__(self, init=None, init_structure=None, positions=None, cells=None, num_neighbors=12, **kwargs):
        """

        Args:
            init_structure (pyiron_atomistics.atomistics.structure.atoms.Atoms): Any given structure of the trajectory
            positions (numpy.ndarray): The cartesian positions of the trajectories
            cells (numpy.ndarray/None): The varying cell shapes
            num_neighbors (int): The cutoff for the number of neighbors
            **kwargs (dict): Additional arguments to be passed to the `get_neighbors()` routine
                             (eg. cutoff_radius, norm_order , etc.)
        """
        super().__init__(init=init)
        if init is None:
            self._init_structure = init_structure
            self._neighbor_indices = None
            self._neighbor_distances = None
            self._neighbor_vectors = None
            self._positions = positions
            self._cells = cells
            self._num_neighbors = num_neighbors
            self._get_neighbors_kwargs = kwargs
```

pmrv (approved):

sudarsan-surendralal:
Since a `HasStructure` instance is now an attribute of the neighbor class, calling `to_hdf` now tries to write the entire trajectory (or structure container) again in the hdf5 file. I think this is a bit redundant!

pyiron_atomistics/atomistics/structure/neighbors.py
```diff
@@ -1081,3 +1083,101 @@ def get_volume_of_n_sphere_in_p_norm(n=3, p=2):
...
+
+class NeighborsTraj(DataContainer):
+
+    """
+    This class generates the neighbors for a given atomistic trajectory. The assumption here is that the trajectory is
+    canonical (no change in the number and types of species throughout the trajectory). The resulting indices,
+    distances, and vectors are stored as numpy arrays.
+
+    """
+
```
pmrv, line 1016:
```suggestion
def __new__(cls, *args, **kwargs):
    instance = super().__new__(cls, *args, **kwargs)
    object.__setattr__(instance, "_has_structure", None)
    return instance
```
This will make sure the `HasStructure` instances is not saved in the container and will therefore not be written to HDF twice.  On the other hand this creates the problem, that when loading a `NeighborsTraj` it can't recreate it. 

One solution for atomistic jobs and the structure container would be to save the job id in the container instead of the whole instance, but that wouldn't work for `Trajectories` since they don't have an ID.

Another thing I can think of is to have `_get_neighbors` become a method of `NeighborsTraj` and have it populate positions, indices and cells attributes as it works on the structures from `get_stucture`, save them and then use these to re-create a `Trajectory` on HDF loading.

Since we're not using `to_hdf/from_hdf` yet, I'm not sure if it's worth the extra work.  In this case there's yet another way to consider: don't derive from `DataContainer` and make the caller of `NeighborsTraj` responsible for saving the output.  E.g. `AtomisticJob.get_neighbors` would create a `NeighborsTraj`, save the output to its own `self['output/neighbors/...'] then return the results.

pyiron_atomistics/atomistics/structure/neighbors.py
```diff
@@ -1081,3 +1083,101 @@ def get_volume_of_n_sphere_in_p_norm(n=3, p=2):
     """
     return (2*gamma(1+1/p))**n/gamma(1+n/p)
 
+
+class NeighborsTraj(DataContainer):
```
pmrv:
```suggestion
class NeighborsTrajectory(DataContainer):
```
I'd actually prefer to spell it out, since users don't have to type it.

pmrv:
@samwaseda & @sudarsan-surendralal I just saw that the attribute names between `Neighbors` and `NeighborsTraj` are not consistent.  I don't have a preference either way, but I think they should be the same, otherwise it'll be confusing to use.

pyiron_atomistics/atomistics/structure/neighbors.py
```diff
@@ -1081,3 +1083,101 @@ def get_volume_of_n_sphere_in_p_norm(n=3, p=2):
...
+        """
+        Compute the neighbors across the trajectory
+        """
+        self._neighbor_indices, self._neighbor_distances, self._neighbor_vectors = \
+            _get_neighbors(self._has_structure,
+                           num_neighbors=self._num_neighbors, **self._get_neighbors_kwargs)
+
+
+def _get_neighbors(has_structure, num_neighbors=20, **kwargs):
+    n_steps = len(has_structure)
```
pmrv:
```suggestion
    n_steps = has_structure.number_of_structures
```
like we discussed in the meeting.

PR #258, opened 2021-06-28
Raise error on non-zero pressure

pmrv:
If I read the vasp reference correctly more has to be done to set an external pressure, but the current code silently reverts back to fixed volume for `pressure != 0`, so I raise an error for now.


sudarsan-surendralal (approved):
Nice catch. LGTM!

PR #259, opened 2021-06-28
More structures

liamhuber:
Adding Laves phase creation routines onto the structure factory.

So far I did C15, although there might be a more elegant way to create the sublattices.

C14 and C36 should look similar, including some tests to make sure they're working.

### DIFF
--- a/pyiron_atomistics/atomistics/structure/factories/compound.py
+++ b/pyiron_atomistics/atomistics/structure/factories/compound.py
@@ -49,8 +49,36 @@ class CompoundFactory:
         return _ase.crystal((element_a, element_b), [(0, 0, 0), (1/2, 1/2, 1/2)], spacegroup=221, cell=(a, a, a))
 
     @staticmethod
-    def C14():
-        raise NotImplementedError
+    def C14(element_a, element_b, a=None, a_over_c=1, x1=0.1, z1=0):
+        """
+        Builds a hexagonal $A B_2$ C14 Laves phase cell.
+
+        .. attention:: Change in Stochiometry possible!
+
+            If any of the fractional coordinates fall onto their high symmetry values the atoms may be placed on another
+            Wyckoff position, with less sites and therefor the cell composition may change.
+
+        Args:
+            element_a (str, ase.Atom): specificies A
+            element_b (str, ase.Atom): specificies B
+            a (float): length of a & b cell vectors
+            c_over_a (float): c/a ratio
+            x1 (float): fractional x coordinate of B atoms on Wyckoff 4e
+            z1 (float): fractional z coordinate of A atoms on Wyckoff 4e
+        """
+        a = 2 * _bcc_lattice_constant_from_nn_distance(element_a) if a is None else a
+        c = a / a_over_c
+
+        # https://www.atomic-scale-physics.de/lattice/struk/c14.html
+        return _ase.crystal(( element_a, element_a, element_b ),
+                            [ # wyckoff 2a -- Zn-I
+                              (  0,   0,   0),
+                              # wyckoff 6h -- Zn-II
+                              ( 1*x1,  2*x1, 1/4),
+                              # wyckoff 4f -- Mg
+                              ( 1/3, 2/3,       z1 ),
+                            ],
+                            spacegroup=194, cell=(a, a, c, 90, 90, 120))
 
     @staticmethod
     def C15(element_a, element_b, a=None):
@@ -77,8 +105,44 @@ class CompoundFactory:
         return _ase.crystal((element_a, element_b), [(0, 0, 0), (1/8, 5/8, 1/8)], spacegroup=227, cell=(a, a, a))
 
     @staticmethod
-    def C36():
-        raise NotImplementedError
+    def C36(element_a, element_b, a=None, c_over_a=1, x1=0.1, z1=0, z2=0, z3=0):
+        """
+        Create hexagonal $A B_2$ C36 Laves phase.
+
+        .. attention:: Change in Stochiometry possible!
+
+            If any of the fractional coordinates fall onto their high symmetry values the atoms may be placed on another
+            Wyckoff position, with less sites and therefor the cell composition may change.
+
+
+        Args:
+            element_a (str, ase.Atom): specificies A
+            element_b (str, ase.Atom): specificies B
+            a (float): length of a & b cell vectors
+            c_over_a (float): c/a ratio
+            x1 (float): fractional x coordinate of B atoms on Wyckoff 4e
+            z1 (float): fractional z coordinate of A atoms on Wyckoff 4e
+            z2 (float): fractional z coordinate of A atoms on Wyckoff 4f
+            z3 (float): fractional z coordinate of B atoms on Wyckoff 4f
+        """
+        a = 2 * _bcc_lattice_constant_from_nn_distance(element_a) if a is None else a
+        c = c_over_a * a
+
+        # See: https://www.atomic-scale-physics.de/lattice/struk/c36.html
+        return _ase.crystal( (element_a, element_a, element_b, element_b, element_b),
+                [ # Wyckoff 4e -- A-I
+                  (0, 0, z1),
+                  # 4f -- A-II
+                  (1/3, 2/3, z2),
+                  # 4f -- B-I
+                  (1/3, 2/3, z3),
+                  # 6g -- B-II
+                  (1/2, 0, 0),
+                  # 6h -- B-III
+                  (x1, 2*x1, 1/4)
+                 ], spacegroup=194, cell=(a, a, c, 90, 90, 120))
+
+
 
     @staticmethod
     def D03(element_a, element_b, a=None):
--- a/pyiron_atomistics/atomistics/structure/factories/compound.py
+++ b/pyiron_atomistics/atomistics/structure/factories/compound.py
@@ -49,9 +49,13 @@ class CompoundFactory:
         return _ase.crystal((element_a, element_b), [(0, 0, 0), (1/2, 1/2, 1/2)], spacegroup=221, cell=(a, a, a))
 
     @staticmethod
-    def C14(element_a, element_b, a=None, c_over_a=1.626, x1=0.5, z1=0.5):
+    def C14(element_a, element_b, a=None, c_over_a=1.626, x1=0.1697, z1=0.5629):
         """
         Builds a hexagonal $A B_2$ C14 Laves phase cell.
+
+        Default fractional coordinates are chosen to reproduce CaMg2 Laves phase from the Springer Materials Database
+
+        https://materials.springer.com/isp/crystallographic/docs/sd_1822295
 
         .. attention:: Change in Stochiometry possible!
 
--- a/pyiron_atomistics/atomistics/structure/factories/compound.py
+++ b/pyiron_atomistics/atomistics/structure/factories/compound.py
@@ -21,6 +21,32 @@ class CompoundFactory:
     """A collection of routines for constructing Laves phase structures."""
 
     @staticmethod
+    def _bcc_lattice_constant_from_nn_distance(element):
+        """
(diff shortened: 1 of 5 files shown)


pyiron_atomistics/atomistics/structure/factories/laves.py
```diff
@@ -0,0 +1,71 @@
...
+            element_b (str): The chemical symbol for the B element.
+            a (float): The lattice constant. (Default is None, which uses the default nearest-neighbour distance for
+                the A-type element.)
+
+        Returns:
+            (Atoms): The C15 unit cell.
+        """
+        bulk = AseFactory().bulk
+        if a is None:
+            a = bulk(name=element_a).get_neighbors(num_neighbors=1).distances[0, 0] * (4 / np.sqrt(3))
```
pmrv:
```suggestion
            a = bulk(name=element_a, cubic=True).cell[0, 0]
```
We could also look them up from ASE.ChemicalElements or mendeleev, but I'd have to look that up.

pyiron_atomistics/atomistics/structure/factory.py
```diff
@@ -62,6 +64,10 @@ def ase(self):
     def aimsgb(self):
         return self._aimsgb
 
+    @property
+    def laves(self):
```
pmrv: I'm not sure if the nesting only for the three laves phases makes sense, if we don't want them poluting the `structure` namespace maybe `compound` or `binary` is more general.

pmrv (changes requested):
I think `laves` is a bit too specific for it's own namespace otherwise looks good.

pyiron_atomistics/atomistics/structure/factories/laves.py
liamhuber: > We could also look them up from ASE.ChemicalElements or mendeleev, but I'd have to look that up.

I think that's what the ASE creator is doing anyhow. IMO it's not enough of a performance hit for the more complicated.

I can't merge the commit though as it breaks in a few ways -- and makes a test fail, which is a good thing here! The base cell of C15 is diamond cubic, so what I'm actually doing is making sure the less-common species has the same nearest neighbour distance as it would in the bulk. I just checked, and using Fe it's off by a factor of two. For HCP the test breaks entirely because of `cubic=True` and is off by not quite a factor of two (I guess because of the c/a ratio).

liamhuber: I'm open to just `2*bulk(name=element_a).cell[0, 0]` if that's a good enough approximation. I haven't played around relaxing these phases and have no intuition for how much they contract or expand relative to the pure bulk phases.

pyiron_atomistics/atomistics/structure/factory.py
liamhuber: Yeah, I was hoping for some critical feedback here! I had been thinking `extra.laves` in the long run, but I also don't mind `compound`. I would like to avoid `binary` because I'm hoping that the C36 method will allow mixing species on one of the sublattices (for the benefit of the SFB).

liamhuber: After sleeping on it, I really like `compound` and made a commit for it with both of us as authors. Now I can move the B2 and D03 stuff here from the FeAl project as well, which is nice! 

I'm going to leave the discussion here unresolved though, in case there is an even better term than `compound` floating around that neither of us have thought of.

pmrv: Jup, compound works I think.

pmrv:
So I brushed up on my crystallography again and found the following snippet that generates C15 directly via the ASE spacegroup package
```python
a = ...
ase_to_pyiron(ase.spacegroup.crystal(('Mg', 'Cu'), [(0,0,0), (1/8,5/8,1/8)], spacegroup=227, cellpar=(a,a,a,90,90,90)))
```
227 is the space group number of C15 and the positions are fractional coordinates that I picked from wyckoff positions with compatible multiplicity.

Even though it's a bit cryptic to find out, I think this is nicer than constructing each of the structures by hand (and potentially making mistakes).  If you don't oppose I'd put this solution and similar ones for the other ones in.

liamhuber:
Draft is pending on Sudarsan's `PyironTestCase` getting into the `pyiron_atomistics` dependencies -- I think this will resolve the failed tests.

pmrv:
I've added methods for C14 & C36.  I don't know which are good defaults for c/a and the respective fractional coordinates, so I've left these for now.  Can you have a look @alitehranch ?

liamhuber:
@pmrv these stoichiometry changes are disconcerting and I wasn't aware of them. Is it possible to avoid this by adding a `+ epsilon` to the positions?

pmrv:
I've documented values that would/should cause a symmetry increase and adopted the defaults.  Adding epsilons could work, but you probably wouldn't end up with a structure you'd want.  E.g. if `z1 = 0` results in three equivalent positions
```
-z1
0.5 + z1
0.5 - z1
```
then you'd get positions with distances of `2*epsilon` that would blow the structure up in a minimizer.

pmrv:
I think I made a mistake previously for C15, which I have corrected now.  Again, since I'm just reading this from tables, I'd be nice if someone with Laves-experience can make a quick sanity check.

pyiron_atomistics/atomistics/structure/factories/compound.py
```diff
@@ -0,0 +1,180 @@
...
+            a (float): length of a & b cell vectors
+            c_over_a (float): c/a ratio
+            x1 (float): fractional x coordinate of B atoms on Wyckoff 6h
+            z1 (float): fractional z coordinate of A atoms on Wyckoff 4f
+        """
+        a = 2 * _bcc_lattice_constant_from_nn_distance(element_a) if a is None else a
+        c = a / a_over_c
+
+        # https://www.atomic-scale-physics.de/lattice/struk/c14.html
+        return _ase.crystal(( element_a, element_a, element_b ),
```
pmrv: Do a composition check?

pmrv:
I rebased this to use the `PyironTestCase` from base and am working on the tests.

pmrv:
I've run geometry optimizations on the current defaults and compared to @zendegani's thesis.  It seems to match, so it's good from my side.

pmrv:
I've double-checked with @zendegani 's thesis, that for the default values given now, the minimized structures come out right for C14 & C36. I'm looking into C15 now.

EDIT: C15 seems good.

pmrv:
Meh, actually the C15 structure comes out as A_2B instead of AB_2.  If I read the crystallographic info [here](https://www.atomic-scale-physics.de/lattice/struk/c15.html) correctly, my code should come out right.  I'm sure the bug is in front of the screen, but I can't figure it out right now.

niklassiemer:
> Meh, actually the C15 structure comes out as A_2B instead of AB_2. If I read the crystallographic info [here](https://www.atomic-scale-physics.de/lattice/struk/c15.html) correctly, my code should come out right. I'm sure the bug is in front of the screen, but I can't figure it out right now.

I just stumbled over this and followed your link - as far as I can see, they also want A_2B as the structure (Cu2Mg) where A (Cu) is on the 16d and B (Mg) is on the 8a side. Thus, you just need to either switch A and B in the implementation or rewrite the docstring of what you will get.

pmrv:
> > Meh, actually the C15 structure comes out as A_2B instead of AB_2. If I read the crystallographic info [here](https://www.atomic-scale-physics.de/lattice/struk/c15.html) correctly, my code should come out right. I'm sure the bug is in front of the screen, but I can't figure it out right now.
> 
> I just stumbled over this and followed your link - as far as I can see, they also want A_2B as the structure (Cu2Mg) where A (Cu) is on the 16d and B (Mg) is on the 8a side. Thus, you just need to either switch A and B in the implementation or rewrite the docstring of what you will get.

I guess I was really tired back then, but still I put `element_a` on the 8a site and `element_b` on the 16d site, so I should get A8B16, no?

pyiron_atomistics/atomistics/structure/factories/compound.py
```diff
@@ -0,0 +1,189 @@
...
+                              ( 1/3,   2/3,  z1 ),
+                              # wyckoff 2a -- B-I
+                              (   0,     0,   0 ),
+                              # wyckoff 6h -- B-II
+                              ( 1*x1, 2*x1, 1/4 ),
+                            ],
+                            spacegroup=194, cell=(a, a, c, 90, 90, 120))
+        if len(s) != 2 + 6 + 4:
+            raise ValueError("Given internal coordinates reduced symmetry, check the docstring for degenerate values!")
+        return s
```
LVC14C15: Hello, 
I didn't obtained a correct c14 structure with these atoms coordinate: to many Mg atoms were in the triple layer and the Ca atoms were not good located as well. By using these values (1/3, 2/3, 0.5629),(0,0,0), (0.1697,0.3394,1/4), it works well :)
I found the correct value here: https://materials.springer.com/isp/crystallographic/docs/sd_1822295

niklassiemer: Do you get your intended structure if you specify `x1=0.1697` and `z1=0.5629`? If yes, we may argue about the defaults, here, but the general function is working.

LVC14C15: I use directly these lines of the code:
`c14=pr.create.structure.ase.crystal(('Ca','Mg','Mg'), 
                             ((1/3, 2/3, 0.5629),(0,0,0), (0.1697,0.3394,1/4)),
                            spacegroup=194, cell=(6.146, 6.146, 9.979, 90, 90, 120)).repeat([3,3,3])`
So I guess the general functions is working, but x1 and z1 have to change to reproduce the c14 correctly :)

pmrv: I updated the defaults, also the ones for C36.

LVC14C15: The C36 structure seems good now for me with the new values, as well as the C15.

pmrv:
> > > Meh, actually the C15 structure comes out as A_2B instead of AB_2. If I read the crystallographic info [here](https://www.atomic-scale-physics.de/lattice/struk/c15.html) correctly, my code should come out right. I'm sure the bug is in front of the screen, but I can't figure it out right now.
> > 
> > 
> > I just stumbled over this and followed your link - as far as I can see, they also want A_2B as the structure (Cu2Mg) where A (Cu) is on the 16d and B (Mg) is on the 8a side. Thus, you just need to either switch A and B in the implementation or rewrite the docstring of what you will get.
> 
> I guess I was really tired back then, but still I put `element_a` on the 8a site and `element_b` on the 16d site, so I should get A8B16, no?

I've simply swapped A & B now, so it should work.

pmrv:
>  AssertionError: 1 not less than 0.0 : Primary A site should be significantly non-symmetric.

@liamhuber What does this test mean? :S

liamhuber:
> > AssertionError: 1 not less than 0.0 : Primary A site should be significantly non-symmetric.
> 
> @liamhuber What does this test mean? :S

@pmrv, sorry I totally missed this comment! The message is trying to say that in $AB_2$ C15 Laves phase, the A-site is not centrosymmetric. I (hopefully?) improved the clarity of the error message. More substantially, it was failing because I never made sure that the order of my symmetry-unique IDs was [A, B], so I was testing the wrong site! I fixed that too.

pmrv:
> > > AssertionError: 1 not less than 0.0 : Primary A site should be significantly non-symmetric.
> > 
> > 
> > @liamhuber What does this test mean? :S
> 
> @pmrv, sorry I totally missed this comment! The message is trying to say that in $AB_2$ C15 Laves phase, the A-site is not centrosymmetric. I (hopefully?) improved the clarity of the error message. More substantially, it was failing because I never made sure that the order of my symmetry-unique IDs was [A, B], so I was testing the wrong site! I fixed that too.

Ah, cool. I merge it tomorrow then. Ping @LVC14C15

PR #260, opened 2021-06-29
Add Wrapper for Atomsk Structure Creation

pmrv:
It turned out to be quite simple, so I've added it here instead of a new repo.  I'll brush up the docstrings a bit and add examples and tests I got from @zhuochengXIE, but the main funtionality is there.

### DIFF
--- a/pyiron_atomistics/atomistics/structure/factories/atomsk.py
+++ b/pyiron_atomistics/atomistics/structure/factories/atomsk.py
@@ -2,7 +2,6 @@
 # Copyright (c) Max-Planck-Institut für Eisenforschung GmbH - Computational Materials Design (CM) Department
 # Distributed under the terms of "New BSD License", see the LICENSE file.
 
-import io
 import subprocess
 import tempfile
 import os.path
--- a/pyiron_atomistics/atomistics/structure/factories/atomsk.py
+++ b/pyiron_atomistics/atomistics/structure/factories/atomsk.py
@@ -11,7 +11,11 @@ from pyiron_atomistics.atomistics.structure.atoms import ase_to_pyiron
 
 from ase.io import read, write
 
+class AtomskError(Exception):
+    pass
+
 class AtomskBuilder:
+    """Class to build CLI arguments to Atomsk."""
 
     def __init__(self):
         self._options = []
@@ -87,9 +91,15 @@ class AtomskBuilder:
                 write(os.path.join(temp_dir, "input.exyz"), self._structure, format="extxyz")
             proc = subprocess.run(["atomsk", *" ".join(self._options).split()],
                                   capture_output=True, cwd=temp_dir)
-            return ase_to_pyiron(read(io.StringIO(proc.stdout.decode("utf8")), format="extxyz"))
+            output = proc.stdout.decode("utf8")
+            for l in output.split("\n"):
+                if l.strip().startswith("X!X ERROR:"):
+                    raise AtomskError(f"atomsk returned error: {output}")
+            return ase_to_pyiron(read(io.StringIO(output), format="extxyz"))
 
     def __getattr__(self, name):
+        # magic method to map method calls of the form self.foo_bar to options like -foo-bar; arguments converted str
+        # and appended after option, keyword arguments are mapped to strings like 'key value'
         def meth(*args, **kwargs):
             args_str = " ".join(map(str, args))
             kwargs_str = " ".join(f"{k} {v}" for k, v in kwargs.items())
--- a/pyiron_atomistics/atomistics/structure/factories/atomsk.py
+++ b/pyiron_atomistics/atomistics/structure/factories/atomsk.py
@@ -11,6 +11,17 @@ from pyiron_atomistics.atomistics.structure.atoms import ase_to_pyiron
 
 from ase.io import read, write
 
+__author__ = "Marvin Poul"
+__copyright__ = (
+    "Copyright 2021, Max-Planck-Institut für Eisenforschung GmbH - "
+    "Computational Materials Design (CM) Department"
+)
+__version__ = "1.0"
+__maintainer__ = "Marvin Poul"
+__email__ = "poul@mpie.de"
+__status__ = "production"
+__date__ = "Jun 30, 2021"
+
 class AtomskError(Exception):
     pass
 
--- a/pyiron_atomistics/atomistics/structure/factories/atomsk.py
+++ b/pyiron_atomistics/atomistics/structure/factories/atomsk.py
@@ -5,6 +5,7 @@
 import subprocess
 import tempfile
 import os.path
+import shutil
 
 from pyiron_atomistics.atomistics.structure.atoms import ase_to_pyiron
 
@@ -21,6 +22,8 @@ __maintainer__ = "Marvin Poul"
 __email__ = "poul@mpie.de"
 __status__ = "production"
 __date__ = "Jun 30, 2021"
+
+_ATOMSK_EXISTS = shutil.which("atomsk") != None
 
 class AtomskError(Exception):
     pass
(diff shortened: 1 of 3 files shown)


samwaseda:
Nice! Is there a way to include dislocations? doesn't have to be this PR.

pmrv:
> Nice! Is there a way to include dislocations? doesn't have to be this PR.

Yes, it's not properly documented yet, but all atomsk flags are supported.  For dislocations it would be something like
```python
pr.create.structure.atomsk.create('fcc', 3.6, 'Cu').duplicate(5, 1, 5).dislocation('0.498*box',  '0.500*box', 'edge_add', 'Y', 'Z', 3.6, 0.217).build()
```
where the last two numbers are the burgers vector and poisson ratio.

pmrv:
> > Nice! Is there a way to include dislocations? doesn't have to be this PR.
> 
> Yes, it's not properly documented yet, but all atomsk flags are supported. For dislocations it would be something like
> 
> ```python
> pr.create.structure.atomsk.create('fcc', 3.6, 'Cu').duplicate(5, 1, 5).dislocation('0.498*box',  '0.500*box', 'edge_add', 'Y', 'Z', 3.6, 0.217).build()
> ```
> 
> where the last two numbers are the burgers vector and poisson ratio.

The pattern is that all method calls on `AtomskBuilder` are translated like
```method_call(a1, a2, ...) -> -method-call a1 a2 ...```

File: pyiron_atomistics/atomistics/structure/factories/atomsk.py [hunk 1]
```diff
@@ -0,0 +1,122 @@
...
+    def create(cls, lattice, a, *species, c=None, hkl=None):
+        """
+        See https://atomsk.univ-lille.fr/doc/en/mode_create.html
+        """
+
+        self = cls()
+
+        a_and_c = str(a) if c is None else f"{a} {c}"
+        line = f"--create {lattice} {a_and_c} {' '.join(species)}"
+        if hkl is not None:
```
liamhuber, line 54:
```suggestion
        if hkl is not None and np.array(hkl).shape != (3, 3):
            raise ValueError('hkl must have the shape 3x3 if provided.')
```

pyiron_atomistics/atomistics/structure/factories/atomsk.py
```diff
@@ -0,0 +1,122 @@
...
+        """
+        self._options.append("-orthogonal-cell")
+        return self
+
+    def build(self):
+        self._options.append("- exyz") # output to stdout as exyz format
+        with tempfile.TemporaryDirectory() as temp_dir:
+            if self._structure is not None:
+                write(os.path.join(temp_dir, "input.exyz"), self._structure, format="extxyz")
+            proc = subprocess.run(["atomsk", *" ".join(self._options).split()],
```
liamhuber, line 105: I think we need to do something about the dependencies. Is it sufficient to conda-install atomsk to get this running,  or would I further need to modify the command line name after installation? Is atomsk available across all architectures? How can we `ImportAlarm` something that's not a python package? Since we're working in `atomistics` and not `contrib` I'd like to get at least some of these answered before we merge...

pyiron_atomistics/atomistics/structure/factories/atomsk.py
```diff
@@ -0,0 +1,122 @@
...
+    Cu: [0.  1.8 1.8]
+    Cu: [1.8 0.  1.8]
+    Cu: [3.6 0.  0. ]
+    Cu: [5.4 1.8 0. ]
+    Cu: [3.6 1.8 1.8]
+    Cu: [5.4 0.  1.8]
+    pbc: [ True  True  True]
+    cell: 
+    Cell([10.2, 3.6, 3.6])
+    """
```
liamhuber, line 165: copypasta examples to unit tests folder? Will need to be excluded from the CI until we get the dependencies better sorted, but should at least be there for local machines to run.

liamhuber (commented):
Very nice, I'm happy (a) this happened so quickly, and (b) that you found an easy and direct way to bind things. A few discussion points follow.

pyiron_atomistics/atomistics/structure/factories/atomsk.py
pmrv: Installing the conda package is enough, but there's only a linux build on conda, not sure if a windows build is possible.  Using `ImportAlarm` is not possible since it's not a python package, but I could modify the `StructureFactory` to only export `atomsk` if it is found in the `PATH`.

pmrv:
Tests and docstrings are in, so this is done from my side.

@zhuochengXIE @LVC14C15 Can you have a look if this meets your needs (or is understandable at all)?

jan-janssen:
> Nice! Is there a way to include dislocations? doesn't have to be this PR.

Did not you do this in https://github.com/eisenforschung/pyiron_mpie/blob/master/pyiron_mpie/interactive/dislocation.py already? I thought we already had an atomsk wrapper, did not we?

pmrv:
I found one place in `pyiron_mpie` that uses atomsk before [here](https://github.com/eisenforschung/pyiron_mpie/blob/7baa1a1aaa3dcde3702f83add4c2eba20fc92bd7/pyiron_mpie/interactive/dislocation.py#L198).  The reason to add the wrapper here was not just dislocations though, but that Eric's group is using it generally a lot.

pmrv:
Previously I used a feature introduced [here](https://github.com/pierrehirel/atomsk/issues/30) that is not yet available in the conda version.  I have therefore rewritting the part calling atomsk to avoid it.

This is now done from my side, unless there's additional feature requests.

jan-janssen:
> Previously I used a feature introduced [here](https://github.com/pierrehirel/atomsk/issues/30) that is not yet available in the conda version. I have therefore rewritting the part calling atomsk to avoid it.

To be future proof I would suggest to update the conda package - I opened a corresponding pull request here https://github.com/conda-forge/atomsk-feedstock/pull/14 - the issues is that they no longer publish tagged releases on Github https://github.com/pierrehirel/atomsk/tags so in case anybody is in contact with the developer that is maybe something we can ask them for.

jan-janssen:
I updated the conda package `0.11` and asked the developer to update their tags on Github https://github.com/pierrehirel/atomsk/issues/35

tests/atomistics/structure/factories/test_atomsk.py
```diff
@@ -0,0 +1,59 @@
+# coding: utf-8
+# Copyright (c) Max-Planck-Institut für Eisenforschung GmbH - Computational Materials Design (CM) Department
+# Distributed under the terms of "New BSD License", see the LICENSE file.
+
+import shutil
+
+from pyiron_atomistics._tests import TestWithProject
+from pyiron_atomistics.atomistics.structure.factories.atomsk import AtomskFactory, AtomskError
+
+if shutil.which("atomsk") is not None:
```
liamhuber: I'm a little uncomfortable hard-coding the executable name (e.g. what if someone has multiple atomsk versions installed and renames them accordingly?) but I don't think it needs to stop the merge.

liamhuber (approved):

tests/atomistics/structure/factories/test_atomsk.py
pmrv: Agree, ideally this would somehow reuse the `Executable` class that we have already for jobs, so you'd be able to configure your particular atomsk, but I'm drawing a blank how to actually do it right now.

liamhuber: Fair!

pmrv:
Latest version is up on conda now.

pmrv:
I've made it so now, that the code is only exported in the `StructureFactory` when the `atomsk` executable is available.

PR #261, opened 2021-06-30
Centrosymmetry

samwaseda:
Now via

```
neigh = structure.get_neighbors()
neigh.centrosymmetry
```

it should be possible to get the centro-symmetry value for the given environment.

Tests and DosString will follow.

### DIFF
--- a/pyiron_atomistics/atomistics/structure/neighbors.py
+++ b/pyiron_atomistics/atomistics/structure/neighbors.py
@@ -585,20 +585,31 @@ class Tree:
             for m in np.arange(-l, l+1)
         ], axis=0))
 
-    @property
-    def centrosymmetry(self):
-        l = self.distances.shape[-1]
-        if l%2!=0:
-            raise AssertionError(
-                'Centrosymmetry parameter can be calculated only if the number of neighbors is even'
-            )
+    @staticmethod
+    def _get_all_possible_pairs(l):
         all_arr = np.array(list(itertools.permutations(np.arange(l),l)))
         all_arr = all_arr.reshape(len(all_arr), -1, 2)
         all_arr.sort(axis=-1)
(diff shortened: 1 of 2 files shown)


liamhuber:
What is the difference compared to `structure.analyze.pyscal_centro_symmetry`? Is it just to avoid writing `num_neighbors` a second time, or does this some how play well with the ragged version of `Neighbors` so that the CSP uses a different number of neighbors for each atom?

samwaseda:
> What is the difference compared to `structure.analyze.pyscal_centro_symmetry`? Is it just to avoid writing `num_neighbors` a second time, or does this some how play well with the ragged version of `Neighbors` so that the CSP uses a different number of neighbors for each atom?

It's attached to `Neighbors`, so you can calculate the value for any point in the box, i.e. something like this is also possible:

```
neigh = structure.get_neighborhood(my_favourite_positions)
print(neigh.centrosymmetry)
```

pyiron_atomistics/atomistics/structure/neighbors.py
```diff
@@ -584,6 +585,36 @@ def get_steinhardt_parameter(self, l, cutoff_radius=np.inf):
             for m in np.arange(-l, l+1)
         ], axis=0))
 
+    @staticmethod
+    def _get_all_possible_pairs(l):
```
samwaseda, line 589: This function works maybe up to `l=10` but after that it takes too much memory and fails. If there's someone who can come up with a nice idea let me know. It in principle does what's described on [this page](https://stackoverflow.com/questions/5360220/how-to-split-a-list-into-pairs-in-all-possible-ways).

samwaseda:
Following the pyiron meeting last week, we might change the internal algorithm, but since for the user the output is not going to change, I merge it now.

PR #262, opened 2021-06-30
Save all VCSGC parameters in generic input

pmrv:
Several parameters to `calc_vcsgc` were not saved before, so… I just do that now.  This is helpful when doing pyiron tables on SGC jobs and you need to know e.g. the chemical potential.

Will convert from draft as soon as I tested it on the cluster.

### DIFF
--- a/tests/lammps/test_base.py
+++ b/tests/lammps/test_base.py
@@ -506,7 +506,7 @@ class TestLammps(unittest.TestCase):
         self.job_vcsgc_input.calc_vcsgc(**args)
         self.assertEqual(self.job_vcsgc_input.input.control['fix___vcsgc'], input_string)
 
-        args['temperature_mc'] = 100.,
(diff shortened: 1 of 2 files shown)


tests/lammps/test_base.py
```diff
@@ -506,7 +506,7 @@ def test_vcsgc_input(self):
         self.job_vcsgc_input.calc_vcsgc(**args)
         self.assertEqual(self.job_vcsgc_input.input.control['fix___vcsgc'], input_string)
 
-        args['temperature_mc'] = 100.,
+        args['temperature_mc'] = 100.
```
pmrv, line 509: @liamhuber This broke the tests.  I don't think this is intentional or is it and I'm missing it?

liamhuber: Just a copy-paste error from a function signature I bet.

PR #270, opened 2021-07-02
Add HasStructure to Trajectory

pmrv:
That would be probably be very useful for #251.

### DIFF
--- a/pyiron_atomistics/atomistics/job/atomistic.py
+++ b/pyiron_atomistics/atomistics/job/atomistic.py
@@ -741,7 +741,7 @@ class Trajectory(HasStructure):
         # new_structure.arrays['cells'] = new_structure.cell
         return new_structure
 
-    def _get_structure(self, frame, wrap_atoms=True):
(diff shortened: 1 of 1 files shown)


pyiron_atomistics/atomistics/job/atomistic.py
```diff
@@ -741,9 +741,15 @@ def __getitem__(self, item):
         # new_structure.arrays['cells'] = new_structure.cell
         return new_structure
 
+    def _get_structure(self, frame, wrap_atoms=True):
```
sudarsan-surendralal: Nit: Codeacy complains that the signature is different from the base class implementation. Setting `frame-1` will get rid of this.

sudarsan-surendralal (approved):
LGTM! Only a minor nit!

pyiron_atomistics/atomistics/job/atomistic.py
pmrv: Good catch!

PR #276, opened 2021-07-06
Use HasStructure for _get_neighbors instead of re-creating structures…

pmrv:
… ourselves

Changes as discussed in #251.  I can't test it right now, because of the jupyter situation, but maybe tomorrow.

### DIFF
--- a/pyiron_atomistics/atomistics/job/atomistic.py
+++ b/pyiron_atomistics/atomistics/job/atomistic.py
@@ -804,10 +804,7 @@ class Trajectory(HasStructure):
         if snapshot_indices is None:
             snapshot_indices = np.arange(len(self), dtype=int)
 
-        n_obj = NeighborsTraj(init_structure=self._structure,
-                              positions=self._positions[snapshot_indices],
-                              cells=self._cells[snapshot_indices],
-                              num_neighbors=num_neighbors, **kwargs)
+        n_obj = NeighborsTraj(self, num_neighbors=num_neighbors, **kwargs)
         n_obj.compute_neighbors()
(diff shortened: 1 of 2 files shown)


sudarsan-surendralal (approved):
Looks good to me! I see this is WIP. Once you are done with this, do merge this with the other `neighbors_traj` branch .

pmrv:
Had some typos left, but seems to work in my simple example.

I tested on ~100 snapshots from a 54 atom lammps run, the original PR takes ~580ms and this takes ~860ms.  It's a bit slower, but I feel the extra generality is worth it.  If you want it to be faster I can have a look at a proper profile.

pmrv:
So I took a profile anyway and it turns the cause is in the original `get_neighbors()`

https://github.com/pyiron/pyiron_atomistics/blob/2a19caebff467f41385aec547573f4b849172e44/pyiron_atomistics/atomistics/structure/neighbors.py#L67

It makes a copy of the structure it analyzes.  Since #251 only copies one structure and then modifies it there's ~101 calls (1 + 100 from the `Neighbors` class) to `Atoms.copy` whereas here `Trajectory._get_neighbors` calls `Atoms.copy` once per iteration there's ~200 calls (100 + 100 from the `Neighbors` class).  

During this I also saw again that the copy is only costly because of the problems with the periodic table already addressed #224, as much as 90% of the time copying a structure is spent loading the periodic table!

So to make this faster again, would be a) merge #224 (pinky promise I'll fix the tests!) and b) @samwaseda is it really necessary to copy the structure for the neighbor analysis?

PR #280, opened 2021-07-14
Equivalent points

### LINKED ISSUE #268: What to use instead of get_symmetry?
`Atoms.get_symmetry` is flagged `@deprecatesoon`, but I can't find any alternatives. What is it that I should be using instead?

samwaseda:
# Most important changes

- All symmetry-related operations are now found in `symmetry.py` (which allowed me to reduce 200 lines in `atoms.py`)
- Initialization via `symmetry = structure.get_symmetry()`, after which the attributes of `symmetry` can be used for different purposed (in the end this follows @jan-janssen's suggestion)
- Original output of `structure.get_symmetry()` is still available (i.e. `structure.get_symmetry()['rotations']` etc.)
- !!! Unit tests are still relatively weak

## Original purpose of this PR (i.e. `get_arg_equivalent_sites()`)

This is very much equivalent to `structure.group_points_by_symmetry`, although the old function didn't work in many cases. Also instead of grouping the sites by lists, the new function returns their ID's.

Example: Group octahedral interstitial sites according to the box symmetries:

```
n_repeat = 2
a_0 = 4.
structure = pr.create.structure.bulk('Al', cubic=True, a=a_0).repeat(n_repeat)
sites = structure.get_wrapped_coordinates(structure.positions+np.array([0, 0, 0.5*a_0]))
del structure[0]
print(structure.get_symmetry().get_arg_equivalent_sites(sites))
```

This returns:

```
[0 1 1 2 1 1 3 2 1 3 1 2 3 3 3 2 0 0 0 2 1 0 1 2 1 1 0 2 3 1 1 2]
```

This should be very much equivalent to the distances between the vacancy and these positions:

```
print(np.round(structure.get_distances_array(v_position, sites), decimals=2))
```

```
[0 2 2 1 2 2 3 1 2 3 2 1 3 3 3 1 0 0 0 1 2 0 2 1 2 2 0 1 3 2 2 1]
```

# Comments

- There are some properties/functions with no DocString -> there I don't know what they exactly do, so I didn't add stuff
- spglib-output was repeated in every function in the former version. This is now unified in `symmetry._get_spglib_cell()`. However, for some reason `structure.get_cell().T` was used for most of the places **except** for `get_spacegroup()` (which is now a property, `spacegroup`). As I have the feeling that `.T` should be removed everywhere, I removed it -> if you know more about this, and if this presents a problem, please let me know.


pyiron_atomistics/atomistics/structure/analyse.py
```diff
@@ -569,3 +569,90 @@ def get_voronoi_vertices(self, epsilon=2.5e-4, distance_threshold=0, width_buffe
...
+        self,
+        points,
+        use_magmoms=False,
+        use_elements=True,
+        symprec=1e-5,
+        angle_tolerance=-1.0,
+        epsilon=1.0e-8
+    ):
+        """
+
```
pmrv:
```suggestion
```

pyiron_atomistics/atomistics/structure/analyse.py
```diff
@@ -569,3 +569,90 @@ def get_voronoi_vertices(self, epsilon=2.5e-4, distance_threshold=0, width_buffe
...
+        points,
+        use_magmoms=False,
+        use_elements=True,
+        symprec=1e-5,
+        angle_tolerance=-1.0,
+        return_unique=True,
+        epsilon=1.0e-8
+    ):
+        """
+
```
pmrv: Missing introduction.

pyiron_atomistics/atomistics/structure/analyse.py
```diff
@@ -569,3 +569,90 @@ def get_voronoi_vertices(self, epsilon=2.5e-4, distance_threshold=0, width_buffe
...
+        R = symmetry_operations['rotations']
+        t = symmetry_operations['translations']
+        x = np.einsum('jk,nj->nk', np.linalg.inv(self._structure.cell), np.atleast_2d(points))
+        x = np.einsum('nxy,my->mnx', R, x)+t
+        if any(self._structure.pbc):
+            x[:,:,self._structure.pbc] -= np.floor(x[:,:,self._structure.pbc]+epsilon)
+        if not return_unique:
+            return np.einsum(
+                'ji,mnj->mni', self._structure.cell, x
+            ).reshape((len(R),)+np.asarray(points).shape)
```
pmrv:
```suggestion
            ).reshape((len(R),)+np.shape(points))
```
exactly the same code-wise, but slightly easier to read.

pyiron_atomistics/atomistics/structure/analyse.py
```diff
@@ -569,3 +569,90 @@ def get_voronoi_vertices(self, epsilon=2.5e-4, distance_threshold=0, width_buffe
...
+        x = x.reshape(-1, 3)
+        _, indices = np.unique(
+            np.round(x, decimals=int(-np.log10(symprec))), return_index=True, axis=0
+        )
+        return np.einsum('ji,mj->mi', self._structure.cell, x[indices])
```
pmrv:
```suggestion
        x = x.reshape(-1, 3)
        if not return_unique:
            indices = slice(0, None)
        else:
            _, indices = np.unique(
                np.round(x, decimals=int(-np.log10(symprec))), return_index=True, axis=0
            )
        return np.einsum('ji,mj->mi', self._structure.cell, x[indices])
```
Can we not do this do unify both branches?

pyiron_atomistics/atomistics/structure/analyse.py
```diff
@@ -569,3 +569,90 @@ def get_voronoi_vertices(self, epsilon=2.5e-4, distance_threshold=0, width_buffe
...
+            symprec=symprec,
+            angle_tolerance=angle_tolerance,
+            epsilon=epsilon,
+            return_unique=False
+        )
+        decimals = int(-np.log10(symprec))
+        _, inverse = np.unique(
+            np.round(all_points.reshape(-1, 3), decimals=decimals), axis=0, return_inverse=True
+        )
+        inverse = inverse.reshape(all_points.shape[:-1][::-1])
```
pmrv: Why not?
```suggestion
        inverse = inverse.reshape(all_points.shape[-2::-1])
```

pyiron_atomistics/atomistics/structure/analyse.py
```diff
@@ -569,3 +569,90 @@ def get_voronoi_vertices(self, epsilon=2.5e-4, distance_threshold=0, width_buffe
...
+            points=points,
+            use_magmoms=use_magmoms,
+            use_elements=use_elements,
+            symprec=symprec,
+            angle_tolerance=angle_tolerance,
+            epsilon=epsilon,
+            return_unique=False
+        )
+        decimals = int(-np.log10(symprec))
+        _, inverse = np.unique(
```
pmrv: I honestly don't understand what this does, or how it relates to the box symmetry, so either a comment or more expanded docstring would be nice.

pmrv (commented):
Just briefly going over, I have to admit that I don't understand the uses cases `get_symmetry()` vs. `get_equivalent_point()` vs. `group_points_by_symmetry()`.  I'll try to have a detailed look again tomorrow, but I do feel this is something the docstring should tell me without reading the source.

pyiron_atomistics/atomistics/structure/analyse.py
samwaseda: This is not the same thing. If not `return_unique`, then the shape will be `n_symmetry x n_input_points x 3`

pmrv: Ah, I didn't get that.  It's ok to leave it then, but the difference in shape should be clearly documented.

PR #281, opened 2021-07-15
Automatically set interactive_enforce_structure_reset, when setting s…

pmrv:
…tructure of different length

Also changes explicit calls to superclass getters/setters to super


samwaseda:
To be honest I have the feeling that we're actually only smothering the consequence of a much larger problem... I guess at some point we should think of how to see what to update more in detail.

pmrv:
I'm open to suggestions, it's just that I feel I shouldn't be bothered to remember typing this all the time. :')

jan-janssen:
But the super calls are not the same.

pmrv:
> But the super calls are not the same.

Yep, teaches me not to do too many things at once. :/ I'll remove those again.

pmrv:
From the meeting: change instead `GenericInteractive.run_if_interactive:133`.

PR #284, opened 2021-07-21
Hotfix: Handle empty indices

sudarsan-surendralal:


samwaseda:
Shouldn't the same apply to `self._cells`? This being said I don't really understand why `self._indices` is allowed to be `None`.

sudarsan-surendralal:
> 
> 
> Shouldn't the same apply to `self._cells`? This being said I don't really understand why `self._indices` is allowed to be `None`.

You're right! It's the cells that need to be handled this way, the indices are never `None`. This was an issue with my subclass!

sudarsan-surendralal:
@samwaseda Shall I merge this now?

PR #291, opened 2021-07-26
Add error message if Murnaghan.plot is called on unfinished jobs

pmrv:
Just a small QoL fix to fail early.

### DIFF
--- a/pyiron_atomistics/atomistics/master/murnaghan.py
+++ b/pyiron_atomistics/atomistics/master/murnaghan.py
@@ -775,6 +775,8 @@ class Murnaghan(AtomisticParallelMaster):
             self._fit_eos_general(fittype=self.input["fit_type"])
 
     def plot(self, num_steps=100, plt_show=True):
+        if not self.status.finished:
+            raise ValueError("Job must be successfully run, before calling this method.")
         try:
             import matplotlib.pylab as plt
         except ImportError:


pyiron_atomistics/atomistics/master/murnaghan.py
```diff
@@ -775,6 +775,8 @@ def collect_output(self):
             self._fit_eos_general(fittype=self.input["fit_type"])
 
     def plot(self, num_steps=100, plt_show=True):
+        if not self.status.finished:
+            raise ValueError("Job must be successfully run, before calling this method.")
```
jan-janssen, line 779: But maybe I want to use the plot method with 10 points if out of my 11 calculation one failed. Meaning I see why you want this while at the same time it might also break the workflow of others.

pmrv: In my case it was more that the calculations hadn't finished yet, so `energy` and `volume` are not in the HDF yet.  What about
```suggestion
        if not (self.status.finished or self.status.not_converged):
            raise ValueError("Job must be successfully run, before calling this method.")
```
In case the job aborts, I suppose there's no point in plotting, since there will also be no data in the HDF.

jan-janssen: My bad - we can keep the initial modification - when the collect is successful the status is set to `finished` https://github.com/pyiron/pyiron_base/blob/master/pyiron_base/master/parallel.py#L458 I mixed it up with my own workflow which is based on pyiron tables.

jan-janssen (approved):

Issue #288, opened 2021-07-25
Benchmark unittest


niklassiemer:
With our new Benchmark unit tests we may run into unintended failures like in the last PR I merged. We should separate them from the normal unit tests such that we are not CI server load dependent for the overal test suite.

niklassiemer:
Fixed in #313

PR #292, opened 2021-07-27
Run benchmark tests in separate step with continue-on-error true

niklassiemer:


pmrv:
Are the results still visible with this or do we always have to manually check then if they passed or not?  I understand that #288 is annoying, but if they actually fail for real it should be very obvious.

niklassiemer:
The main difference is actually only that the other tests are not stopped due to a failure in the benchmarks. The benchmark tests are now an additional step in the unittests (we do not need to rebuild). For the decision if it is a fail, the overall workflow is judged and, therefore, it should after result in an red X in our overview... (I would prefer a yellow !)

pmrv:
> The main difference is actually only that the other tests are not stopped due to a failure in the benchmarks. The benchmark tests are now an additional step in the unittests (we do not need to rebuild). For the decision if it is a fail, the overall workflow is judged and, therefore, it should after result in an red X in our overview... (I would prefer a yellow !)

Sounds good then!

niklassiemer:
Converted to draft, since I want to resolve the windows CI issue first.

niklassiemer:
Closed in favor of #313.

PR #293, opened 2021-07-27
Just testing to only skip the cache test 

niklassiemer:


niklassiemer:
@pmrv This seems to be unrelated to the Windows not working issue and is not present in master, however, I do not understand how skipping one particular test breaks the overall tests?! Do you have any idea what is going on here and how to fix this?

PR #295, opened 2021-07-28
Interactive units

sudarsan-surendralal:
Ensures that unit conversions are proper even for interactive jobs

### DIFF
--- a/notebooks/integration/interactive_units.ipynb
+++ b/notebooks/integration/interactive_units.ipynb
@@ -1,9 +1,15 @@
 {
  "cells": [
   {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "This notebooks compares the outputs from two static LAMMPS jobs one of which is an interactive job. This is to ensure that the outputs from interactive and non-interactive jobs are consistent."
+   ]
+  },
+  {
    "cell_type": "code",
    "execution_count": 1,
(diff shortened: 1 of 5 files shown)


pyiron_atomistics/lammps/base.py
```diff
@@ -61,6 +61,28 @@ def __init__(self, project, job_name):
...
+        """
+        if self.input.control["units"] is not None:
+            return self.input.control["units"]
+        else:
+            # Default to metal units
+            return "metal"
+
+    @units.setter
+    def units(self, val):
+        allowed_types = ['metal', 'real', 'si', 'cgs', 'lj']
```
jan-janssen: Having the static list defined here is dangerous, as we might forget to update it later. Can you use `LAMMPS_UNIT_CONVERSIONS.keys()` ?

notebooks/integration/interactive_units.ipynb
```diff
@@ -0,0 +1,161 @@
+{
```
jan-janssen, line 1: An explanation in the notebook would be nice.

sudarsan-surendralal: I added some text in the beginning of the notebook!

sudarsan-surendralal:
@jan-janssen shall I merge this for now?

PR #298, opened 2021-07-29
Just testing to get Windows CI running again

niklassiemer:
Revert to be279d0d52d501b3b6683bbda39c3ca18cd66ba3

### DIFF
--- a/.ci_support/environment-notebooks.yml
+++ b/.ci_support/environment-notebooks.yml
@@ -1,14 +1,14 @@
 channels:
-- conda-forge
+  - conda-forge
 dependencies:
-- gpaw >=20
-- jupyter
-- lammps >=2021.05.27
-- nglview
-- papermill
-- seaborn
-- sphinxdft >=2.7.0
-- sphinxdft-data
-- iprpy-data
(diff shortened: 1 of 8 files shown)


pmrv:
If I remember correctly the windows CI ended up taking hours and got cancelled before I merged it.  I had not investigated why though.  Since I don't have a windows machine accessible can you check whether running the tests locally on a windows machine takes as much time as on CI?

niklassiemer:
No they run fast for me. I will update my windoof test suite conda again to make sure this is not related to some other package being updated. 
However, I got a file not found error on "tests/static/lammps_test_files/get_structure_test.tar.gz", probably since I started the tests in the wrong directory. We should make this independent, though.
Another issue, we should run the integration tests in `pyiron_base` with all OS not only linux, I will make a PR for that.

pmrv:
Aha, no I think I didn't include those files in #241. :/

EDIT: No `get_structure_test.tar.gz` is correctly located in `test/static/lammps_test_files`.

niklassiemer:
Might be, however, I tested with the current master and I have these files... In addition, the windows tests should then just fail and not run forever!?

PR #302, opened 2021-07-30
Debug failing windows tests 

niklassiemer:


jan-janssen:
At least we now have the confirmation that it is again the non modal Murnaghan job which fails.

niklassiemer:
Not yet 100%. It also might be that I now checked out the head of this pr instead of the merge with master to test. Therefore, it might be that the non modal test would run. I will test that as well... if it runs with the non modal one, than it is somewhere in the last PRs...

niklassiemer:
There we go, running the CI on our latest release does not break the windows tests (Failing notebook fixed in https://github.com/pyiron/pyiron_atomistics/pull/304) ! Thus the problem _is_ in one of the PRs  

#299 #241 #291 #287 #224 #285 #280

PR #305, opened 2021-07-31
Revert "Make Atoms() faster via caches"

niklassiemer:
Reverts pyiron/pyiron_atomistics#224: Fix Windows tests


niklassiemer:
Did not help.

PR #313, opened 2021-08-03
Introduce benchmark tests as additional workflow

niklassiemer:


niklassiemer:
I now deactivated the benchmark test in the usual test suite by omitting the actual testing. I tried to remove the test, however, for some reason this [broke the CI](#293)?!

niklassiemer:
Right now, this 100% feels like a waste of resources to build the environment (~5min with mamba on linux, ~6min Mac, >12min on windows) for one test which lasts less than 10 seconds right now. 
On the other hand it is more convenient to see the red X at the benchmarks instead of at the unittests in case the benchmark fails.

PR #327, opened 2021-08-25
Move StructureStorage from pyiron_contrib

pmrv:
Simply moves the files from contrib to here.


pmrv:
Broken tests are again related to sphinx, so I'll just ignore it for now.

PR #334, opened 2021-08-31
Update to pyiron_base 0.3.1

niklassiemer:


niklassiemer:
```
ERROR: setUpClass (pyiron_base._tests.TestWithProject)
  File "/usr/share/miniconda/envs/test/lib/python3.9/site-packages/pyiron_base/_tests.py", line 64, in setUpClass
    cls.project = Project(cls.project_path)
  File "/usr/share/miniconda/envs/test/lib/python3.9/site-packages/pyiron_base/project/generic.py", line 118, in __init__
    super(Project, self).__init__(path=path)
  File "/usr/share/miniconda/envs/test/lib/python3.9/site-packages/pyiron_base/project/path.py", line 224, in __init__
    generic_path = self._convert_str_to_generic_path(path)
  File "/usr/share/miniconda/envs/test/lib/python3.9/site-packages/pyiron_base/project/path.py", line 377, in _convert_str_to_generic_path
    root_path, project_path = self._get_project_from_path(path)
  File "/usr/share/miniconda/envs/test/lib/python3.9/site-packages/pyiron_base/project/path.py", line 410, in _get_project_from_path
    root = Settings().top_path(full_path)
  File "/usr/share/miniconda/envs/test/lib/python3.9/site-packages/pyiron_base/settings/generic.py", line 306, in top_path
    raise ValueError(
ValueError: the current path /usr/share/miniconda/envs/test/lib/python3.9/site-packages/pyiron_base/_tests/ is not included in the .pyiron configuration. ['/home/runner/work/pyiron_atomistics/pyiron_atomistics/']
```
I am wondering why we try to run that test?!

niklassiemer:
Maybe, since it is derived from `unittest.TestCase` and it now has a `test_`something (i.e. `test_docstring`) method and is imported while running the `pyiron_atomistics` test suite our template tests are actually tested although they are `ABC`? If this should be the case, we have to `unittest.skip()` our template test classes to not run them?!

liamhuber:
I'm not even clear in what test this is actually being invoked? The workflow should only be discovering tests in `pyiron_atomistics/tests`, not in `pyiron_atomistics/pyiron_atomistics/tests_`. I'm quite lost.

liamhuber:
Ok, I think I get it, it's literally saying it's running the test `TestWithProject` in `pyiron_base._tests`. So now I'm up to speed with you, but still just as confused why it would try to do that to begin with.

liamhuber:
My understanding must just be wrong. I thought 

```python
run: coverage run -m unittest discover tests
```

uses `unittest` to look in `./tests` for classes derived from `TestCase`. But actually our workflow explicitly says:

```python
run: coverage run --omit pyiron_atomistics/_version.py -m unittest discover tests
```

So we're force-ignoring `./pyiron_atomistics/_version.py`, which is outside `./tests`. Therefore I didn't get it.

I bet adding `--omit pyiron_base/_test.py` will resolve this -- but why didn't we run into the same problem on base?

niklassiemer:
> I bet adding `--omit pyiron_base/_test.py` will resolve this -- but why didn't we run into the same problem on base?

I am not so sure about that. `./pyiron_atomistics/_version.py` is a file which is actually present in the checked out repo, while `pyiron_base` is only installed from conda... The reason we did not have problems in base is the `.pyiron` configuration in base which allows pyiron to create a `Project` in the repository structure. However, we do not allow to make projects under the conda installation path :)

niklassiemer:
I just read https://stackoverflow.com/questions/4566910/abstract-test-case-using-python-unittest/4566992
and it seems that all classes derived from `unittest` which are imported will be tested. They mention a few ideas how to avoid the execution of the base class itself. One is just to 
`del BaseTestClass` at the end of the module such that the actual class is no longer around, but I actually do not know if this works. Also a plain `@unittest.skip()` to the ABC could work if that is not inherited?

niklassiemer:
Once pyiron_base is uploaded on conda, I will start a new PR from the pyiron_base-0.3.2 branch.

PR #337, opened 2021-09-02
Run tests with specific pyiron_base branch

niklassiemer:


niklassiemer:
`@unittest.skip()` is inherited, thus derived tests are also skipped.

niklassiemer:
Checking `cls = BaseClass` and raising the `unittest.SkipTest` Error seems to work.

PR #338, opened 2021-09-02
Consistent atoms order

### LINKED ISSUE #335: How to ensure consistent indices when reading from POSCAR?
I'm reading a number of POSCAR files to run some lammps on it.  The issue is that even though the elements are always defined in the same order in the POSCAR, the association `index <-> element symbols` is seemingly random, when I read the files with either `pr.create.structure.ase.read` or `pyiron_atomistics.vasp.structure.read_atoms`.

If this is coming from ASE then I suppose there's not a lot we can do about it, but is there a way to change what indices are associated with what symbol after an `Atoms` instance is created?  I've tried setting `atoms.indices`, but this changes the the symbol of that changed atom accordingly (as it should).

sudarsan-surendralal:
Create a consistent order of the atoms and corresponding indices when creating atoms


pmrv:
It seems to work, but I'm still a bit stuck on how.  Will give it another reading later. :/

EDIT: Also according to what do we want to sort?  The examples I gave in #335 sort alphabetically, but the test gives 'Ca', 'Mg', 'Al'.

liamhuber:
For maximum safety, shouldn't the test be verifying that species get mapped to the desired cartesian positions?

niklassiemer:
> It seems to work, but I'm still a bit stuck on how. Will give it another reading later. :/
> 
> EDIT: Also according to what do we want to sort? The examples I gave in #335 sort alphabetically, but the test gives 'Ca', 'Mg', 'Al'.

That sorting also does not make sense to me. I would either sort alphabetically or by atomic number.

sudarsan-surendralal:
The random ordering of the species was due to this line:
```python
self.set_species(list(set(el_object_list)))
```
When you create a set from a list of objects (not symbols) the sorting of the set is random which is why you get the random ordering of the species. My current implementation sorts the species depending on the order of the elements defined and not the alphabetical order. For example, if you define:

```python
struct = Atoms(elements=["Ca", "Mg", "Al". "Ca"], positions=<positions>, cell=<cell>)
print(struct.get_chemical_symbols()
``` 
You would get `["Ca", "Mg", "Al"]` which makes sense to me.  `struct.get_species_symbols()` would give you the alphabetically ordered symbols  `["Al", "Ca", "Mg"]`.

pmrv:
So the order of the original `elements`? 

I'd rather like one fixed mapping between symbols and indices, because if I create multiple structures independently and then put them in an interactive job, structure or training container, I don't want to keep track of this order manually.  If we sort alphabetically or by atomic number then it's always clear how to make the indices consistent if you add a structure to an existing set of structures. 

I have a working workaround though, so we can discuss on Monday with everyone.

sudarsan-surendralal:
> 
> 
> So the order of the original `elements`?

Yes.
> 
> I'd rather like one fixed mapping between symbols and indices, because if I create multiple structures independently and then put them in an interactive job, structure or training container, I don't want to keep track of this order manually. If we sort alphabetically or by atomic number then it's always clear how to make the indices consistent if you add a structure to an existing set of structures.
> 
> I have a working workaround though, so we can discuss on Monday with everyone.

Yeah let's do that!

pmrv:
Check in interactive jobs, they solve the same problem.

pmrv (approved):
This is ok from my end.  I might need a method at some point to reorder the indices, but I can add that once I'm there.

pyiron_atomistics/vasp/vasprun.py
```diff
@@ -644,6 +644,8 @@ def get_initial_structure(self):
             el_list = self.vasprun_dict["atominfo"]["species_list"]
             cell = self.vasprun_dict["init_structure"]["cell"]
             positions = self.vasprun_dict["init_structure"]["positions"]
+
+            # print("vasprun elements", el_list)
```
liamhuber: Nit: remove dead debug print?

liamhuber (approved):

PR #339, opened 2021-09-03
Don't specify force tolerance if only optimizing cell

pmrv:
For `ISIF >= 5` only the cell parameters are allowed to change, so the atomic forces don't naturally decrease, hence they should not be used as a stopping criterion.


sudarsan-surendralal (approved):
I haven't used `ISIF >= 5` but this LGTM!

pmrv:
It's set when passing `volume_only` to `VaspBase.calc_minimize` which is where it tripped me up.

PR #346, opened 2021-09-06
Update base.py

JNmpi:
Lammps may not keep the original order of the atom IDs (see e.g. https://matsci.org/t/atom-id-is-changing/31701). I experienced this when running reverse scaling where multple runs are appended. The following patch orders the atom IDs and thus guarantees that the order remains unchanged.

### DIFF
--- a/pyiron_atomistics/lammps/base.py
+++ b/pyiron_atomistics/lammps/base.py
@@ -983,7 +983,7 @@ class LammpsBase(AtomisticGenericJob):
                 pd.read_csv(
                     StringIO("\n".join(dump[llst:llen]).replace("ITEM: ATOMS ", "")),
                     delim_whitespace=True,
-                ).sort_values(by='id')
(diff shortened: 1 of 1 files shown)


pyiron_atomistics/lammps/base.py
```diff
@@ -983,7 +983,7 @@ def collect_dump_file(self, file_name="dump.out", cwd=None):
                 pd.read_csv(
                     StringIO("\n".join(dump[llst:llen]).replace("ITEM: ATOMS ", "")),
                     delim_whitespace=True,
-                )
+                ).sort_values(by='id')
```
pmrv:
```suggestion
                ).sort_values(by="id", ignore_index=True)
```
This simply makes sure that the pandas index is ascending.  Otherwise there might be issues if someone later indexes the `DataFrame` naively.

pmrv (approved):

PR #347, opened 2021-09-07
Bind creator at import

liamhuber:
A prototype for how we might dynamically extend the `Project` class as pyiron submodules get imported.

Depends on [pyiron_base#427](https://github.com/pyiron/pyiron_base/pull/427).


pyiron_atomistics/toolkit.py
```diff
@@ -0,0 +1,106 @@
...
+__maintainer__ = "Liam Huber"
+__email__ = "huber@mpie.de"
+__status__ = "production"
+__date__ = "Sep 7, 2021"
+
+
+class JobFactory(JobFactoryCore):
+    @property
+    def _job_class_dict(self) -> dict:
+        return {
```
pmrv, line 57: So this imports all job types at `pyiron_atomistic` import time?  Can we check the effect on the start up time?  I remember me and @jan-janssen trying something similar last year and it increasing startup time a lot.

liamhuber: No effect, but maybe I'm not testing the way you imagine.

Setup:
```python
from time import time
t0 = time()
from pyiron_base import Project
import pyiron_atomistics
# from pyiron_atomistics import Project
dt = time() - t0
print(dt)
```

Restart kernel and re-execute repeatedly.

With `base:creator_registration` and `atomistics:direct_binding`, times were `[4.79, 3.91, 2.65, 2.84, 2.49, 2.61]`.

With `base:master` and `atomistics:master`, times were `[3.85, 2.57, 2.64, 2.67, 2.67, 2.64]`.

On master I also re-tried just with the `from pyiron_atomistics import Project` and it was similar around 2.6s. I guess the initial drop in times in each case has to do with my laptop getting a bit more clever about caching. Overall I don't see a difference. Lmk if I'm testing wrong though.

pmrv: No, sounds good, I'm happy then.

liamhuber:
nm, waiting on pyiron_base 0.3.3

liamhuber:
waiting on pyiron_base 0.3.3

PR #350, opened 2021-09-08
change pointer to copy

### LINKED ISSUE #341: Should `Atoms.get_chemical_indices()` return pointer?
I find the following lines a bit counterintuitive.

```python
>>> structure = pr.create.structure.bulk('Al', cubic=True)
>>> structure[0] = 'Fe'
>>> indices = structure.get_chemical_indices()
>>> print(indices)
>>> structure[:] = 'Al'
>>> print(indices)
[1 0 0 0]
[0 0 0 0]
```

I would rather expect that `Atoms.get_chemical_indices()` returns a copy of the indices and not the pointer, also because changing `indices` has (for some reason) no effect on the atom species. But before I start changing this I just wanted to make sure that this is not the desired behaviour. @sudarsan-surendralal @pmrv @jan-janssen

samwaseda:
Closes #341

### DIFF
--- a/pyiron_atomistics/atomistics/structure/atoms.py
+++ b/pyiron_atomistics/atomistics/structure/atoms.py
@@ -720,7 +720,7 @@ class Atoms(ASEAtoms):
             numpy.ndarray: A list of chemical indices
 
         """
-        return self.indices
+        return self.indices.copy()
 
     def get_atomic_numbers(self):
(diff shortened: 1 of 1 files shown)


samwaseda:
Since this has been discussed in #341, I'm going to merge it as soon as the tests pass

PR #348, opened 2021-09-07
Fix import in integration tests

pmrv:
Was pyiron instead of pyiron_atomistics

PR #352, opened 2021-09-08
Update integration.yml

sudarsan-surendralal:
Currently, integration notebook tests run only when one creates a pull request and sets the label `integration`. This means in most of the cases, these tests never run and problems with these notebooks are hard to detect and debug.  Even in a PR with such a label, this runs only once when creating the PR and not when further commits are made to the branch (see https://github.com/pyiron/pyiron_atomistics/pull/348). I understand that these tests only run when needed (for testing time reasons) but there has to be a better criteria to trigger these tests. In this implementation, the tests run when changes are made to the integration notebooks.


max-hassani:
I have an issue with this strategy. I believe the integration tests should be run when they are needed and labeled, not only when some changes are made to the proposed notebooks. 
I understand they must run on each push, and not only once, but I guess there should be better solutions.

.github/workflows/integration.yml
```diff
@@ -5,11 +5,15 @@ name: Notebooks for Integration Tests
 
 on:
   pull_request:
-    types: [ labeled ]
+      paths:
```
max-hassani: I think if we modify here as:
```
on:
  push:
    branches: [ master ]
  pull_request:
    branches: [ master ]
```
Then keeping the condition `if: ${{ github.event.label.name == 'integration' }}` should be sufficient to trigger the test on every push, if the event label is 'integration`.

max-hassani: I am not 100% sure, though. Therefore, it worths testing.

pmrv:
Maybe we can unify this and automatically label every PR that touches notebooks/integration as `integration`.

sudarsan-surendralal:
Okay so based on the comments so far, we need to modify this configuration file such that the integration tests run when:

1. The label "integration" is used in the PR. The tests should run not only when we create the PR, but also when subsequent commits are made to that particular PR
2. Automatically when changes are made to the integration notebooks directory
3. My suggestion: Also when we merge to the master branch

I'll try to modify the yaml file. I'm not a GitHub Actions expert so I look forward to your suggestions.

pmrv:
I've added automatic labels for PRs that touch notebooks [here](https://github.com/pyiron/pyiron_atomistics/pull/361).  Still can't figure out how to re-run them on subsequent pushes.

niklassiemer:
@sudarsan-surendralal There is an conflict with the notebook you edited - from the github workflow I am happy with this PR.

niklassiemer:
It also ran the integration tests on master 🎉

Issue #349, opened 2021-09-08
Units inconsistent with latest lammps Version


jan-janssen:
@sudarsan-surendralal Currently the unit tests fail for the new units notebook, for example here https://github.com/pyiron/pyiron_atomistics/pull/348 . Can you take a look? It might be related to the latest LAMMPS version. 

```
Input Notebook:  notebooks/integration/interactive_units.ipynb
20
Output Notebook: notebooks/integration/interactive_units-out.ipynb
21

22
Executing:   0%|          | 0/8 [00:00<?, ?cell/s]Executing notebook with kernel: python3
23

24
Executing:  12%|█▎        | 1/8 [00:01<00:07,  1.14s/cell]
25
Executing:  25%|██▌       | 2/8 [00:02<00:09,  1.53s/cell]
26
Executing:  50%|█████     | 4/8 [00:03<00:02,  1.65cell/s]
27
Executing:  62%|██████▎   | 5/8 [00:03<00:01,  1.73cell/s]
28
Executing:  62%|██████▎   | 5/8 [00:04<00:02,  1.17cell/s]
29
Traceback (most recent call last):
30
  File "/usr/share/miniconda/envs/test/bin/papermill", line 10, in <module>
31
    sys.exit(papermill())
32
  File "/usr/share/miniconda/envs/test/lib/python3.9/site-packages/click/core.py", line 1137, in __call__
33
    return self.main(*args, **kwargs)
34
  File "/usr/share/miniconda/envs/test/lib/python3.9/site-packages/click/core.py", line 1062, in main
35
    rv = self.invoke(ctx)
36
  File "/usr/share/miniconda/envs/test/lib/python3.9/site-packages/click/core.py", line 1404, in invoke
37
    return ctx.invoke(self.callback, **ctx.params)
38
  File "/usr/share/miniconda/envs/test/lib/python3.9/site-packages/click/core.py", line 763, in invoke
39
    return __callback(*args, **kwargs)
40
  File "/usr/share/miniconda/envs/test/lib/python3.9/site-packages/click/decorators.py", line 26, in new_func
41
    return f(get_current_context(), *args, **kwargs)
42
  File "/usr/share/miniconda/envs/test/lib/python3.9/site-packages/papermill/cli.py", line 250, in papermill
43
    execute_notebook(
44
  File "/usr/share/miniconda/envs/test/lib/python3.9/site-packages/papermill/execute.py", line 122, in execute_notebook
45
[...]

PR #353, opened 2021-09-08
Explicitly specify water potential

sudarsan-surendralal:
Could possible fix the failing integration notebooks!

### DIFF
--- a/.github/workflows/notebooks.yml
+++ b/.github/workflows/notebooks.yml
@@ -1,37 +1,37 @@
 # This workflow will install Python dependencies, run tests and lint with a variety of Python versions
 # For more information see: https://help.github.com/actions/language-and-framework-guides/using-python-with-github-actions
 
-name: Notebooks
+# name: Notebooks
 
-on:
-  push:
-    branches: [ master ]
-  pull_request:
-    branches: [ master ]
+# on:
+#   push:
+#     branches: [ master ]
+#   pull_request:
+#     branches: [ master ]
 
-jobs:
-  build:
+# jobs:
+#   build:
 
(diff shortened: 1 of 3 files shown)


sudarsan-surendralal:
It turns out that the integration tests fail because the TIP3P water potential, which is now explicitly specified in the notebook, is not available:

```python
~/work/pyiron_atomistics/pyiron_atomistics/pyiron_atomistics/atomistics/job/potentials.py in find_by_name(self, potential_name)
     74         mask = (self._potential_df["Name"] == potential_name)
     75         if not mask.any():
---> 76             raise ValueError("Potential '{}' not found in database.".format(
     77                              potential_name))
     78         return self._potential_df[mask]
ValueError: Potential 'H2O_tip3p' not found in database.
```

I think the reason is that these potentials are not included in [pyiron-resources](https://github.com/pyiron/pyiron-resources).  @jan-janssen, this also explains https://github.com/pyiron/pyiron_atomistics/issues/349. When the water potentials are not found, pyiron tries to search for alternate potentials with `H` and `O` which are some of the REAXFF potentials which are developed for oxides and adsorbates that happend to have hydrogen and oxygen but not for water molecules. But to even use these potentials, explicit charge has to be specified for each of the atoms. Which means we need to reevaluate how we automatically assign potentials in pyiron!

jan-janssen:
@sudarsan-surendralal Can not you define the water potential, just like you did in https://github.com/pyiron/pyiron_atomistics/blob/master/notebooks/water_MD.ipynb ?

sudarsan-surendralal:
Thanks this should work now!

PR #354, opened 2021-09-09
VASP doesn't save stresses to HDF when run on ISIF=2 (default)

pmrv:
I'm running some VASP calculations and would like to have the full stress tensor after the (static) calculation.  I've checked that my `OUTCAR` does contain the information.  (Let me know if you need an example, github doesn't let me upload tar archives).

I've traced the code and it seems we do parse the stresses, but then only save the averaged pressure.  I've modified that code in this PR to save also the stress to the `log_dict`, which I thought should be enough to get it into HDF5, but from my quick testing that doesn't seem to be the case.

Any ideas where to put this so it works?

### DIFF
--- a/pyiron_atomistics/vasp/base.py
+++ b/pyiron_atomistics/vasp/base.py
@@ -1946,6 +1946,7 @@ class Output:
                 vasprun_working = True
         if outcar_working:
             log_dict["temperature"] = self.outcar.parse_dict["temperatures"]
+            log_dict["stresses"] = self.outcar.parse_dict["stresses"]
             log_dict["pressures"] = self.outcar.parse_dict["pressures"]
             log_dict["elastic_constants"] = self.outcar.parse_dict["elastic_constants"]
             self.generic_output.dft_log_dict["n_elect"] = self.outcar.parse_dict[
--- a/pyiron_atomistics/vasp/base.py
+++ b/pyiron_atomistics/vasp/base.py
@@ -2017,6 +2017,7 @@ class Output:
                 raise VaspCollectError("Error in parsing OUTCAR")
             log_dict["energy_tot"] = self.outcar.parse_dict["energies"]
             log_dict["temperature"] = self.outcar.parse_dict["temperatures"]
+            log_dict["stresses"] = self.outcar.parse_dict["stresses"]
             log_dict["pressures"] = self.outcar.parse_dict["pressures"]
             log_dict["forces"] = self.outcar.parse_dict["forces"]
             log_dict["positions"] = self.outcar.parse_dict["positions"]
--- a/pyiron_atomistics/vasp/outcar.py
+++ b/pyiron_atomistics/vasp/outcar.py
@@ -91,7 +91,7 @@ class Outcar(object):
         self.parse_dict["fermi_level"] = fermi_level
         self.parse_dict["scf_dipole_moments"] = scf_moments
         self.parse_dict["kin_energy_error"] = kin_energy_error
-        self.parse_dict["stresses"] = stresses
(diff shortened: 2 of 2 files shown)


pyiron_atomistics/vasp/base.py
```diff
@@ -2017,6 +2017,7 @@ def collect(self, directory=os.getcwd(), sorted_indices=None):
                 raise VaspCollectError("Error in parsing OUTCAR")
             log_dict["energy_tot"] = self.outcar.parse_dict["energies"]
             log_dict["temperature"] = self.outcar.parse_dict["temperatures"]
+            log_dict["stresses"] = self.outcar.parse_dict["stresses"]
```
sudarsan-surendralal, line 2021: Since this is inside an `elif` statement, this segment of the code would work only when vasprun files are absent. You would need to put this in if statement block:

https://github.com/pyiron/pyiron_atomistics/blob/d1d3ed1cb7657633c5a67703f5b32f4a8d8ef200/pyiron_atomistics/vasp/base.py#L1947

sudarsan-surendralal (changes requested):

pmrv:
This works, but the pressures come out in units of `kB`. I'll have to check first how to convert.

samwaseda:
Just for me to understand: Is stress a per-atom quantity? Or is it simply meant to be pressure with the opposite sign?

pmrv:
> Just for me to understand: Is stress a per-atom quantity? Or is it simply meant to be pressure with the opposite sign?

Here stress is the full tensor in voigt notation, what is saved as `pressures` is [the average stress](https://github.com/pyiron/pyiron_atomistics/blob/190eeecc791b22ec2c21b0597f03b574226522e6/pyiron_atomistics/vasp/outcar.py#L105).

samwaseda:
> Here stress is the full tensor in voigt notation, what is saved as `pressures` is [the average stress](https://github.com/pyiron/pyiron_atomistics/blob/190eeecc791b22ec2c21b0597f03b574226522e6/pyiron_atomistics/vasp/outcar.py#L105).

OH! That's a suuuuuuuuuper questionable definition. At least it's not compatible with the LAMMPS definition at all. What's actually behind the idea of having the average pressure in the first place, instead of having nothing or the full tensor?

pmrv:
@sudarsan-surendralal Are you aware of anyone in our group that extensively uses VASP "pressures" via pyiron?  I'd like the output of lammps and vasp to be consistent, which in this case would be saving the full 3x3 matrix under "pressures".  For know I can leave the field in VASP untouched, but it'd be good to identify a time window after which we can make the breaking change "stress" -> "pressures".

sudarsan-surendralal:
> 
> 
> @sudarsan-surendralal Are you aware of anyone in our group that extensively uses VASP "pressures" via pyiron? I'd like the output of lammps and vasp to be consistent, which in this case would be saving the full 3x3 matrix under "pressures". For know I can leave the field in VASP untouched, but it'd be good to identify a time window after which we can make the breaking change "stress" -> "pressures".

Sorry, I'm not aware if anyone uses pressures from VASP (maybe @jan-janssen). I'm not too familiar with the definitions of stress and pressure in the context of DFT simulations but I remember reading (from [this book](https://www.cambridge.org/core/books/electronic-structure/DDFE838DED61D7A402FDF20D735BC63A)) that the definitions are not that straightforward at least wrt. LAMMPS. I have the feeling that this needs to be discussed in our developers meeting or even in a purely scientific meeting since there could be non-pyiron users who have a better idea.

pmrv:
I'll merge this today as-is and work around the inconsistent naming on my end for a while, but we should be this on the radar.

PR #355, opened 2021-09-10
replace strain by master job name

samwaseda:
As all child jobs start with `strain_` it wasn't possible to see the content via `pr.load` whenever there were more than 1 Murnaghan job. Now it appends the name of the master job instead.

And I also removed the default job name because it makes little sense.


jan-janssen (approved):
look good to me

PR #361, opened 2021-09-20
Add automatic labeling for integration tests

pmrv:
This should add the label 'integration' to all PRs that modify
integration notebooks.

Imo this makes #352 unnecessary.

(Was fighting with github for a bit in there, hope no one got spammed too much.)


sudarsan-surendralal (approved):
I just made a dummy commit to check this and it works! Thank you @pmrv, this LGTM.

PR #365, opened 2021-09-23
Support full style without bonds

jan-janssen:
To support full style in LAMMPS without requiring bonds. To automatically generate water bonds set:
```
job.interactive_water_bonds = True
```


sudarsan-surendralal (approved):
Thanks this looks good to me. I'll already merge this with the interactive notebook branch

PR #376, opened 2021-09-28
Don't conflict with HasHDF in StructureStorage

pmrv:
Drops support for legacy storage format.

afaik, the old format of saving `StructureStorage` to HDF is no longer used and therefore drop it,  could this bite you in some way @Leimeroth ?

Depends #375

### DIFF
--- a/pyiron_atomistics/atomistics/structure/structurestorage.py
+++ b/pyiron_atomistics/atomistics/structure/structurestorage.py
@@ -205,22 +205,5 @@ class StructureStorage(FlattenedStorage, HasStructure):
         return len(self)
 
 
-    def to_hdf(self, hdf, group_name="structures"):
-        # just overwrite group_name default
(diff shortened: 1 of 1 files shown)


Leimeroth:
I think I loaded all jobs that used this and rewrote hdf files using to_hdf() manually at some point, but I will have to check.
Even if there are some problems I guess it would still be possible to load and rewrite old jobs using an older version if you merge this and  than switch back to the new version or am I overlooking something?

pmrv:
> I think I loaded all jobs that used this and rewrote hdf files using to_hdf() manually at some point, but I will have to check.
> Even if there are some problems I guess it would still be possible to load and rewrite old jobs using an older version if you merge this and than switch back to the new version or am I overlooking something?

Yeah, downgrading pyiron temporarily, rewriting and then upgrading should be no problem.

PR #386, opened 2021-10-07
Adapt to removal of load_object

pmrv:
Depends on [this](https://github.com/pyiron/pyiron_base/pull/462).


niklassiemer:
This should be independent, shouldn't it? It is more like the other way round. If we remove this in base this method will break. However, the `to_object` functionality should already be fine?

pmrv:
> This should be independent, shouldn't it? It is more like the other way round. If we remove this in base this method will break. However, the `to_object` functionality should already be fine?

Fair, this is actually applicable already.

PR #387, opened 2021-10-07
Use FlattenedStorage in NeighborsTrajectory

pmrv:
This also allows it to be used with grand-canonical trajectories.  Has some small fixes for DataContainer compat too, see commit messages.

### DIFF
--- a/pyiron_atomistics/atomistics/job/atomistic.py
+++ b/pyiron_atomistics/atomistics/job/atomistic.py
@@ -826,7 +826,7 @@ class Trajectory(HasStructure):
         if snapshot_indices is None:
             snapshot_indices = np.arange(len(self), dtype=int)
 
-        n_obj = NeighborsTrajectory(self[snapshot_indices], num_neighbors=num_neighbors, **kwargs)
+        n_obj = NeighborsTrajectory(has_structure=self[snapshot_indices], num_neighbors=num_neighbors, **kwargs)
         n_obj.compute_neighbors()
         return n_obj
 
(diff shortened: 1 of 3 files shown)


sudarsan-surendralal (commented):
This looks good to me. If possible, can you also add a test to check if the shapes are consistent?

pmrv:
I realize now that this will change the shape of the attributes `neighbor_indices`, etc. because you can't put them in full numpy arrays anymore.  (Actually right now the code stores the flat arrays, but that probably breaks you old code).  

What if I make it so that the property methods check whether the trajectory canonical or grand-canonical and in the latter case return full arrays with filler values (basically like @samwaseda does it for the neighbor class).  Then old code won't break and new code that wants to deal with grand-canonical trajectories is better off dealing with the underlying `FlattenedStorage` anyway.

sudarsan-surendralal (approved):
LGTM!

sudarsan-surendralal (changes requested):
Oops one of the tests fail

```
AttributeError: 'float' object has no attribute 'sqrt'

The above exception was the direct cause of the following exception:

  File "/home/runner/work/pyiron_atomistics/pyiron_atomistics/tests/lammps/test_base.py", line 412, in test_dump_parser_water
    self.assertTrue(np.allclose(np.linalg.norm(neigh_traj_obj.vecs, axis=-1),
  File "<__array_function__ internals>", line 5, in norm
  File "/usr/share/miniconda/envs/test/lib/python3.9/site-packages/numpy/linalg/linalg.py", line 2561, in norm
    return sqrt(add.reduce(s, axis=axis, keepdims=keepdims))
TypeError: loop of ufunc does not support argument 0 of type float which has no callable sqrt method
```

pmrv:
Seems it's a bug in `FlattenedStorage`, will fix in base.

Issue #372, opened 2021-09-27
Be careful with `random` in tests


liamhuber:
**Summary**

Atomistics unit tests failed, then passed again on Rerun.

**pyiron Version and Platform**

Github

**Expected Behavior**

Time-independent tests

**Actual Behavior**

Time-dependent tests

**Steps to Reproduce**

Drop out of warp and channel a tachyon pulse from the main deflector to resonate with the shield frequency.

**Further Information, Files, and Links**

[Here is the failing test](https://github.com/pyiron/pyiron_atomistics/blob/dbe0868fa6c32628b167348c1deedd0a6a29b435/tests/atomistics/structure/test_atoms.py#L1065), and it has an `np.random.random` call right in the second line. I don't know exactly how yet, but I am quite certain that sometimes this introduces a numeric instability to the tests. Jan suggested that, e.g., we might be zeroing out an entire lattice vector by chance before trying to invert the matrix. I'm sure he's on the right track.

@samwaseda blame says you touched this most recently, so maybe you will get to it before I do. It's somewhere on my list now though.

samwaseda:
ok I agree that it's a problem that it **sometimes** fails, but I might be reluctant to define a specific number for this kind of test, since it should be in principle okay with any random number.

liamhuber:
Yeah, I'm definitely OK with using random in tests, but if it leads to erroneous failures in needs to be patched.

I haven't looked at this in any detail yet. Maybe we can use random to generate some sort of orthonormal basis of allowable deformations? I.e. both make it more fragile to functionality changes by testing an entire basis worth of conditions, and more robust in that we do some manipulation/verification to make sure random isn't taking us into disallowed territory.

liamhuber:
Ok, it wasn't the matrix inversion at all -- the matrix has diagonal entries of 4.2, and is pertrubed by random numbers with magnitude <= 0.1, so there is no way to drive it to non-invertiblity.

Rather, the problem was that `get_scaled_positions` ensures that our values lie in [0, 1), and the manual thing we were comparing against didn't. So every now and then it would fail because -eps != 1-eps.

Patched in #389

PR #389, opened 2021-10-11
Fix scaled test

liamhuber:
for some cell perturbations we ran into the problem that: -eps != 1 - eps. In `get_scaled_positions` we really force them to lie in [0, 1), so now the test case does too.

Solves #372


pmrv:
@liamhuber I suppose this is good to go?

liamhuber:
> @liamhuber I suppose this is good to go?

Oh, geez, this is *my* PR? 😆 tbh I forgot about it. Yes, it's totally good to go!

PR #395, opened 2021-10-21
increase default number of empty states

samwaseda:
I realized that in many of my Fe calculations the number of empty states was not sufficient. Since having too many is far less problematic than too few, I increased them by 7.


samwaseda:
This was actually not necessary

PR #400, opened 2021-10-28
High index surfaces

skatnagallu:
A new function to get a high index surface based on the step length, step orientation and kink orientation.


pyiron_atomistics/atomistics/structure/factory.py
```diff
@@ -407,3 +407,32 @@ def from_pymatgen(pymatgen_obj):
     @wraps(ovito_to_pyiron)
     def from_ovito(ovito_obj):
         return ovito_to_pyiron(ovito_obj)
+
+    def get_high_ind_surf_sk(self, element='Ni', Xstr='fcc', lattice_constant=3.526, terraceOrientation=[1,1,1],
```
sudarsan-surendralal: A more descriptive name like `get_high_index_surface` would be better

pyiron_atomistics/atomistics/structure/factory.py
```diff
@@ -407,3 +407,32 @@ def from_pymatgen(pymatgen_obj):
...
+            please change the kinkOrientation and try again')
+
+        temp = (np.cross(np.squeeze(eqvdirK[eqvdirK_ind[0]]), np.squeeze(eqvdirS))).tolist().index(terraceOrientation)
+        fin_kinkOrientation = eqvdirK[eqvdirK_ind[0]]
+        fin_stepOrientation = eqvdirS[temp]
+        vec1 = (np.asanyarray(fin_stepOrientation).dot(lengthStep)) + (np.asanyarray(fin_kinkOrientation).dot(lengthKink))
+        vec2 = (np.asanyarray(fin_kinkOrientation).dot(lengthTerrace)) + stepDownVector
+        highIndexSurface = np.cross(np.asanyarray(vec1),np.asanyarray(vec2))
+        highIndexSurface = np.array(highIndexSurface/np.gcd.reduce(highIndexSurface), dtype=int)
+        return fin_kinkOrientation, fin_stepOrientation, highIndexSurface
```
sudarsan-surendralal: Here, you return the indices of the kink. step and the high index surface. But I guess it would be nice to go one step further and use these quantities to actually return the structure

sudarsan-surendralal (changes requested):
This looks great @skatnagallu although I think some changes need to be more (see below)

pyiron_atomistics/atomistics/structure/factory.py
liamhuber: Indeed, with tab completion we don't need to be afraid of the slightly longer name.
I'd go with `high_index_surface` to better match the pattern of the other factory methods.

liamhuber (changes requested):
Cool, very nice contribution! 👍 

In addition to Sudarsan's comments:
- Please make this PEP8 compliant (capitalization, etc). The easiest way to this is probably just download an IDE, e.g. [pycharm](https://www.jetbrains.com/pycharm/download/), and open the atomistics repository as a project. For pycharm you can then just go through and get rid of all the yellow squiggles, most of which they'll probably have a recommendation for.
- Please add a docstring

The final thing is to add a unit test. You're very welcome to do this yourself, but if it's new to you then once the other items have been dealt with I'd be happy to write the test up so you have an example pertaining to content you're super familiar with. Just ping me for re-review when you're ready 😄

pyiron_atomistics/atomistics/structure/factory.py
pmrv: We should also think about defaults for `element` and `Xstr` (maybe `crystalstructure` is a better name like for the ASE functions).  Our other methods don't have defaults for elements and I'd prefer this to be consistent.

pyiron_atomistics/atomistics/structure/factory.py
```diff
@@ -407,3 +416,68 @@ def from_pymatgen(pymatgen_obj):
...
+        vec2 = (np.asanyarray(fin_kink_orientation).dot(length_terrace)) + step_down_vector
+        high_index_surface = np.cross(np.asanyarray(vec1), np.asanyarray(vec2))
+        high_index_surface = np.array(high_index_surface / np.gcd.reduce(high_index_surface), dtype=int)
+        surf = ase_surf(basis, high_index_surface, layers, vacuum)
+        sga = SpacegroupAnalyzer(pyiron_to_pymatgen(ase_to_pyiron(surf)))
+        pmg_refined = sga.get_refined_structure()
+        slab = pymatgen_to_pyiron(pmg_refined)
+        slab.positions[:, 2] = slab.positions[:, 2] - np.min(slab.positions[:, 2])
+        slab.set_pbc = True
+        slab.set_initial_magnetic_moments(np.repeat(1.0, len(slab)))
```
sudarsan-surendralal: Is this line necessary? Why set initial magnetic moments to one?

tests/atomistics/structure/test_high_index_surface.py
```diff
@@ -0,0 +1,23 @@
+import unittest
+from pyiron_atomistics.atomistics.structure.factory import StructureFactory
+
+
+class TestHighIndexSurface(unittest.TestCase):
+    def test_high_index_surface(self):
+        slab, h, s, k = StructureFactory().get_high_index_surface(element='Ni', crystal_structure='fcc',
```
sudarsan-surendralal: The structure `slab` is unused in the tests. You could check it's length/chemical formula for completeness.

pyiron_atomistics/atomistics/structure/factory.py
```diff
@@ -407,3 +416,68 @@ def from_pymatgen(pymatgen_obj):
     @wraps(ovito_to_pyiron)
     def from_ovito(ovito_obj):
         return ovito_to_pyiron(ovito_obj)
+
+    def get_high_index_surface(self, element='Ni', crystal_structure='fcc', lattice_constant=3.526,
```
sudarsan-surendralal: I agree with @liamhuber. For consistency, the name `high_index_surface` would be a better name.

sudarsan-surendralal (commented):
This looks good now. Some minor suggestions for improvement below

pyiron_atomistics/atomistics/structure/factory.py
skatnagallu: For SPHInX jobs, if I do not initialize the magnetic moments to some initial values, it's not considering spins for calculations. That's why I have been initializing them to 1. Let me know if there is another way.

sudarsan-surendralal: > 
> 
> For SPHInX jobs, if I do not initialize the magnetic moments to some initial values, it's not considering spins for calculations. That's why I have been initializing them to 1. Let me know if there is another way.

But since this routine will be used to set up high index surfaces in general for both spin/non-spin polarized as well as non-electronic structure calculations, this line should be removed and should only be used when setting up sphinx calculations with magnetic moments.

tests/atomistics/structure/test_high_index_surface.py
```diff
@@ -0,0 +1,24 @@
...
+                                                              step_down_vector=[1, 1, 0], length_step=2,
+                                                              length_terrace=3,
+                                                              length_kink=1, layers=60,
+                                                              vacuum=10)
+        self.assertEqual(len(h), 3)
+        self.assertEqual(h[0], -9)
+        self.assertEqual(len(k), 3)
+        self.assertEqual(len(s), 3)
+        self.assertEqual(len(slab), 60)
+
```
liamhuber, line 41: Please add assertions which cause each of your value errors to raise. The syntax can look like this:

```python
with self.assertRaises(ValueError):
    StructureFactory().high_index_surface(**error_causing_kwargs)
```

pyiron_atomistics/atomistics/structure/factory.py
```diff
@@ -71,24 +73,29 @@ def aimsgb(self):
 
     def cut(self, *args, **kwargs):
         return self.ase.cut(*args, **kwargs)
+
```
liamhuber, line 80: I think elsewhere we do the doc addition immediately adjacent to the function definition, so for consistency let's just keep doing that instead of these newlines.

pyiron_atomistics/atomistics/structure/factory.py
```diff
@@ -407,3 +416,68 @@ def from_pymatgen(pymatgen_obj):
     @wraps(ovito_to_pyiron)
     def from_ovito(ovito_obj):
         return ovito_to_pyiron(ovito_obj)
+
+    def high_index_surface(self, element='Ni', crystal_structure='fcc', lattice_constant=3.526,
+                           terrace_orientation=None, step_orientation=None, kink_orientation=None,
```
liamhuber: It looks to me like `terrace/step/kink/_orientation` are actually required input. Please either make them args (instead of kwargs), or in the body set a default like `terrace_orientation = terrace_orientation if terrace_orientation is not None else [1, 0, 0]`

(but kudos for not putting mutable datatypes like [1, 0, 0] directly as default values! That is indeed a bad idea in python)

pyiron_atomistics/atomistics/structure/factory.py
```diff
@@ -407,3 +416,68 @@ def from_pymatgen(pymatgen_obj):
...
+
+        Args:
+            element (str): The parent element eq. "N", "O", "Mg" etc.
+            crystal_structure (str): The crystal structure of the lattice
+            lattice_constant (float): The lattice constant
+            terrace_orientation (list): The miller index of the terrace eg., [1,1,1]
+            step_orientation (list): The miller index of the step eg., [1,1,0]
+            kink_orientation (list): The miller index of the kink eg., [1,1,1]
+            step_down_vector (list): The direction for stepping down from the step to next terrace eg., [1,1,0]
+            length_terrace (int): The length of the terrace along the kink direction in atoms eg., 3
```
liamhuber, line 443: Instead of 'eg' please more specifically inform the reader that these are the defaults.

pyiron_atomistics/atomistics/structure/factory.py
```diff
@@ -407,3 +416,68 @@ def from_pymatgen(pymatgen_obj):
...
+        high_index_surface = np.cross(np.asanyarray(vec1), np.asanyarray(vec2))
+        high_index_surface = np.array(high_index_surface / np.gcd.reduce(high_index_surface), dtype=int)
+        surf = ase_surf(basis, high_index_surface, layers, vacuum)
+        sga = SpacegroupAnalyzer(pyiron_to_pymatgen(ase_to_pyiron(surf)))
+        pmg_refined = sga.get_refined_structure()
+        slab = pymatgen_to_pyiron(pmg_refined)
+        slab.positions[:, 2] = slab.positions[:, 2] - np.min(slab.positions[:, 2])
+        slab.set_pbc = True
+        slab.set_initial_magnetic_moments(np.repeat(1.0, len(slab)))
+        return slab, high_index_surface, fin_kink_orientation, fin_step_orientation
```
liamhuber: I definitely see the merit to giving users access to the surface and orientation data, but from a design perspective I'm a little uncomfortable packing it all together in a big tuple -- the standard behaviour for `StructureFactory` methods is to return just a structure.

What about doing something roughly analogous to how we handle `aimsgb` and refactoring this into two methods: `high_index_sufrace_info` and `high_index_surface`, where the latter returns *only* a structure and calls the former internally, and the former gives the remaining elements of this tuple?

It's slightly less efficient since the vectors will get calculated twice if someone really wants both the structure *and* the vectors, but the calculation doesn't look terribly expensive and I would be willing to pay that price for consistency in the return type.

Alternatively, you could add a flag like `return_vectors=False` which makes this output optional. (Personally I prefer multiple functions to extremely variable return cases, but even numpy does this so I wouldn't let me preferences stand in the way)

liamhuber (commented):
Very nice! I have a couple of nits still, and a test request, but IMO the only "real" outstanding issue is just a design question on how we should handle the return value(s) 😄

liamhuber (commented):
Good stuff. I would like to wait to merge until the tests are passing -- I think the current failures relate to issues with hdf5io so this should only require some patience and some merging master into this branch.

pyiron_atomistics/atomistics/structure/factory.py
```diff
@@ -407,3 +416,105 @@ def from_pymatgen(pymatgen_obj):
...
+            vacuum (float): Thickness of vacuum on the top of the slab. default:10
+
+        Returns:
+            slab: pyiron_atomistics.atomistics.structure.atoms.Atoms instance Required surface
+        """
+        basis = self.crystal(element=element, bravais_basis=crystal_structure, lattice_constant=lattice_constant)
+        high_index_surface, \
+            fin_kink_orientation, \
+            fin_step_orientation = self.high_index_surface_info(element='Ni',
+                                                                crystal_structure='fcc', lattice_constant=3.526,
```
niklassiemer: Here, you need to pass the given arguments of the function call! 
```suggestion
            fin_step_orientation = self.high_index_surface_info(element=element,
                                                                crystal_structure=crystal_structure, lattice_constant=lattice_constant,
```
Indeed, I would prefer no defaults for these in the first place.

pyiron_atomistics/atomistics/structure/factory.py
```diff
@@ -407,3 +416,106 @@ def from_pymatgen(pymatgen_obj):
...
+        fin_step_orientation = eqvdirs[temp]
+        vec1 = (np.asanyarray(fin_step_orientation).dot(length_step)) + \
+               (np.asanyarray(fin_kink_orientation).dot(length_kink))
+        vec2 = (np.asanyarray(fin_kink_orientation).dot(length_terrace)) + step_down_vector
+        high_index_surface = np.cross(np.asanyarray(vec1), np.asanyarray(vec2))
+        high_index_surface = np.array(high_index_surface / np.gcd.reduce(high_index_surface), dtype=int)
+
+        return high_index_surface, fin_kink_orientation, fin_step_orientation
+
+    def high_index_surface(self, element='Ni', crystal_structure='fcc', lattice_constant=3.526,
```
sudarsan-surendralal: To be consistent with other methods in this class, I think there shouldn't be default values for `element`, `crystal_structure`, and `lattice_constant` but rather require that the user specify these values.

pyiron_atomistics/atomistics/structure/factory.py
```diff
@@ -407,3 +416,106 @@ def from_pymatgen(pymatgen_obj):
...
+            fin_step_orientation = self.high_index_surface_info(element=element,
+                                                                crystal_structure=crystal_structure,
+                                                                lattice_constant=lattice_constant,
+                                                                terrace_orientation=terrace_orientation,
+                                                                step_orientation=step_orientation,
+                                                                kink_orientation=kink_orientation,
+                                                                step_down_vector=step_down_vector,
+                                                                length_step=length_step,
+                                                                length_terrace=length_terrace,
+                                                                length_kink=length_kink)
```
sudarsan-surendralal: Avoiding unused variables

```suggestion
        high_index_surface, _, _ = self.high_index_surface_info(element=element,
                                                                crystal_structure=crystal_structure,
                                                                lattice_constant=lattice_constant,
                                                                terrace_orientation=terrace_orientation,
                                                                step_orientation=step_orientation,
                                                                kink_orientation=kink_orientation,
                                                                step_down_vector=step_down_vector,
                                                                length_step=length_step,
                                                                length_terrace=length_terrace,
                                                                length_kink=length_kink)
```

sudarsan-surendralal (approved):
Apart from a couple of minor changes, LGTM!

PR #412, opened 2021-11-08
Strain

samwaseda:
I added the Green-Lagrangian strain for `apply_strain`. This one is probably more compatible with most of the strain calculation algorithms in atomistic simulations, but I still don't want to make it the default behaviour because it doesn't allow a non-symmetric strain.

### DIFF
--- a/pyiron_atomistics/atomistics/structure/atoms.py
+++ b/pyiron_atomistics/atomistics/structure/atoms.py
@@ -2275,29 +2275,45 @@ class Atoms(ASEAtoms):
                 else:
                     self.selective_dynamics[atom_ind] = [True, True, True]
 
-    def apply_strain(self, epsilon, return_box=False):
+    def apply_strain(self, epsilon, return_box=False, mode='linear'):
         """
-        Apply a given strain on the structure
+        Apply a given strain on the structure. It applies the matrix `F` in the manner:
+
+        ```
+            new_cell = F @ current_cell
+        ```
 
         Args:
-            epsilon (float/list/ndarray): epsilon matrix. If a single number is set, the same strain
-                                          is applied in each direction. If a 3-dim vector is set, it
-                                          will be multiplied by a unit matrix.
+            epsilon (float/list/ndarray): epsilon matrix. If a single number is set, the same
+                strain is applied in each direction. If a 3-dim vector is set, it will be
+                multiplied by a unit matrix.
             return_box (bool): whether to return a box. If set to True, only the returned box will
-                               have the desired strain and the original box will stay unchanged.
+                have the desired strain and the original box will stay unchanged.
+            mode (str): `linear` or `lagrangian`. If `linear`, `F` is equal to the epsilon value.
+                If `lagrangian`, epsilon is given by `(F^T * F - 1) / 2`. It raises an error if
+                the strain is not symmetric (if the shear components are given).
         """
+        if mode not in ['lagrangian', 'linear']:
+            raise ValueError('mode must be `linear` or `lagrangian`')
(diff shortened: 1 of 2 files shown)


pyiron_atomistics/atomistics/structure/atoms.py
```diff
@@ -2275,29 +2275,45 @@ def set_constraint(self, constraint=None):
...
         if return_box:
             structure_copy = self.copy()
         else:
             structure_copy = self
         cell = structure_copy.cell.copy()
-        cell = np.matmul(epsilon + np.eye(3), cell)
+        if mode == 'linear':
+            F = epsilon + np.eye(3)
+        else:
+            E, V = np.linalg.eigh(2*epsilon+np.eye(3))
```
sudarsan-surendralal, line 2314: Minor nit: I know that `np.einsum` is an efficient way to do tensor/matrix operations but it is hard to understand what you end up with on first glance. Perhaps adding a comment explaining whats going on would be nice.

tests/atomistics/structure/test_atoms.py
```diff
@@ -1086,6 +1086,15 @@ def test_apply_strain(self):
         basis_Fe = CrystalStructure("Fe", bravais_basis="bcc", lattice_constants=2.85)
         basis_Fe.apply_strain(0.01*np.eye(3))
         self.assertAlmostEqual(basis_Fe.cell[0,0], 2.85*1.01)
+        self.assertRaises(ValueError, basis_Fe.apply_strain, epsilon=np.eye(3), mode='hello')
+        self.assertRaises(
+            ValueError,
+            basis_Fe.apply_strain, epsilon=np.random.random((3,3)), mode='lagrangian'
+        )
+        basis_Fe = CrystalStructure("Fe", bravais_basis="bcc", lattice_constants=2.85)
+        self.assertTrue(np.allclose(basis_Fe.apply_strain(
```
sudarsan-surendralal, line 1095: For completeness, can you add a positive test for a vector `epsilon` input?

pyiron_atomistics/atomistics/structure/atoms.py
samwaseda: The problem for me is that while many people told me that they have trouble with `einsum`, no one made me understand what they find difficult ... Is it the Einstein notation itself? Or do you prefer simply a completely different notation like this (to be honest this one below is super difficult to read for me)?

```python
F = np.sum(V[:, np.newaxis, :] * np.sqrt(E)[np.newaxis, np.newaxis, :] * V.T[np.newaxis, :, :], axis=1)
```

By the way, mostly the reason I use `einsum` is because it's easy to understand which axes are multiplies with which axes & which axes are summed, and less because of its efficiency...

sudarsan-surendralal: > 
> 
> The problem for me is that while many people told me that they have trouble with `einsum`, no one made me understand what they find difficult ... Is it the Einstein notation itself? Or do you prefer simply a completely different notation like this (to be honest this one below is super difficult to read for me)?
> 
> ```python
> F = np.sum(V[:, np.newaxis, :] * np.sqrt(E)[np.newaxis, np.newaxis, :] * V.T[np.newaxis, :, :], axis=1)
> ```

From this line, I can now tell that you reshape the eigenvalue and eigenvectors to perform two element-wise multiplication operations and then sum over one axis. For me personally, the logic is more apparent in this example because I am not too familiar with the `einsum` notation (which is my fault). I agree with you that using `einsum` is cleaner and more `numpy` like. I just thought having a comment explaining what you do would help people like me.

Issue #409, opened 2021-11-07
h5io version 0.1.4 causing several problems


Leimeroth:
**Summary**
 Updating from h5io version 0.1.2 to version 0.1.4 breaks all kinds of functionality.
I experienced that lammps jobs don't run and get structure does not work. This seems to be related to a change in the returned datatype, f.e. the positions array of a structure had dtype=object instead of float and could not be cast, causing an error. 
Below is the hdf part of the backtrace of a lammps job. 

EDIT:
Also added get_structure error trace.

**Steps to Reproduce**

Update h5io to version 0.1.4 and try to run jobs or load structures from existing ones.

```
**Further Information, Files, and Links**
~/git_projects/pyiron_atomistics/pyiron_atomistics/atomistics/structure/atoms.py in to_hdf(self, hdf, group_name)
    426                         el.to_hdf(hdf_species)
    427             hdf_structure["species"] = [el.Abbreviation for el in self.species]
--> 428             hdf_structure["indices"] = self.indices
    429 
    430             with hdf_structure.open("tags") as hdf_tags:

~/git_projects/pyiron_base/pyiron_base/generic/hdfio.py in __setitem__(self, key, value)
    172         elif isinstance(value, tuple):
    173             value = list(value)
--> 174         h5io.write_hdf5(
    175             self.file_name,
    176             value,

~/miniconda3/lib/python3.9/site-packages/h5io/_h5io.py in write_hdf5(fname, data, overwrite, compression, title, slash, use_json)
    109             del fid[title]
    110         cleanup_data = []
--> 111         _triage_write(title, data, fid, comp_kw, str(type(data)),
    112                       cleanup_data, slash=slash, title=title,
    113                       use_json=use_json)

~/miniconda3/lib/python3.9/site-packages/h5io/_h5io.py in _triage_write(key, value, root, comp_kw, where, cleanup_data, slash, title, use_json)
    185     elif isinstance(value, np.ndarray):
    186         if not (value.dtype == np.dtype('object') and
--> 187                 len(set([sub.dtype for sub in value])) == 1):
    188             _create_titled_dataset(root, key, 'ndarray', value)
    189         else:

~/miniconda3/lib/python3.9/site-packages/h5io/_h5io.py in <listcomp>(.0)
    185     elif isinstance(value, np.ndarray):
    186         if not (value.dtype == np.dtype('object') and
--> 187                 len(set([sub.dtype for sub in value])) == 1):
[...]

PR #414, opened 2021-11-09
Pyscal solid liquid

Leimeroth:
Adds pyscals ability to find solid and liquid atoms in a structure. 
Also adds a get_system convenience function to have easier access to pyscal functionality that is not directly accessible and replaces repetitive code with the get_system function.

### DIFF
--- a/pyiron_atomistics/atomistics/structure/analyse.py
+++ b/pyiron_atomistics/atomistics/structure/analyse.py
@@ -531,6 +531,10 @@ class Analyse:
     def pyscal_voronoi_volume(self):
         """    Calculate the Voronoi volume of atoms        """
         return analyse_voronoi_volume(atoms=self._structure)
+    
+    def pyscal_find_solids(self, neighbor_method="cutoff", cutoff=0, bonds=6, threshold=0.5, avgthreshold=0.6, cluster=False):
+        return analyse_find_solids()
+
 
     def get_voronoi_vertices(self, epsilon=2.5e-4, distance_threshold=0, width_buffer=10):
         """
--- a/pyiron_atomistics/atomistics/structure/analyse.py
+++ b/pyiron_atomistics/atomistics/structure/analyse.py
@@ -8,7 +8,7 @@ from sklearn.cluster import AgglomerativeClustering, DBSCAN
 from scipy.sparse import coo_matrix
 from scipy.spatial import Voronoi
 from pyiron_atomistics.atomistics.structure.pyscal import get_steinhardt_parameter_structure, analyse_cna_adaptive, \
-    analyse_centro_symmetry, analyse_diamond_structure, analyse_voronoi_volume
+    analyse_centro_symmetry, analyse_diamond_structure, analyse_voronoi_volume, get_system, analyse_find_solids
 from pyiron_atomistics.atomistics.structure.strain import Strain
 from pyiron_base.generic.util import Deprecator
 from scipy.spatial import ConvexHull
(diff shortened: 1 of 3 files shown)


pyiron_atomistics/atomistics/structure/pyscal.py
```diff
@@ -234,13 +223,63 @@ def analyse_voronoi_volume(atoms):
...
     """
-    s.publication_add(publication())
-    sys = pc.System()
-    sys.read_inputfile(atoms, format="ase")
+    sys = get_system(atoms)
     sys.find_neighbors(method="voronoi")
     atoms = sys.atoms
     return np.array([atom.volume for atom in atoms])
 
+def get_system(atoms):
```
pmrv: I think for consistency we would want to call that `pyiron_to_pyscal`, similar to `pyiron_to_ase` and `pyiron_to_pymatgen`.

Leimeroth: renamed it. I also moved it next to the to_ase, to_pymatgen and to_ovito methods

pyiron_atomistics/atomistics/structure/pyscal.py
```diff
@@ -234,13 +223,63 @@ def analyse_voronoi_volume(atoms):
...
+    """
+    s.publication_add(publication())
+    sys = pc.System()
+    sys.read_inputfile(
+        pyiron_atomistics.atomistics.structure.atoms.pyiron_to_ase(atoms),
+        format="ase",
+        )
+    return sys
+
+def analyse_find_solids(atoms, neighbor_method="cutoff",
```
pmrv, line 250: This returns the number of *atoms* that are inside a solid phase?  If so it'd cool to also return an index array, that shows which are solid and which are liquid, maybe behind a switch, like the `pyscal_adaptive_cna` method.

pyiron_atomistics/atomistics/structure/analyse.py
```diff
@@ -532,6 +532,40 @@ def pyscal_voronoi_volume(self):
...
         return analyse_voronoi_volume(atoms=self._structure)
 
+    def pyscal_find_solids(self, neighbor_method="cutoff",
+        cutoff=0, bonds=0.5,
+        threshold=0.5, avgthreshold=0.6,
+        cluster=False, q=6, right=True,
+        return_sys=False,
+        ):
+        """
+        Get the number of solids or the corresponding pyscal system.
```
pmrv, line 542: Typo?

pyiron_atomistics/atomistics/structure/pyscal.py
Leimeroth: This only returns the number by default, but when using return_system=True it returns the complete pyscal system object, which allows to get the indices. I thought this is the most simple in the default case and the most flexible approach when using the return_system option.
Alternatively I could change the default to return an array with the indices of all atoms that are identified as solid, so that the len(returned_array) gives the number of solids I guess. But I do not know if the order of the atoms stays the same when converting between pyiron ase and pyscal system, so I am not sure if indices will always be correct

pyiron_atomistics/atomistics/structure/analyse.py
Leimeroth: what do you mean?

pyiron_atomistics/atomistics/structure/pyscal.py
pmrv: Ah, I had simply missed that while skimming.  Since anything might happen while changing from pyiron to pyscal returning the full system is ok, I suppose.

pyiron_atomistics/atomistics/structure/analyse.py
pmrv: Since I misunderstood the method, I thought `or` should be `of` or some such...  I promise to read more carefully next time. ;)

Leimeroth:
I think the tests are failing due to #409, since the error messages seem unrelated to my changes. I will wait till this is sorted out and than merge this when tests run, if you don't see any other problems @pmrv.

pmrv:
Yep, once the h5io is sorted out, go ahead and merge.

pmrv (approved):

niklassiemer:
I fixed the h5io version (for now) in #410, thus, try to rerun the tests :)

liamhuber:
Cool! We are really getting a lot of pyscal mapped functions -- what about doing a refactor (after this PR, not part of it) so that we nest these bindings, i.e. `structure.analyse.pyscal.find_solids` instead of `structure.analyse.pyscal_find_solids`?

Leimeroth:
> Cool! We are really getting a lot of pyscal mapped functions -- what about doing a refactor (after this PR, not part of it) so that we nest these bindings, i.e. `structure.analyse.pyscal.find_solids` instead of `structure.analyse.pyscal_find_solids`?

I always tried to type exactly that.

PR #416, opened 2021-11-09
Make Methfessel Paxton default in SPHInX

samwaseda:
Since Methfessel-Paxton is now available in SPHInX I made it the default behaviour in pyiron.

### DIFF
--- a/pyiron_atomistics/sphinx/base.py
+++ b/pyiron_atomistics/sphinx/base.py
@@ -373,6 +373,7 @@ class SphinxBase(GenericDFTJob):
         self.input.KpointCoords = [0.5, 0.5, 0.5]
         self.input.KpointFolding = [4, 4, 4]
         self.input.EmptyStates = "auto"
+        self.input.MethfesselPaxton = 1
         self.input.Sigma = 0.2
         self.input.Xcorr = "PBE"
         self.input.VaspPot = False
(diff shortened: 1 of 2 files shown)


sudarsan-surendralal:
There is a danger to setting this as the default. The Methfessel-Paxton scheme is not recommended for semi-conductors and insulators where you could end up with occupancies less than 0 (see the [discussion](https://www.vasp.at/wiki/index.php/ISMEAR) on the VASP wiki). There is no way of knowing if a structure is metallic or not beforehand and therefore wouldn't something like the Fermi/Gaussian type smearing be better?

samwaseda:
But shouldn't we follow the default behaviour in VASP? (cf. [this page](https://www.vasp.at/wiki/index.php/ISMEAR))

sudarsan-surendralal:
> 
> 
> But shouldn't we follow the default behaviour in VASP? (cf. [this page](https://www.vasp.at/wiki/index.php/ISMEAR))

Generally yes, but in this case, we need to consider the possibility that the VASP default may not be the right option. But I'm not sure.

samwaseda:
ok but I guess it's a bigger problem that SPHInX and VASP have different default settings, isn't it?

sudarsan-surendralal:
> 
> 
> ok but I guess it's a bigger problem that SPHInX and VASP have different default settings, isn't it?

Ah yes, then what you implemented here is correct (both VASP and SPHINX now have the same default parameters). But I guess people working with semiconductors/insulators should be aware of this.

sudarsan-surendralal (approved):

PR #419, opened 2021-11-10
Cluster (atom) positions

samwaseda:
Cluster points which lie close to each other, also considering PBC.

Super simple example:

```python
positions = atoms.analyse.cluster_points(eps=2)
```

This example should return the atom positions, if no two atoms lie within a distance of 2. If there are at least two atoms which lie within a distance of 2, their entries will be replaced by their mean position.


samwaseda:
Sorry there are a lot of one line changes because I was correcting PEP8, but hopefully you'll notice that what matters is only the function that comes towards the end of the file.

liamhuber (commented):
Nice! Docstring should be explicit that the clustering technique is DBScan, and please expose the remaining dbscan Meath (minsamples i think was)

pmrv (approved):

PR #423, opened 2021-11-12
Parse irreductible kpoints properly

sudarsan-surendralal:


pyiron_atomistics/vasp/base.py
```diff
@@ -2183,7 +2183,7 @@ def to_hdf(self, hdf):
                     hdf=hdf5_output, group_name="electronic_structure"
                 )
 
-            if self.outcar.parse_dict:
+            if len(self.outcar.parse_dict.keys()):
```
samwaseda: `if len(self.outcar.parse_dict.keys())>0`?

PR #424, opened 2021-11-15
Use state instead of Settings

liamhuber:
A surprising amount of the time that just meant getting rid of the Settings import, as it was unused. I also snuck in a couple other unused import removals and some encoding info additions here and there, depending on the repo.


niklassiemer:
You are a bit early, since the `state` PR is not yet released to conda 😃

liamhuber:
> You are a bit early, since the `state` PR is not yet released to conda 😃

😆 yep, we're looking at it simultaneously

samwaseda (approved):
I'm not particularly happy with the fact that there's no test, but other than that it looks good to me. :)

liamhuber:
> I'm not particularly happy with the fact that there's no test, but other than that it looks good to me. :)

We'll, the basic state stuff is tested over in base. It's true though that I didn't add any special tests, and Marvin did indeed track down a state related (but maybe not state or caused?) issue. I don't know how to conveniently link issues on my phone in the app though, so you'll just have to go look for it 🤣

PR #425, opened 2021-11-15
Some test modernization

liamhuber:
A bunch of tests weren't passing on my macbook, so I'm fixing that and simultaneously updating them to use more modern syntax (mostly the custom test case classes from base and `state`).


liamhuber:
Ok, so I actually did not put much effort into modernizing the sphinx and vasp tests. I think they could probably benefit greatly from `pyiron_base._tests.TestWithProject`, but I will be a little lazy here and leave that to the sphinx/vasp users (you know who you are and I'll probably ping you for review once this isn't a draft 😉)

liamhuber:
@samwaseda you were saying earlier that you wanted to spend more time reviewing unit tests, right? 😜

liamhuber:
I'm also a little sketched that the diff is so big. A lot of that looks like it should already be in master from #424 and other recent PRs...Maybe there is some refresh on github's side that I need to figure out?

liamhuber:
@samwaseda bump

tests/atomistics/job/test_atomistic.py
```diff
@@ -77,19 +74,16 @@ def test_get_structure(self):
...
+        os.unlink("export.csv")
+        os.unlink("get_structure_test.tar.gz")
 
         for i, struct in enumerate(job.iter_structures()):
             # breakpoint()
             self.assertTrue(np.allclose(job.output.positions[i], struct.positions))
             self.assertTrue(np.allclose(job.output.cells[i], struct.cell.array))
             self.assertTrue(np.allclose(job.output.indices[i], struct.indices))
-
-if __name__ == "__main__":
```
samwaseda, line 94: Since the tests don't work in the terminal I'd like to keep these ones

samwaseda (approved):
Well to be honest I don't understand much of the changes but I trust you :D

liamhuber:
> Well to be honest I don't understand much of the changes but I trust you :D

😝 well I don't trust me, but the tests are passing and that gives me some confidence, haha

tests/atomistics/job/test_atomistic.py
liamhuber: Ah, sorry, I didn't see this in the mobile app.

Anyhow, it really necessary? I guess if you do `python tests/atomistics/job/test_atomistic.py` it is, but if you do `python -m unittest discover tests/atomistics/job` it will run this and the other ones there... but maybe you can append `/test_atomistic.py` to get *just* these? Tbh I haven't played around with it much from command line, but I'm 99% sure you can swing it this way

niklassiemer: Yes, with `python -m unittest path/to/test_file.py` you run all tests in this file (If you point to one specific one, you do not need to discover). So Sam, just make an `alias run_test_py="python -m unittest discover "` and you may fire any directory you want by `run_test_py test_dir` :)

PR #439, opened 2021-11-25
replace full output info in __str__ by chemical formula

samwaseda:
In order to exploit the feature [here](https://github.com/pyiron/pyiron_base/pull/561/files), I would like to suggest to largely simplify `__str__` in `Atoms` by replace the full info by the chemical formula. this allows us to set something like:

```python
job = pr.create.job.Lammps(('lammps', structure))
```

### DIFF
--- a/pyiron_atomistics/atomistics/structure/atoms.py
+++ b/pyiron_atomistics/atomistics/structure/atoms.py
@@ -2031,9 +2031,6 @@ class Atoms(ASEAtoms):
     #     return len(self.indices)
 
     def __repr__(self):
-        return self.__str__()
-
-    def __str__(self):
         if len(self) == 0:
             return "[]"
         out_str = ""
@@ -2051,6 +2048,9 @@ class Atoms(ASEAtoms):
             out_str += "cell: \n"
             out_str += str(self.cell) + "\n"
         return out_str
+
+    def __str__(self):
+        return self.get_chemical_formula()
 
     def __setitem__(self, key, value):
         if isinstance(key, (int, np.integer)):
--- a/tests/atomistics/structure/test_atoms.py
+++ b/tests/atomistics/structure/test_atoms.py
@@ -1558,6 +1558,14 @@ class TestAtoms(unittest.TestCase):
         structure = ase_to_pyiron(molecule('H2COH'))
         structure.set_dihedral(4, 0, 1, 2, angle=90)
 
+    def test_str_repr(self):
+        H2 = Atoms(positions=[3*[0], 3*[0.5]], cell=np.eye(3), elements=2*['H'], pbc=True)
+        self.assertEqual(
+            repr(H2),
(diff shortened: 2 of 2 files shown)


niklassiemer (approved):
I like this solution. In my notebook testing nothing changes on just providing a structure, since the notebook ignores `__str__` and uses `__repr__`. Just if you `print(structure)` you just get the chemical formula.

pmrv:
I see how this is useful, but I also think it's nice to have just the text dump of a structure, especially when checking small structures.

samwaseda:
> I see how this is useful, but I also think it's nice to have just the text dump of a structure, especially when checking small structures.

Is there a case that `__str__` is used and not `__repr__`?

niklassiemer:
On `print` or `str() ` 😅

liamhuber:
Maybe we replace `__str__` with the formula as suggested, but add a `to_str()` method (or similarly named) that wraps the string rep in `__repr__`

samwaseda:
But I still don't understand what would make you use `print` or `str`. If that's really the case, I would rather support @liamhuber 's idea of adding `to_str`, but frankly I'd personally not know when I'd ever use it.

niklassiemer:
I am not sure if we would need this. With this change we would get after all:
```python
Fe = pr.create.structure.bulk('Fe')
Fe
>>> Fe: [0. 0. 0.]
>>> pbc: [ True  True  True]
>>> cell: 
>>> Cell([[-1.435, 1.435, 1.435], [1.435, -1.435, 1.435], [1.435, 1.435, -1.435]])
# same as `print(repr(Fe))`
```
only 
```python
print(Fe)
>>> Fe
str(Fe)
>>> 'Fe'
```
would change. The only reason for a `to_str()` method would be if we would need the long string output for something else than to display for human readability. I.e. `structure_str = stucture.to_str()` and I just do not see any case where I would need to get a string `Fe: [0. 0. 0.]\npbc: [...]`.

liamhuber:
Yes, it's exactly as you say, Niklas. I am also not immediately sure what the application might be for needing the full string, but I share Marvin's reticence for getting rid of it entirely. It clutters the namespace a little bit, but I'd be more comfortable if we kept the string accessible by a special method. Otherwise installs like the change in the default behaviour

niklassiemer:
I am ok with adding that method, however, if I really needed that string I could just do `structure_as_string = repr(Fe)`?

samwaseda:
If we don't know whether `to_str` could be useful or not, then I'd rather not add it in the first place, because `Atoms` has in my opinion largely surpassed the number of options that can be seen in the auto complete. So whoever for some reason happens to need the string version of `__repr__` will most probably simply take `repr(structure)` (as @niklassiemer suggests) and not look for the extra function that may or may not exist and may or may not do what the user wants it to do.

samwaseda:
Following today's pyiron meeting, I'll merge this one as soon as the tests pass.

PR #444, opened 2021-11-26
Store shells in Neighbors Trajectory

pmrv:
I've benchmarked this and it's only marginally slower to also compute the shells additionally (maybe ~5%), so I'm always adding this because I want it for my calculations.  If anyone thinks it's too much, I will make it configurable whether shells are included or not.

### DIFF
--- a/pyiron_atomistics/atomistics/structure/neighbors.py
+++ b/pyiron_atomistics/atomistics/structure/neighbors.py
@@ -1087,7 +1087,7 @@ class NeighborsTrajectory(DataContainer):
         For trajectories with non constant amount of particles this array may contain -1 for invalid values, i.e. 
 
         Returns:
-            numpy.ndarray: An int array of dimension N_steps / stride x N_atoms x N_neighbors x 3
+            ndarray: An int array of dimension N_steps / stride x N_atoms x N_neighbors x 3
         """
         return self._flat_store.get_array_filled("shells")
 
--- a/pyiron_atomistics/atomistics/structure/neighbors.py
+++ b/pyiron_atomistics/atomistics/structure/neighbors.py
@@ -1036,6 +1036,7 @@ class NeighborsTrajectory(DataContainer):
         self._flat_store.add_array("indices", dtype=np.int64, shape=(num_neighbors,), per="element")
         self._flat_store.add_array("distances", dtype=np.float64, shape=(num_neighbors,), per="element")
         self._flat_store.add_array("vecs", dtype=np.float64, shape=(num_neighbors, 3), per="element")
+        self._flat_store.add_array("shells", dtype=np.int64, shape=(num_neighbors,), per="element")
         self._neighbor_indices = None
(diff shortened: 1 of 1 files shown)


samwaseda:
`shells` works only for super perfect structures so I'm not really sure if it ever really plays a role in trajectories

pmrv:
I'm using `NeighborsTrajectory` over in [pyiron_contrib](https://github.com/pyiron/pyiron_contrib/pull/274) also for the `TrainingContainer` so it's a "trajectory" only in the broadest sense.  As I said, though I can put it behind an option.

pyiron_atomistics/atomistics/structure/neighbors.py
```diff
@@ -1078,6 +1079,18 @@ def vecs(self):
...
 
+    @property
+    def shells(self):
+        """
+        Neighbor shell indices (excluding itself) of each atom computed using the get_neighbors_traj() method.
+
+        For trajectories with non constant amount of particles this array may contain -1 for invalid values, i.e. 
+
+        Returns:
+            numpy.ndarray: An int array of dimension N_steps / stride x N_atoms x N_neighbors x 3
```
sudarsan-surendralal:
```suggestion
            ndarray: An int array of dimension N_steps / stride x N_atoms x N_neighbors x 3
```


`ndarray` rather than `numpy.ndarray`

sudarsan-surendralal (approved):
Looks good to me!

samwaseda:
I have a totally different suggestion: How about storing only the vectors and indices, and make distances and shells functions? I am badly concerned by the amount of space this class could accidentally take by mistake (since we're talking about `n_snapshots x n_neighbors x n_atoms x n_dim` and therefore I'm not super eager to extend the space even more.

pmrv:
> I have a totally different suggestion: How about storing only the vectors and indices, and make distances and shells functions? I am badly concerned by the amount of space this class could accidentally take by mistake (since we're talking about `n_snapshots x n_neighbors x n_atoms x n_dim` and therefore I'm not super eager to extend the space even more.

I see the issue, but at least for my application we're only talking about MB, so I always want to store this information.  How about I put in a keyword argument to enable saving the "derived" data and only store vecs + indices as you proposed if it's not given?  How easy is it to instantiate `Neighbors` to recalculate the data without the starting from the structure?

samwaseda:
> How easy is it to instantiate `Neighbors` to recalculate the data without the starting from the structure?

I guess this is the **real** solution that we should aim for, because I have the feeling that it's not `Neighbors` that's slow, but loading the structures. In situations where the data is so large that calling `Neighbors` is too costly, it's certainly also imperative to store only the minimum amount of data, meaning most likely only the vectors and the indices.

Realizing all this is unfortunately not quite straightforward in the current format of `get_neighbors`, which probably I should take on. Since also the user interface will stay the same one way or other, you can also merge this PR as it is right now.

samwaseda (approved):

PR #446, opened 2021-11-27
Fix Lammps h5md parser

jan-janssen:


jan-janssen:
The HDF5 file no longer contains the indices: 
```
h5md
    version: [1 0]
h5md/author
    name: b'N/A'
h5md/creator
    name: b'lammps'
    version: b'29 Sep 2021'
observables
parameters
particles
particles/all
particles/all/box
    boundary: [b'periodic' b'periodic' b'periodic']
    dimension: 3
particles/all/box/edges
particles/all/box/edges/step
particles/all/box/edges/time
particles/all/box/edges/value
particles/all/force
particles/all/force/step
particles/all/force/time
particles/all/force/value
particles/all/position
particles/all/position/value
```

jan-janssen:
An example expository is available at https://github.com/jan-janssen/pyiron-lammps-h5md

pmrv:
Not having the indices in the HDF5 will break `get_structure` or not?

jan-janssen:
> Not having the indices in the HDF5 will break `get_structure` or not?

`get_structure` works just fine - it is using the indices from the input structure. I tested it with the example notebook above.

PR #447, opened 2021-11-28
Some docstrings for phono

pmrv:
I've also noticed that the units are a bit wonky.  `PhonopyJob.dos_energies` actually gives frequencies and `PhonopyJob.dos_total` is not at all specified.


jan-janssen (approved):
Looks good to me

pyiron_atomistics/atomistics/master/phonopy.py
```diff
@@ -406,27 +402,32 @@ def dynamical_matrix_at_q(self, q):
...
             import pylab as plt
         except ImportError:
             import matplotlib.pyplot as plt
-        if ax is None:
-            fig, ax = plt.subplots(1, 1)
-        ax.plot(self["output/dos_energies"], self["output/dos_total"], *args, **qwargs)
-        ax.set_xlabel("Frequency [THz]")
-        ax.set_ylabel("DOS")
-        ax.set_title("Phonon DOS vs Energy")
+        if ax is not None and axis is None:
```
liamhuber, line 434: I think you can one line this like _, ax = ply.subplots() if axis is None else None, axis. If you hate it don't bother but else a 4:1 line ratio is not too shabby

liamhuber (approved):
Super. One optional nit

pyiron_atomistics/atomistics/master/phonopy.py
pmrv: I guess this wouldn't work when `ax` is defined but not `axis`.  I don't like to layer inline if/else to deeply, so I'd rather forgo it.

liamhuber: Ah, you're right. Can still one-line the two clauses independently if you want

```python
axis = ax if ax is not None and axis is None else axis
_, axis = plt.subplots() if axis is None else _, axis
```

The first clause needs to be expanded if you want to explicitly throw an error when both the old and new kwargs are used though.

PR #448, opened 2021-11-28
Write POSCAR in direct coordinate when selective dynamics is on

usaikia:
When I run a vasp job with POSCAR written in cartesian coordinates vasp returns the following warning message:

"using selective dynamics as specified on POSCAR
WARNING: If single coordinates had been selected the selection of coordinates
is made according to the corresponding d i r e c t coordinates!
Don't support selection of single cartesian coordinates -- sorry ..."

Hence, I would like to write POSCAR in direct coordinates when selective dynamics is on.

### DIFF
--- a/pyiron_atomistics/vasp/structure.py
+++ b/pyiron_atomistics/vasp/structure.py
@@ -106,6 +106,7 @@ def write_poscar(structure, filename="POSCAR", write_species=True, cartesian=Tru
         f.write(endline)
         if "selective_dynamics" in structure.get_tags():
             selec_dyn = True
+	    cartesian = False
             f.write("Selective dynamics" + endline)
         sorted_coords = list()
         selec_dyn_lst = list()
--- a/pyiron_atomistics/vasp/structure.py
+++ b/pyiron_atomistics/vasp/structure.py
@@ -106,7 +106,7 @@ def write_poscar(structure, filename="POSCAR", write_species=True, cartesian=Tru
         f.write(endline)
         if "selective_dynamics" in structure.get_tags():
             selec_dyn = True
-	    cartesian = False
+            cartesian = False
             f.write("Selective dynamics" + endline)
         sorted_coords = list()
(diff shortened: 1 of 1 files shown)


liamhuber:
Indentation of the added line is wrong. Try aligning it with the others in the if clause and then it should be fine. Otherwise lgtm, but I'm not a vasp user so wait for Marvin too 😉

usaikia:
@liamhuber Thanks for pointing out the wrong indentation. I will correct it.

pmrv:
Have you checked that this actually causes an issue?  From reading the warning again and the [wiki](https://www.vasp.at/wiki/index.php/POSCAR) I currently understand that VASP only supports fixing the fractional/direct coordinates of an atom, i.e. even when writing as Cartesian it will internally convert to fractional/direct and then fix this position.  The distinction therefore only plays a role when the volume changes.

Just as a heads-up, I'm still happy to put this in to silence the warning.

usaikia:
@pmrv I did some test calculations from my side. For my calculations with cartesian poscar, apart from giving the warning message, vasp did the fixing correctly. I also checked the vasp wiki page, but not sure this warning message is not relevant anymore for these newer version of vasp or there is something else (maybe some special case where this is relevant). Hence, I thought it's a good idea to avoid this warning message by just writing the poscar in direct coordinate.

pmrv (approved):
> @pmrv I did some test calculations from my side. For my calculations with cartesian poscar, apart from giving the warning message, vasp did the fixing correctly. I also checked the vasp wiki page, but not sure this warning message is not relevant anymore for these newer version of vasp or there is something else (maybe some special case where this is relevant). Hence, I thought it's a good idea to avoid this warning message by just writing the poscar in direct coordinate.

Sure, looks good to me now!

pmrv:
@usaikia Do you want to add more things here or can I merge this?

usaikia:
@pmrv I don't have anything else to add here. You can merge it.

PR #453, opened 2021-11-30
Do not polute resource_path

### LINKED ISSUE #452: s.resource_list sometimes grows without bounds
**Summary**

I have a long running kernel (~1000 cells, a few thousand jobs started) and noticed that simply creating new jobs starts to get really slow (~3min for a Vasp job), because somewhere `Settings.resource_list` gets spurious entries of the `iprpy` path, like this

```python
['/cmmc/u/zora/pyiron/resources',
 '/cmmc/u/system/SLES12/soft/pyiron/dev/pyiron-resources-cmmc',
 '/cmmc/u/system/SLES12/soft/pyiron/dev/anaconda3/share/pyiron',
 '/u/system/SLES12/soft/pyiron/dev/anaconda3/share/iprpy',
 '/u/system/SLES12/soft/pyiron/dev/anaconda3/share/iprpy',
... # 2000 more entries
]
```

I'm on the cluster on pyiron_base version 331642e8, pyiron_atomistic at #444.

I can't reproduce it for normal Vasp jobs, but I suppose one of the job classes adds the `iprpy` path without checking whether it is really there and that then accumulates.  I'll have a grep later to see if I can find it.

pmrv:
Fixes #452

### DIFF
--- a/pyiron_atomistics/lammps/potential.py
+++ b/pyiron_atomistics/lammps/potential.py
@@ -74,7 +74,9 @@ class LammpsPotential(GenericParameters):
             resource_path_lst = state.settings.resource_paths
             for conda_var in ["CONDA_PREFIX", "CONDA_DIR"]:
                 if conda_var in env.keys():  # support iprpy-data package
-                    resource_path_lst += [os.path.join(env[conda_var], "share", "iprpy")]
+                    path_to_add = os.path.join(env[conda_var], "share", "iprpy")
+                    if path_to_add not in resource_path_lst:
+                        resource_path_lst.append(path_to_add)
             for path in relative_file_paths:
                 absolute_file_paths.append(find_potential_file_base(
                     path=path,
--- a/pyiron_atomistics/lammps/potential.py
+++ b/pyiron_atomistics/lammps/potential.py
@@ -74,7 +74,7 @@ class LammpsPotential(GenericParameters):
             resource_path_lst = state.settings.resource_paths
             for conda_var in ["CONDA_PREFIX", "CONDA_DIR"]:
                 if conda_var in env.keys():  # support iprpy-data package
-                    path_to_add = os.path.join(env[conda_var], "share", "iprpy")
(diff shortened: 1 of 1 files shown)


pyiron_atomistics/lammps/potential.py
```diff
@@ -78,7 +78,9 @@ def files(self):
             resource_path_lst = s.resource_paths
             for conda_var in ["CONDA_PREFIX", "CONDA_DIR"]:
                 if conda_var in env.keys():  # support iprpy-data package
-                    resource_path_lst += [os.path.join(env[conda_var], "share", "iprpy")]
+                    path_to_add = os.path.join(env[conda_var], "share", "iprpy")
+                    if path_to_add not in resource_path_lst:
+                        resource_path_lst.append(path_to_add)
```
liamhuber, line 79: cf. my larger concerns in the issue, but here at least we also need to cast the path as posix

pmrv: Aren't all operations on this going to be on the filesystem directly via function in `os.*`?  If so they should be system-dependent paths to work on windows as well.

liamhuber: 💯 yes, pyiron should still run it because we `os.*` on it later, *but* [we claim](https://github.com/pyiron/pyiron_base/blob/97e1d0a523c45720f897a5dee906e3ed9efe8449/pyiron_base/state/settings.py#L98) that we *store* paths in posix form. I'm by no means married to that claim, I just carried it over during my refactor...but as long as those words are there we should stay true to them.

pmrv: Ah, I'll use that then and work on a "nicer" interface elsewhere.

liamhuber:
@pmrv you're going to need to merge in master and resolve the conflicts (easy) -- I merged the `state` PR so `s` --> `state.settings`.

liamhuber (approved):
Other than the merge conflict (which I think github would anyhow force you to resolve) I'd say it's good to go.

PR #458, opened 2021-12-03
Add callback function for LAMMPS

samwaseda:
`callback` is one of the features of LAMMPS which makes it extremely efficient while being still very interactive. In short, you can append a function that returns forces which are then added to the total forces calculated in LAMMPS. You can find more info on [this page](https://docs.lammps.org/fix_external.html).

Example I: Add random forces

```python
def random_forces(x, *args):
    return 0.1 * np.random.randn(*x.shape)

lmp = pr.create_job('Lammps', 'random_forces')
lmp.structure = pr.create.structure.bulk('Ni', cubic=True).repeat(3)
lmp.server.run_mode.interactive = True
lmp.set_callback(random_forces)
lmp.calc_md()
lmp.run()
lmp.interactive_close()
```

Example II: Push H in (1, 1, 1)-direction

```python
class CallBack:
    def __init__(self, job):
        self.job = job

    def push_hydrogen(self, x, *args):
        f = np.zeros_like(self.job.structure.positions)
        f[self.job.structure.select_index('H')[0]] = np.array(3*[0.5])
        return f

lmp = pr.create_job('Lammps', 'push_hydrogen')
lmp.structure = pr.create.structure.bulk('Ni', cubic=True).repeat(3)
lmp.structure += pr.create.structure.atoms(
    elements=['H'], positions=[lmp.structure.analyse.get_voronoi_vertices()[0]]
)
callback = CallBack(lmp)
lmp.server.run_mode.interactive = True
lmp.set_callback(callback.push_hydrogen)
lmp.calc_md()
lmp.run()
lmp.interactive_close()
```

### DIFF
--- a/pyiron_atomistics/lammps/interactive.py
+++ b/pyiron_atomistics/lammps/interactive.py
@@ -245,9 +245,47 @@ class LammpsInteractive(LammpsBase, GenericInteractive):
 
     def set_callback(self, function, n_call=1, n_apply=1):
         """
+        **********************
         *** Expert feature ***
-        Set callback function that will be evaluated every `n_call` steps and applied every
-        `n_apply` steps to modify forces of atoms.
+        **********************
+
+        Set callback function that modifies forces.
+        
+        Args:
+            function (function): User-defined function that returns forces (see below)
+            n_call (int): Make callback every `n_call` steps
+            n_apply (int): Apply callback forces every `n_apply` steps
+
+        `function` must have the following form:
+
+        ```
+        def function(positions, ntimestep, nlocal):
+            your_evaluation
+            return forces
+        ```
+        where `positions` is the positions of all atoms on the local processor, `ntimestep` is the
+        current timestep and `nlocal` is the number of atoms on the current processor. `forces` 
+        must be of the shape `(n_atoms, 3)`. The total translational force will be eliminated
+        inside pyiron.
+
+        Example: Add random forces
+
+        ```
+        from pyiron import Project
+
+        def random_forces(x, *args):
+            return 0.1 * np.random.randn(*x.shape)
+
+        pr = Project('RANDOM')
+        lmp = pr.create.job.Lammps('random_forces')
+        lmp.structure = your_structure
+        lmp.potential = your_potential
+        lmp.server.run_mode.interactive = True
+        lmp.set_callback(random_forces)
+        lmp.calc_md()
+        lmp.run()
+        lmp.interactive_close()
+        ```
         """
         if not self.server.run_mode.interactive:
             raise AssertionError('Callback works only in interactive mode')
--- a/pyiron_atomistics/lammps/interactive.py
+++ b/pyiron_atomistics/lammps/interactive.py
@@ -243,7 +243,7 @@ class LammpsInteractive(LammpsBase, GenericInteractive):
         self._reset_interactive_run_command()
         self.interactive_structure_setter(self.structure)
 
-    def set_callback(self, function, n_call=1, n_apply=1):
+    def set_callback(self, function, n_call=1, n_apply=1, overload_internal_callback=False):
         """
         **********************
         *** Expert feature ***
@@ -255,6 +255,10 @@ class LammpsInteractive(LammpsBase, GenericInteractive):
             function (function): User-defined function that returns forces (see below)
             n_call (int): Make callback every `n_call` steps
             n_apply (int): Apply callback forces every `n_apply` steps
+            overload_internal_callback (bool): Whether to overload internal callback (see below).
+                Overloading the internal callback function will have the advantage that the code
+                will not need to do expensive copying, BUT it is extremely error-prone. Make
+                sure that the code works without overloading the internal callback function first.
 
         `function` must have the following form:
 
@@ -267,6 +271,27 @@ class LammpsInteractive(LammpsBase, GenericInteractive):
         current timestep and `nlocal` is the number of atoms on the current processor. `forces`
         must be of the shape `(n_atoms, 3)`. The total translational force will be eliminated
         inside pyiron.
+
+        If `overload_internal_callback` is set to `True`, then `function` must have the following
+        form:
+
+        ```
+        def function(ptr, timestep, nlocal, ids, x, fexternal):
+            your_evaluation
+        ```
+
+        with the following arguemnts:
+
+        - `ptr`: pointer provided by and simply passed back to external driver
+        - `timestep`: current LAMMPS timestep
+        - `nlocal`: # of atoms on this processor
+        - `ids`: list of atom IDs on this processor
+        - `x`: coordinates of atoms on this processor
+        - `fexternal`: forces to add to atoms on this processor
+
+        Note: Do NOT overwrite `fexternal`, because it points to the internal memory of LAMMPS and
+        therefore overwriting it will erase its functionality. E.g. DO `fexternal.fill(0)` and NOT
+        `fexternal = np.zeros_like(x)`.
 
         Example: Add random forces
 
@@ -293,6 +318,8 @@ class LammpsInteractive(LammpsBase, GenericInteractive):
         self.input.control['fix___callback'] = 'all external pf/callback {} {}'.format(
             n_call, n_apply
         )
+        if overload_internal_callback:
+            self._callback = function
 
     def _callback(self, caller, ntimestep, nlocal, tag, x, fext):
         tags = tag.flatten().argsort()
--- a/pyiron_atomistics/lammps/interactive.py
+++ b/pyiron_atomistics/lammps/interactive.py
@@ -250,7 +250,7 @@ class LammpsInteractive(LammpsBase, GenericInteractive):
         **********************
 
         Set callback function that modifies forces.
-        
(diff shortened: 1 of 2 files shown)


pyiron_atomistics/lammps/interactive.py
```diff
@@ -242,6 +243,84 @@ def interactive_initialize_interface(self):
         self._reset_interactive_run_command()
         self.interactive_structure_setter(self.structure)
 
+    def set_callback(self, function, n_call=1, n_apply=1, overload_internal_callback=False):
+        """
+        **********************
+        *** Expert feature ***
+        **********************
+
+        Set callback function that modifies forces.
```
liamhuber: add example 1 to docstring

pyiron_atomistics/lammps/interactive.py
```diff
@@ -242,6 +243,84 @@ def interactive_initialize_interface(self):
...
+        `function` must have the following form:
+
+        ```
+        def function(positions, ntimestep, nlocal):
+            your_evaluation
+            return forces
+        ```
+        where `positions` is the positions of all atoms on the local processor, `ntimestep` is the
+        current timestep and `nlocal` is the number of atoms on the current processor. `forces`
+        must be of the shape `(n_atoms, 3)`. The total translational force will be eliminated
```
liamhuber, line 270:
```suggestion
        must be of the shape `(n_atoms, 3)`. The net translational force will be eliminated
```

pyiron_atomistics/lammps/interactive.py
```diff
@@ -242,6 +243,84 @@ def interactive_initialize_interface(self):
...
+        inside pyiron.
+
+        If `overload_internal_callback` is set to `True`, then `function` must have the following
```
liamhuber: tbh, even with the documentation here I still don't get what this is doing or when I would want to use it. Since you put a big "experts only" warning at the top, I can live with that, but just FYI in case you can think of alternative documentation.

pyiron_atomistics/lammps/interactive.py
```diff
@@ -242,6 +243,84 @@ def interactive_initialize_interface(self):
...
+        from pyiron import Project
+
+        def random_forces(x, *args):
+            return 0.1 * np.random.randn(*x.shape)
+
+        pr = Project('RANDOM')
+        lmp = pr.create.job.Lammps('random_forces')
+        lmp.structure = your_structure
+        lmp.potential = your_potential
+        lmp.server.run_mode.interactive = True
```
liamhuber:
```suggestion
        lmp.interactive_open()
```

For symmetry with the `interactive_close()` at the end.

You could also consider delaying this until [base 565](https://github.com/pyiron/pyiron_base/pull/565) is finished and merged so that you can just use `with lmp.interactive_open() as lmp_int:` and get rid of the close entirely, but if you'd rather merge this sooner that's cool too.

pyiron_atomistics/lammps/interactive.py
```diff
@@ -242,6 +243,84 @@ def interactive_initialize_interface(self):
...
+        """
+        **********************
+        *** Expert feature ***
+        **********************
+
+        Set callback function that modifies forces.
+
+        Args:
+            function (function): User-defined function that returns forces (see below)
+            n_call (int): Make callback every `n_call` steps
```
liamhuber: Specify defaults in docstring

tests/lammps/test_interactive.py
```diff
@@ -140,6 +140,23 @@ def test_calc_minimize_input(self):
...
                          self.minimize_job._interactive_library._command))
 
+    def test_callback(self):
+        def f(x, nt, nl):
+            return np.linspace(0, 1, np.prod(x.shape)).reshape(-1, 3)
+        v = f(self.job.structure.positions, None, None)
+        v -= np.mean(v, axis=0)
+        self.job.server.run_mode.modal = True
+        self.assertRaises(AssertionError, self.job.set_callback, f)
+        self.job.server.run_mode.interactive = True
```
liamhuber, line 150:
```suggestion
        self.job.interactive_open()
```

And for good measure this should be coupled with a close at the end of the function, although I guess this class is inheriting from `TestWithCleanProject` such that `self.job` is anyhow getting safely destroyed. Still seems like best practice though!

pyiron_atomistics/lammps/interactive.py
```diff
@@ -649,3 +731,14 @@ def interactive_close(self):
                 if "interactive" in h5.list_groups():
                     for key in h5["interactive"].list_nodes():
                         h5["generic/" + key] = h5["interactive/" + key]
+
+
+class _CallBack:
```
liamhuber: It's a private class so I don't mind if it's undocumented, *but* if you can think of something useful to say to Fred McFredderson, the MSc student who will join us in five years and wants to fix/extend this class and is looking for some high-level description of intent/promises/etc, it wouldn't hurt.

pyiron_atomistics/lammps/interactive.py
```diff
@@ -242,6 +243,84 @@ def interactive_initialize_interface(self):
...
+        lmp = pr.create.job.Lammps('random_forces')
+        lmp.structure = your_structure
+        lmp.potential = your_potential
+        lmp.server.run_mode.interactive = True
+        lmp.set_callback(random_forces)
+        lmp.calc_md()
+        lmp.run()
+        lmp.interactive_close()
+        ```
+        """
```
liamhuber, line 321: I guess you should also include something along the lines of example 2 in the docstring. If I understand correctly, it's this example that makes differences between `n_apply` and `n_call` relevant -- i.e. with a bare function they must always be the same, but with a class we can use `n_apply < n_call` and use the non-applied calls to build up some internal counter, e.g. using Welfords Online algo to track running means and exploit them in the intermittently applied force.

liamhuber (commented):
Super cool, I think this is absolutely very useful. One small syntax tweak requested in the code itself, and a whole bunch of (sometimes optional) documentation request.

Looking at it from 30k ft, the big issue I see is how to connect this to our serialization and reproducibility. As far as I can see, right now if you activate this functionality the final serialized job will give no indication of what was done, and so you'll just strangely get a trajectory whose forces and evolution don't align with your potential+thermo/barostatting. I guess this is already the case for jobs we mess with interactively. Since you explicitly list this as an "expert feature", I don't think this concern should stop this PR from merging, and I don't have a good solution myself right now anyhow -- but we should think about how to handle this...maybe some sort of `my_callback = pr.atomistics.lammps.Callback(fnc=my_function_defined_elsewhere); lmp.set_callback(my_callback.fnc)` (or `lmp.set_callback(my_callback)`, where we check if the argument is an instance of our fancy new class and parse accordingly), and then handle serialization of the raw function code under the hood?

pyiron_atomistics/lammps/interactive.py
samwaseda: Ok since it's just about DocString and the functionality will anyway not change, I keep the current one (i.e. with `interactive_open/_close`), but I'd use `with` if it happens to be already ready by the time this PR is to be merged.

pyiron_atomistics/lammps/interactive.py
samwaseda: I'm actually waiting for the pyiron meeting today to see what would be the final structure, because creating this class might not be a good idea in the first place anyway, but I promise that I'll add some text before I merge it.

pyiron_atomistics/lammps/interactive.py
samwaseda: yeah actually I don't know how things change if `n_apply` and `n_call` vary. I'm gonna check it this afternoon

tests/lammps/test_interactive.py
samwaseda: I guess there'd be a problem if the job is closed because it's also used elsewhere. Isn't it destroyed only when all tests are over?

liamhuber: > I guess there'd be a problem if the job is closed because it's also used elsewhere. 

Aha. First, you're right -- it doesn't inherit from `TestWithCleanProject` at all, it's entirely written by hand. Using the same job elsewhere is super sketchy to me as it means behaviour in one test has a chance to infect behaviour in an independent test. There is some sort of cleaning happening in the [`setUp`](https://github.com/pyiron/pyiron_atomistics/blob/d2664693698f189e23a4fe3fbad603c4246d2c91/tests/lammps/test_interactive.py#L25) method, but I am not familiar enough with interactive jobs to say whether this is bullet-proof.

>  Isn't it destroyed only when all tests are over?

Yes, ultimately it gets killed in [`tearDownClass`](https://github.com/pyiron/pyiron_atomistics/blob/d2664693698f189e23a4fe3fbad603c4246d2c91/tests/lammps/test_interactive.py#L62)

I don't think fixing these things is the responsibility of this PR, but the current implementation of `TestLammpsInteractive` where the same job is reused in each test seems dangerous to me and should be fixed later in a separate PR.

pyiron_atomistics/lammps/interactive.py
samwaseda: You mean example 2? (1 is already there)

PR #462, opened 2021-12-08
Refactor `symmetrize_vectors`

samwaseda:
I realised that I needed the permutations more often elsewhere than I had imagined so I refactored it. I hope the description in the DocString is acceptable...


samwaseda:
Since in principle nothing changed I'll merge it as soon as the tests finish.

PR #479, opened 2021-12-14
Decrease convergence goal

jan-janssen:


liamhuber (approved):
Can't hurt. These tests are already annoyingly slow on my local machine anyway, and I don't see that we lose anything with this.

PR #483, opened 2021-12-17
Use current structure instead of output to update previous structure

### LINKED ISSUE #481: Why does interactive LAMMPS become slower?
I just realised that interactive LAMMPS gets slower and slower. For this, I created the following short example:

```python
pr = Project('RUNTIME')
import time

structure = pr.create.structure.bulk('Ni').repeat(4)
lmp = pr.create.job.Lammps('time')
lmp.potential = '1995--Angelo-J-E--Ni-Al-H--LAMMPS--ipr1'
lmp.interactive_open()

n_iter = 100
t_lst = np.zeros(n_iter)
t = time.process_time()
for i in range(n_iter):
    for j in range(10):
        lmp.structure = structure.copy()
        lmp.structure.positions += np.random.randn(len(lmp.structure), 3)/1000
        lmp.run()
    t_lst[i] = time.process_time() - t
```

And I observed the following evolution of time (of course I plotted `np.diff(t_lst)` and not `t_lst`). Why does interactive LAMMPS get slower and slower?

Unknown

samwaseda:
The issue is discussed in #481. Now I also did a performance test and got the following results:

Unknown
Unknown-1

This means in particular that after 1000 steps (since each iteration contains 10 steps) the new implementation is going to be around 30 times faster than the old one. 

closes #481

### DIFF
--- a/pyiron_atomistics/atomistics/job/interactive.py
+++ b/pyiron_atomistics/atomistics/job/interactive.py
@@ -156,7 +156,7 @@ class GenericInteractive(AtomisticGenericJob, InteractiveBase):
             self.interactive_indices_setter(self._structure_current.indices)
 
     def interactive_cell_organizer(self):
-        if not np.allclose(
+        if self._generic_input["calc_mode"] != 'static' or not np.allclose(
             self._structure_current.cell,
             self._structure_previous.cell,
             rtol=1e-15, atol=1e-15,
@@ -168,7 +168,7 @@ class GenericInteractive(AtomisticGenericJob, InteractiveBase):
                 del self.interactive_input_functions['cell']
 
     def interactive_positions_organizer(self):
-        if not np.allclose(
+        if self._generic_input["calc_mode"] != 'static' or not np.allclose(
             self._structure_current.positions,
             self._structure_previous.positions,
             rtol=1e-15,
--- a/pyiron_atomistics/atomistics/job/interactive.py
+++ b/pyiron_atomistics/atomistics/job/interactive.py
@@ -266,7 +266,7 @@ class GenericInteractive(AtomisticGenericJob, InteractiveBase):
         return self.initial_structure.get_volume()
 
     def _update_previous_structure(self):
-        """Update the previous structure to the last step configuration"""
+        """Update the previous structure to the last step configuration."""
         self._structure_previous = self.structure.copy()
 
     @staticmethod
--- a/pyiron_atomistics/atomistics/job/interactive.py
+++ b/pyiron_atomistics/atomistics/job/interactive.py
@@ -118,7 +118,6 @@ class GenericInteractive(AtomisticGenericJob, InteractiveBase):
             self.interactive_initialize_interface()
         if self._structure_previous is None:
             self._structure_previous = self.structure.copy()
-        self._update_previous_structure()
         if self._structure_current is not None:
             if (
                 len(self._structure_current) != len(self._structure_previous)
@@ -134,6 +133,7 @@ class GenericInteractive(AtomisticGenericJob, InteractiveBase):
             else:
                 self._logger.debug("Generic library: structure changed!")
                 self.interactive_structure_setter(self._structure_current)
+        self._update_previous_structure()
 
     def interactive_index_organizer(self):
         index_merge_lst = self._interactive_species_lst.tolist() + list(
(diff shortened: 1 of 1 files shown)


liamhuber (approved):
Lgtm. That timing plot in the pr description is a beautiful thing. @raynol-dsouza i guess this is very nice news for every Protocol that exploits lammps

pmrv (approved):
I must admit, I don't fully understand how this changes the timing, but it seems to work and results in less code, so :+1:.

samwaseda:
> I must admit, I don't fully understand how this changes the timing, but it seems to work and results in less code, so 👍.

It's the output values (`output.indices` etc.) that uses `GenericInteractiveOutput._lst_from_cache`. That's why.

niklassiemer:
I am not at all into this so this might be completely off. However, do I understand correctly that this PR just uses the last structure which had been set to the job as the last structure? Is it not also possible to generate multiple structures per run call? Then we would ignore them wouldn't we?

samwaseda:
> I am not at all into this so this might be completely off. However, do I understand correctly that this PR just uses the last structure which had been set to the job as the last structure? Is it not also possible to generate multiple structures per run call? Then we would ignore them wouldn't we?

The structure comparison is done in order to see which values have changed between two runs. So as long as it allows to see which values have to be updated before the next run, it doesn't matter how many structures there are.

pyiron_atomistics/atomistics/job/interactive.py
```diff
@@ -266,27 +266,8 @@ def interactive_volume_getter(self):
...
-        else:
-            el_lst = self._interactive_species_lst.tolist()
-        self._structure_previous = self._structure_previous.__class__(
-            positions=positions,
-            cell=cell,
-            indices=indices,
-            species=[self._periodic_table.element(el) for el in el_lst],
-        )
+        """Update the previous structure to the last step configuration."""
+        self._structure_previous = self.structure.copy()
```
jan-janssen, line 270: This is not the same - for example when running `calc_minimize()` in interactive mode.

jan-janssen (changes requested):
The interactive mode is divided in two parts, a beginners mode `job.structure = new_structure` and an experts mode `job.interactive_positions_setter(new_positions)`. The expert mode is designed for efficiency and the beginners mode is designed for reliability. With the current changes to the beginners mode the structure of the job object `job.structure` is no longer the same as the structure in the interactive job - for example when running lammps with `calc_minimize()` in interactive mode. I think this is a rather risky modification.

samwaseda:
> The interactive mode is divided in two parts, a beginners mode `job.structure = new_structure` and an experts mode `job.interactive_positions_setter(new_positions)`. The expert mode is designed for efficiency and the beginners mode is designed for reliability. With the current changes to the beginners mode the structure of the job object `job.structure` is no longer the same as the structure in the interactive job - for example when running lammps with `calc_minimize()` in interactive mode. I think this is a rather risky modification.

Yeah ok I remember the discussions. But that means in `calc_minimize`, we always have to update at least the positions, and if the pressure is specified also the cell, correct? Doesn't it mean it makes more sense to update them all the time anyway?

samwaseda:
@jan-janssen I added the changes we discussed yesterday.

pyiron_atomistics/atomistics/job/interactive.py
```diff
@@ -118,7 +118,6 @@ def run_if_interactive(self):
             self.interactive_initialize_interface()
         if self._structure_previous is None:
             self._structure_previous = self.structure.copy()
-        self._update_previous_structure()
```
jan-janssen, line 121: I do not understand this part - if you move `update_previous_structure()` to line 136 - then in line 123 the structure is not yet updated.

pyiron_atomistics/atomistics/job/interactive.py
```diff
@@ -156,7 +156,7 @@ def interactive_index_organizer(self):
             self.interactive_indices_setter(self._structure_current.indices)
 
     def interactive_cell_organizer(self):
-        if not np.allclose(
+        if self._generic_input["calc_mode"] != 'static' or not np.allclose(
```
jan-janssen, line 159: This is not necessary for `calc_minimize()` only `calc_minimize(pressure=0)` right? And analogously it is not necessary for `nvt` and `nve` but `npt`.

pyiron_atomistics/atomistics/job/interactive.py
samwaseda: Previous structure has to be set to the current structure only after the variable comparisons are done, i.e. after the run.

pyiron_atomistics/atomistics/job/interactive.py
samwaseda: Is the information about `nvt`, `nve` etc.  included in the input?

pyiron_atomistics/atomistics/job/interactive.py
jan-janssen: Ok, I see.

pyiron_atomistics/atomistics/job/interactive.py
jan-janssen: Not yet but we could include it in https://github.com/pyiron/pyiron_atomistics/blob/master/pyiron_atomistics/atomistics/job/atomistic.py#L226

samwaseda: So we could go ahead with this version for now, right? I don't think communicating the cell size would significantly change something.

Issue #435, opened 2021-11-23
VASP selective dynamics error


usaikia:
I used pyiron to fix position of an atom during vasp ionic relaxation calculation (isif=2). After running the job I found a line in the OUTCAR saying:
"""
using selective dynamics as specified on POSCAR
   WARNING: If single coordinates had been selected the selection of coordinates
            is made according to the corresponding   d i r e c t   coordinates!
            Don't support selection of single cartesian coordinates -- sorry ...
""" 
Is there any way to write the input structure file (POSCAR) for vasp run in Direct coordinate system in pyiron?

Here is my job submission script:

h = pr.create.job.Vasp('test')
basis = pr.create.structure.bulk('Fe', cubic=True)
basis.set_initial_magnetic_moments(2*[3.0])
basis.add_tag(selective_dynamics=[False, False, False])
basis.selective_dynamics[0] = [True, True, True]
h.structure = basis
h.calc_minimize()
h.input.incar['ISIF'] = 2
h.run()

pmrv:
Our `write_poscar` function appears to support direct coordinates

https://github.com/pyiron/pyiron_atomistics/blob/788d9c7e4c2f5251536c1d9794aaf2e22a07dafa/pyiron_atomistics/vasp/structure.py#L82

I guess the way to implement this is to change the `Input` class to ask for direct coordinates once it detects some atom coordinates are fixed here

https://github.com/pyiron/pyiron_atomistics/blob/788d9c7e4c2f5251536c1d9794aaf2e22a07dafa/pyiron_atomistics/vasp/base.py#L1802

Feel free to give this a shot in a PR and add me as a reviewer.

usaikia:
@pmrv Thanks! I will open a PR to implement this.

PR #492, opened 2022-01-03
Poscar control

### LINKED ISSUE #491: Write POSCAR in direct coordinate
I read [here](https://github.com/pyiron/pyiron_atomistics/pull/448) the POSCAR file will be written in direct coordinates when the selective dynamic is on. What if the problem does not have sth to do with selective dynamic and wants to have POSCAR file in direct coordinate? CONTCAR file will be generated in direct cordinate

sudarsan-surendralal:
With these changes, we can now control how POSCAR files are written (in direct/cartesian coordinates) for VASP jobs.

### DIFF
--- a/pyiron_atomistics/vasp/base.py
+++ b/pyiron_atomistics/vasp/base.py
@@ -7,7 +7,6 @@ import os
 import posixpath
 import subprocess
 import numpy as np
-
 from pyiron_atomistics.dft.job.generic import GenericDFTJob
 from pyiron_atomistics.vasp.potential import VaspPotential, VaspPotentialFile, VaspPotentialSetter, Potcar, \
     strip_xc_from_potential_name
@@ -81,6 +80,7 @@ class VaspBase(GenericDFTJob):
         self._potential = VaspPotentialSetter([])
         self._compress_by_default = True
         self.get_enmax_among_species = get_enmax_among_potentials
+        self._write_in_direct_coordinates = False
         state.publications.add(self.publication)
 
     @property
@@ -204,6 +204,20 @@ class VaspBase(GenericDFTJob):
                 "write_resolved_dos, can either be True, False or 0, 1, 2, 5, 10, 11, 12."
             )
         self.input.incar["LORBIT"] = resolved_dos
+
+    @property
+    def write_in_direct_coordinates(self) -> bool:
+        """
+        True if the POSCAR file input is to be written in direct coordinates (as opposed to cartesian coordinates)
+
+        Returns:
+            bool
+        """
+        return self.input.write_in_direct_coordinates
+
+    @write_in_direct_coordinates.setter
+    def write_in_direct_coordinates(self, val: bool):
+        self.input.write_in_direct_coordinates = val
 
     @property
     def sorted_indices(self):
@@ -1805,8 +1819,22 @@ class Input:
         self.incar = Incar(table_name="incar")
         self.kpoints = Kpoints(table_name="kpoints")
         self.potcar = Potcar(table_name="potcar")
-
         self._eddrmm = "warn"
+        self._write_in_direct_coordinates = False
+
+    @property
+    def write_in_direct_coordinates(self):
+        """
+        True if the POSCAR file input is to be written in direct coordinates (as opposed to cartesian coordinates)
+
+        Returns:
+            bool
+        """
+        return self._write_in_direct_coordinates
+
+    @write_in_direct_coordinates.setter
+    def write_in_direct_coordinates(self, val: bool):
+        self._write_in_direct_coordinates = val
 
     def write(self, structure, modified_elements, directory=None):
         """
@@ -1829,6 +1857,7 @@ class Input:
             structure,
             filename=posixpath.join(directory, "POSCAR"),
             write_species=not do_not_write_species,
+            cartesian=not self.write_in_direct_coordinates
         )
 
     def to_hdf(self, hdf):
--- a/tests/vasp/test_vasp.py
+++ b/tests/vasp/test_vasp.py
@@ -320,7 +320,7 @@ class TestVasp(unittest.TestCase):
         self.job_complete.restart_file_list.append(
             posixpath.join(file_directory, "OUTCAR")
         )
-        self.job_complete.run(run_mode="manual")
+        self.job_complete.run(run_mode="manual", delete_existing_job=True)
         self.job_complete.status.collect = True
         self.job_complete.run()
         nodes = [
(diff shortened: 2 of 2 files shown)


samwaseda:
In principle I find it ok that we have this tag, but it's not very clear to me what we need it for. If there are special features in VASP that can be used only if POSCAR is written in direct coordinates, I might prefer to make a function available that makes the purpose clear.

sudarsan-surendralal:
> In principle I find it ok that we have this tag, but it's not very clear to me what we need it for. If there are special features in VASP that can be used only if POSCAR is written in direct coordinates, I might prefer to make a function available that makes the purpose clear.

Scientifically, this makes no difference (to me at least). My guess is that users who like to inspect the VASP input files find the positions in direct coordinates easier to interpret. A couple of users seem to want this feature (https://github.com/pyiron/pyiron_atomistics/issues/435 & https://github.com/pyiron/pyiron_atomistics/issues/491) and it seemed like an easy enough fix.

samwaseda:
> My guess is that who like to inspect the VASP input files find the positions in direct coordinates easier to interpret.

Hmmm but that sounds like something that shouldn't be needed in the first place in pyiron. Frankly, rather than cluttering the job-level entries with even more functions that shouldn't mean anything to anyone, I'd offer the same solution as the one you mentioned in [the issue](https://github.com/pyiron/pyiron_atomistics/issues/491).

sudarsan-surendralal:
> > My guess is that who like to inspect the VASP input files find the positions in direct coordinates easier to interpret.
> 
> Hmmm but that sounds like something that shouldn't be needed in the first place in pyiron. Frankly, rather than cluttering the job-level entries with even more functions that shouldn't mean anything to anyone, I'd offer the same solution as the one you mentioned in [the issue](https://github.com/pyiron/pyiron_atomistics/issues/491).

Not sure if we can (or even should) enforce all pyiron users to avoid such practices. If it helps users debug their VASP calculations then I personally think having this functionality is rather useful.

ligerzero-ai:
We must have this functionality - +1 for request.

Reason: Specifying cleavage planes in GBs is way more intuitively done with fractional coordinates compared to cartesian coordinates, especially when dealing with different sized cells.

liamhuber:
> We must have this functionality - +1 for request.
> 
> Reason: Specifying cleavage planes in GBs is way more intuitively done with fractional coordinates compared to cartesian coordinates, especially when dealing with different sized cells.

Is there a reason cleavage planes can't be specified on the `Atoms` object? It's not clear to me why we'd need to go into VASP IO files for this.

ligerzero-ai:
Of course, in an ideal world we never need to leave our notebooks, but in the case of cleavage plane selection, it is almost mandatory to actually view the structure in VESTA (not nglview) to select the appropriate cleavage planes because of relaxation. 

In this case, both the order of the CONTCAR (e.g. #752) as well as the direct coordinates are highly desirable for convenient selection of the cleavage planes after selection.  (i.e. you can check the structure has been cleaved between atoms N and M), all above the selection in the POSCAR are shifted upwards, all below are moved downwards.

In this manner, you can go "down" the structure's Z-coordinates and select the cleavage planes that are necessary to be tested without duplication, as well as errors between different configurations due to relaxations. Inputting 0.51 (frac coords) to cleave the structure is preferable to the cartesian option of 27.48 or whatever.

liamhuber:
> Of course, in an ideal world we never need to leave our notebooks

Ah, I was maybe unclear. I totally understand that we may need to go outside notebooks, I just don't think we should need to move from the pyiron-level of representation to the wrapped-code level of representation.

> In this case, both the order of the CONTCAR (e.g. https://github.com/pyiron/pyiron_atomistics/issues/752) as well as the direct coordinates are highly desirable for convenient selection of the cleavage planes after selection.

This is already completely possible in pyiron. You can build a structure (order "A"), run VASP on it (internally order "B"), get the output structure (back to order "A"), and if you need to use a third party tool that take POSCAR-formatted files, you just write the output structure (note that we simply wrap `ase.io.write`):

```python
some_output_structure.write(filename='order_A_POSCAR', format='vasp')
# OR
some_output_structure.write(filename='order_A_POSCAR_direct_coords', format='vasp', direct=True)
```

Once you know what cutoff you want from your third-party app, there's no problem using that in pyiron either:

```python
cut_struct = some_output_structure[some_output_structure.positions[:, 2] < 27.48]
# OR
cut_struct = some_output_structure[some_output_structure.get_scaled_positions()[:, 2] < 0.51]
```

Since there are already single-line tools to do all of this in post-processing at the pyiron level, I really don't see the need to mess around with how we currently read and write VASP IO files.

ligerzero-ai:
Firstly, thanks for the examples, I will make use of them

However:

Frankly, I'm more of the opinion that functionality should be provided over philosophical musings about what should and shouldn't be implemented.

I am more in agreement with Sudarsan's assessment that rather than forcing someone to think a certain way, it should rather be implemented and be there if someone wants to use it. It doesn't hurt anybody and there is clearly interest in it, as he said.

The fact that I am having this discussion with you, and previously other users have asked proves that there is interest from new VASP->pyiron users for such a functionality. The fact that it doesn't exist and I had to look/ask for it is inelegant - especially the part with the fractional coordinates being "get_scaled_positions" when it should just be frac_coords and cart_coords, or coords_cart and coords_frac, if there is a preference for tab completion to be convenient.

It is not intuitive to me that I would have to write an extra line to specify direct coordinate POSCAR/CONTCAR generation after job completion, when it should be an option provided at job initiation.

Let us consider the workflow:

With option to specify direct coordinates:
High throughput co-segregation study -> Find cleavage path -> download contcars directly

vs

Without option to specify direct coordinate generation:
High throughput co-segregation study -> Find cleavage path -> Loop through jobs, write direct coordinate CONTCARS and then download contcars

As you can see, this is extra work (however little) inflicted upon the user, and is unnecessary, imho.

liamhuber:
> As you can see, this is extra work (however little) inflicted upon the user, and is unnecessary, imho.

Pros:
- Marginally easier for someone who wants to directly use files generated by VASP

Cons:
- New code needs to be written, read, understood, and maintained
- Users have an extra (usually irrelevant) flag to scroll through on one of our core objects (the VASP job).

This is not a hill I'm willing to die on, if you are passionate about it convince one of the reviewers to approve the PR and go for it. For me the costs outweigh the benefits in this case.


> especially the part with the fractional coordinates being "get_scaled_positions" when it should just be frac_coords and cart_coords, or coords_cart and coords_frac

I'm also not super happy with `get_scaled_positions`. A `scaled_positions` property might be nice. (We have `scaled_pos_xyz`, but this looks like the transpose of what we want.)

The attribute names are not chosen willy-nilly though. We use `positions` and not `coords` because we inherit from the ASE library and this is what they use. The term "fractional" is also no some universally preferred term -- "scaled", "fractional", and "direct" all get used with some regularity in the community. You can take a peek at [ASE's docs on format-specific options for structure file writing](https://wiki.fysik.dtu.dk/ase/ase/io/formatoptions.html) to get a sense of how much these get interchanged.

ligerzero-ai:
Would prefer to have this in compared to out, so I'm afraid I will have to insist on it. Sorry!
@pmrv @jan-janssen 

A small note on the variable/setting terminology;
Looking through some of the code, especially that in VASP base, the nomenclature for variables referring to "positions" is "coords". That is part of the reason why I brought this up. IMHO, the mixing of terminology, even if it is inherited from ASE could make the code harder to maintain overall if different terms are used to refer to the same thing, e.g. if I'm searching for "coords" I also have to perform another search for "positions" since they both refer to the same physical trait. I would prefer that if I'm looking for a single physical trait, it should have consistent terms so all of the code that relates to it pops up under one specific term.

It's not something that is critical, just some of my thoughts on the use of multiple terms to refer to the same thing and its implications for maintenance and readability.

PR #497, opened 2022-01-10
Save VASP stresses as matrix

pmrv:
Turns out VASP writes the stress components in non-Voigt order (*WHY?!*) and I missed that in #354 originally.  This provides the fix.

@niklassiemer This is probably worth pointing out in the next CHANGELOG.

### DIFF
--- a/tests/vasp/test_outcar.py
+++ b/tests/vasp/test_outcar.py
@@ -32,6 +32,10 @@ class TestOutcar(unittest.TestCase):
             filename = posixpath.join(direc, f)
             cls.file_list.append(filename)
 
+    def setUp(self):
+        self.addTypeEqualityFunc(np.ndarray, lambda a, b, msg=None: \
+                np.testing.assert_array_equal(a, b, err_msg=msg))
+
     def test_from_file(self):
         for filename in self.file_list:
             self.outcar_parser.from_file(filename=filename)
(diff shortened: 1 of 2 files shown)


liamhuber:
Am I Right to understand this will resolve those weird artifacts in the ml project strain plots?

pmrv:
> Am I Right to understand this will resolve those weird artifacts in the ml project strain plots?

I very much hope so, but I'll need to regenerate some data to say for sure.

pmrv:
I added up needed to rewire things a bit more.  I've cleaned up the tests to check against all our OUTCARs which exposed a latent bug in the old parsing that I've patched over.

pmrv:
Also as discussed in the pyiron meeting, instead of saving as Voigt notation I save the full matrix now that same we do for lammps.

tests/vasp/test_outcar.py
```diff
@@ -32,6 +32,10 @@ def setUpClass(cls):
             filename = posixpath.join(direc, f)
             cls.file_list.append(filename)
 
+    def setUp(self):
```
sudarsan-surendralal, line 35: This is pretty neat! Maybe we could move this (in a separate PR) to `pyiron_base._tests` so that all pyiron unittests inherit this

sudarsan-surendralal (approved):
LGTM!

tests/vasp/test_outcar.py
niklassiemer: This enables `self.assertEqual(np_array1, np.array2)`? Then I really like your suggestion to make it available more broadly!

PR #505, opened 2022-01-19
Spx refactor

samwaseda:
I mostly refactored the SPHInX output. There's however one important change: the output is now `DataContainer`. I think this allows the user to see a more insightful output data when `job['output/generic']` is called. Older SPHInX calculations can still be loaded (and will also work like before), but you can also update the output via `job.update_sphinx()`. And if the job is meant to be kept for a long time, the update should be effectuated at some point, because probably at some point we (or rather **I**) will drop the conventional hdf format support.


samwaseda:
@niklassiemer Black is failing. Is there a possibility to enforce the conversion?

samwaseda:
ah nice thanks :D

PR #499, opened 2022-01-10
Use black formatting

niklassiemer:

PR #507, opened 2022-01-20
Make codebase black

niklassiemer:
Replaces #499 and adds workflows right away

### DIFF
--- a/.github/workflows/black.yml
+++ b/.github/workflows/black.yml
@@ -1,0 +1,20 @@
+# This workflow will install Python dependencies, run tests and lint with a variety of Python versions
+# For more information see: https://help.github.com/actions/language-and-framework-guides/using-python-with-github-actions
+
+name: Black
+
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    - uses: psf/black@stable
+      with:
+        options: "--check --diff"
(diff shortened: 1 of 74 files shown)


niklassiemer:
That was something I did not expect... The result of `black codebase` was after all not accepted by black?!?
```
--- pyiron_atomistics/thermodynamics/hessian.py	2022-01-20 10:20:22.931989 +0000
+++ pyiron_atomistics/thermodynamics/hessian.py	2022-01-20 10:20:33.322222 +0000
@@ -124,15 +124,18 @@
     def interactive_pressures_getter(self):
         return self._pressure
 
     def interactive_cells_setter(self, cell):
         if np.sum(self._stiffness_tensor) != 0:
-            epsilon = np.einsum(
-                "ij,jk->ik",
-                self.structure.cell,
-                np.linalg.inv(self._reference_structure.cell),
-            ) - np.eye(3)
+            epsilon = (
+                np.einsum(
+                    "ij,jk->ik",
+                    self.structure.cell,
+                    np.linalg.inv(self._reference_structure.cell),
+                )
+                - np.eye(3)
+            )
             epsilon = (epsilon + epsilon.T) * 0.5
             epsilon = np.append(
                 epsilon.diagonal(), np.roll(epsilon, -1, axis=0).diagonal()
             )
             pressure = -np.einsum("ij,j->i", self._stiffness_tensor, epsilon)
would reformat pyiron_atomistics/thermodynamics/hessian.py
Oh no! 💥 💔 💥
1 file would be reformatted, 108 files would be left unchanged.
```
Let's see if a second run also captures this change :)

PR #513, opened 2022-01-28
Selected animation

sudarsan-surendralal:
Introducing options  to animate only a portion of the trajectory (in terms of time steps and atoms)

### DIFF
--- a/pyiron_atomistics/atomistics/job/atomistic.py
+++ b/pyiron_atomistics/atomistics/job/atomistic.py
@@ -320,11 +320,12 @@ class AtomisticGenericJob(GenericJobCore, HasStructure):
 
         Args:
             spacefill (bool): If True, then atoms are visualized in spacefill stype
-            show_cell (bool): True if the unitcell is to be shown
+            show_cell (bool): True if the cell boundaries of the structure is to be shown
             stride (int): show animation every stride [::stride]
                           use value >1 to make animation faster
                            default=1
-            particle_size (float): Radius of the spacefills
+            particle_size (float): Scaling factor for the spheres representing the atoms.
+                                    (The radius is determined by the atomic number)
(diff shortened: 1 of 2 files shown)


pyiron_atomistics/atomistics/job/atomistic.py
```diff
@@ -305,25 +306,30 @@ def store_structure(self):
...
+        snapshot_indices: Union[list, np.ndarray] = None
     ):
         """
         Animates the job if a trajectory is present
 
         Args:
-            spacefill (bool):
-            show_cell (bool):
+            spacefill (bool): If True, then atoms are visualized in spacefill stype
+            show_cell (bool): True if the unitcell is to be shown
```
samwaseda: It shows the frame of the total cell and not only the unit cell, right?

pyiron_atomistics/atomistics/job/atomistic.py
```diff
@@ -305,25 +306,30 @@ def store_structure(self):
...
             stride (int): show animation every stride [::stride]
                           use value >1 to make animation faster
                            default=1
-            center_of_mass (bool):
+            particle_size (float): Radius of the spacefills
```
samwaseda: This is slightly outside of this PR, but `particle_size` in `plot3d()` is not  the radius, but the multiplication factor of the base radius, which is given by the atomic number. It might be helpful to add it at some point (or maybe in this PR already), because I find it difficult to understand in `animate_structure` when small atoms like H or C are shown in the same size as Ni or Fe.

samwaseda (approved):
Looks good to me :)

pyiron_atomistics/atomistics/job/atomistic.py
sudarsan-surendralal: Thanks. I've changed the docstrings

PR #518, opened 2022-02-01
Enhancement: extract charges, energies, and forces from ASE Atoms obj…

lxknll:
As discussed with @pmrv in [this merge request of pyiron_contrib](https://github.com/pyiron/pyiron_contrib/pull/315), it could be interesting to add this functionality to the `StructureStorage` container. 
The proposed changes will automatically detect an ASE Atoms object, extract atomic charges, energies and forces from it, and store them.

@pmrv is this what you had in mind?

Thanks for your comments!

### DIFF
--- a/pyiron_atomistics/atomistics/structure/structurestorage.py
+++ b/pyiron_atomistics/atomistics/structure/structurestorage.py
@@ -13,6 +13,9 @@ from pyiron_atomistics.atomistics.structure.atom import Atom
 from pyiron_atomistics.atomistics.structure.atoms import Atoms
 from pyiron_atomistics.atomistics.structure.has_structure import HasStructure
 
+from ase.atoms import Atoms as ASEAtoms
+from pyiron_atomistics.atomistics.structure.atoms import ase_to_pyiron
+
 
 class StructureStorage(FlattenedStorage, HasStructure):
     """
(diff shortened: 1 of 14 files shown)


pyiron_atomistics/atomistics/structure/structurestorage.py
```diff
@@ -175,6 +178,16 @@ def add_structure(self, structure, identifier=None, **arrays):
             **kwargs: additional arrays to store for structure
         """
 
+        # Extract energies, forces, and atomic charges from ASE Atoms objects.
+        if isinstance(structure, ASEAtoms):
```
pmrv, line 182:
```suggestion
        if isinstance(structure, ASEAtoms) and not isinstance(structure, Atoms):
```
otherwise this will be true for our `Atoms` as well.

pyiron_atomistics/atomistics/structure/structurestorage.py
```diff
@@ -175,6 +178,16 @@ def add_structure(self, structure, identifier=None, **arrays):
...
+            arrays['charges'] = structure.get_initial_charges()
+
+            if structure.calc is not None:
+                arrays['energy'] = structure.calc.get_potential_energy()
+                arrays['forces'] = structure.calc.get_forces()
```
pmrv: Have you checked that this gives the right array shapes and such for `charges`, `energy` and `forces`?  `FlattenedStorage` should do this automatically, but it's best to check.

pmrv (approved):
Looks generally good to me.  Does ASE have a dummy calculator?, if so would be nice to add tests as well.

pmrv:
I'd still like to have this in, @lxknll anything blocking this?

PR #519, opened 2022-02-02
Update Readthedocs

jan-janssen:


jan-janssen:
I removed the Github Workflow which builds the docs as it is anyway inconsistent with the read the docs builds.

PR #526, opened 2022-02-07
Support cube files without atoms

### LINKED ISSUE #525: Read CubeFile without atoms
**Summary**

The current implementation of `read_cube_file` in `pyiron.atomistics.volumetric.generic` assumes that there is at least one atom present in the cube file. However, the electrochemistry hackathon team needs a version which also reads 0-atoms files, so that we can access the volumetric data.

0.4.6 - cluster

**Expected Behavior**
It returns an object with volumetric data and empty list of atoms oder `_atoms` not set at all.

**Actual Behavior**
It refuses to read the cube data.

**Steps to Reproduce**

pmrv:
Fixes #525.

I've cross-checked with [this](http://paulbourke.net/dataformats/cube/) definition of cube files, is that correct @fldei101, @sudarsan-surendralal what code are we targeting here?

I haven't tested this, because I have no example files, but it should be ok.  One issue might be that this will leave `self._atoms` undefined which could break assumptions of downstream code.

### DIFF
--- a/pyiron_atomistics/atomistics/volumetric/generic.py
+++ b/pyiron_atomistics/atomistics/volumetric/generic.py
@@ -399,7 +399,9 @@ class VolumetricData(object):
                     pos_data = np.array([pos_data])
                 atomic_numbers = np.array(pos_data[:, 0], dtype=int)
                 positions = pos_data[:, 2:]
-                self._atoms = Atoms(numbers=atomic_numbers, positions=positions, cell=cell)
+                self._atoms = Atoms(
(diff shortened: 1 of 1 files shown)


niklassiemer:
> I haven't tested this, because I have no example files, but it should be ok. One issue might be that this will leave `self._atoms` undefined which could break assumptions of downstream code.

Maybe we should warn in this case?

sudarsan-surendralal (approved):
LGTM!

jan-janssen:
@pmrv I guess this can be merged - looks good to me

jan-janssen (approved):

PR #527, opened 2022-02-07
Remove _QhullUser import

niklassiemer:
I love type hints, however, not if we have to reach to a private class of a private module.


jan-janssen (approved):
Looks good to me

PR #539, opened 2022-02-15
replace triu by triu_indices

### LINKED ISSUE #246: ElasticTensor bug when use_pressure=False
**Summary**
Elastic Tensor jobs work using pressure, but not using energy my strructures. When the "use_pressure" = False option is set they throw:

```python
_fit_coeffs_with_energies(strain, energy, volume, rotations, max_polynomial_order, fit_first_order)
    150     coeff = np.triu(np.ones((6,6))).flatten()
    151     # Multiply upper triangle with upper triangle coeffs (v.s.)
--> 152     coeff[coeff!=0] *= reg.coef_[:21]*eV_div_A3_to_GPa
    153     coeff = coeff.reshape(6,6)
    154     coeff = 0.5*(coeff+coeff.T)

ValueError: operands could not be broadcast together with shapes (21,) (18,) (21,) 
```

**Steps to Reproduce**

Run a ElasticTensor calculation with the attached structure (lammps format, original atom types are Cu and Zr, but that shouldn't matter I guess) using the use_pressure=False option.
[structure.txt](https://github.com/pyiron/pyiron_atomistics/files/6709559/structure.txt)

samwaseda:
Closes #246

### DIFF
--- a/pyiron_atomistics/atomistics/master/elastic.py
+++ b/pyiron_atomistics/atomistics/master/elastic.py
@@ -143,8 +143,7 @@ def _fit_coeffs_with_energies(
         derivative=False,
     )
     strain_voigt = _convert_to_voigt(strain, rotations=rotations, strain=True)
-    energy = np.tile(energy, len(rotations))
-    volume = np.tile(volume, len(rotations))
+    e_per_v = np.tile(energy / volume, len(rotations))
     # Create symmetric tensor for elastic tensor
     strain = 0.5 * np.einsum("ni,nj->nij", *2 * [strain_voigt])
     # Remove lower triangle
@@ -153,9 +152,8 @@ def _fit_coeffs_with_energies(
         strain = np.concatenate((strain, higher_strains), axis=-1)
     if fit_first_order:
         strain = np.concatenate((strain, strain_voigt), axis=-1)
-    strain = np.einsum("n,ni->ni", volume, strain)
-    reg = LinearRegression().fit(strain, energy)
-    score = reg.score(strain, energy)
+    reg = LinearRegression().fit(strain, e_per_v)
+    score = reg.score(strain, e_per_v)
     # Create base tensor for elastic tensor
     coeff = np.zeros((6, 6))
     # Multiply upper triangle with upper triangle coeffs (v.s.)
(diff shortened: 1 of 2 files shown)


pmrv:
As discussed in #246 I've tried all combinations of running the job for an HCP cell.  Results are below first as matrices then as bar plots.

Generally orthorhombic structures seem to work best, but notice in the bar plot the inconsistency between c31 and c32 w/ pressure and c44 and c55 w/o pressure, which are supposed to be equal in HCP.

cij_bar
cij_matshow

samwaseda:
> As discussed in #246 I've tried all combinations of running the job for an HCP cell. Results are below first as matrices then as bar plots.
> 
> Generally orthorhombic structures seem to work best, but notice in the bar plot the inconsistency between c31 and c32 w/ pressure and c44 and c55 w/o pressure, which are supposed to be equal in HCP.

For primitive I have frankly no idea why it's not working. Otherwise is it possible that the lattice parameter is not relaxed beforehand?

pmrv:
> > As discussed in #246 I've tried all combinations of running the job for an HCP cell. Results are below first as matrices then as bar plots.
> > Generally orthorhombic structures seem to work best, but notice in the bar plot the inconsistency between c31 and c32 w/ pressure and c44 and c55 w/o pressure, which are supposed to be equal in HCP.
> 
> For primitive I have frankly no idea why it's not working. Otherwise is it possible that the lattice parameter is not relaxed beforehand?

I relaxed all the structures with `calc_minimize(pressure=[0]*6, ionic_force_tolerance=1e-6)` beforehand.

samwaseda:
> I relaxed all the structures with `calc_minimize(pressure=[0]*6, ionic_force_tolerance=1e-6)` beforehand.

hm ok then I'm going to include a symmetrisation step

samwaseda:
I realised that it's related to the fact that the point group symmetries that spglib do not help us symmetrize those values, so that for now it can be only avoided by increasing the number of points. In theory I guess it's possible to solve this problem by reducing the structure to the smallest unit and get all the symmetries there, but I'm not really sure how to do it properly. One way or other, this is more a fundamental problem, which is not directly related to this PR, so I'd suggest to merge this one and open another one as soon as we find a solution.

Leimeroth (approved):

pmrv:
> I realised that it's related to the fact that the point group symmetries that spglib do not help us symmetrize those values, so that for now it can be only avoided by increasing the number of points. In theory I guess it's possible to solve this problem by reducing the structure to the smallest unit and get all the symmetries there, but I'm not really sure how to do it properly. One way or other, this is more a fundamental problem, which is not directly related to this PR, so I'd suggest to merge this one and open another one as soon as we find a solution.

I think both spglib and pymatgen have methods to get reduced unit cells from super cells, but I haven't looked into how they work.  If increasing the number of points works, should we just bump the default value for that?

PR #541, opened 2022-02-18
Adjust n_print to be at most max_iter in calc_minimize

pmrv:
Changing max_iter to be less than n_print causes lammps to only print out the initial structure+forces, which is very confusing if you don't expect it.

### DIFF
--- a/pyiron_atomistics/lammps/control.py
+++ b/pyiron_atomistics/lammps/control.py
@@ -231,6 +231,9 @@ class LammpsControl(GenericParameters):
         # defaults stay consistent!
 
         max_evaluations = 100 * max_iter
+        if n_print > max_iter:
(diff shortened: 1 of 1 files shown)


pyiron_atomistics/lammps/control.py
```diff
@@ -231,6 +231,9 @@ def calc_minimize(
         # defaults stay consistent!
 
         max_evaluations = 100 * max_iter
+        if n_print > max_iter:
+            warnings.warn('n_print larger than max_iter, adjusting to n_print=max_iter')
```
niklassiemer: I would `from pyiron_base import state` and then 
```suggestion
            state.logger.warning('n_print larger than max_iter, adjusting to n_print=max_iter')
```

niklassiemer (commented):
Except for actually logging the warning this LGTM!

PR #542, opened 2022-02-18
Quasi newton

samwaseda:
This class essentially does the same as SxExtOpt. Major differences:

- It has more Hessian update schemes, esp. the ones that allow saddle point calculations.
- It's our own QN code, meaning we can implement more detailed tools when needed
- The Hessian matrix can be given beforehand, so that we can pre-determine a Hessian matrix with LAMMPS and then launch a DFT calculation (which, hopefully, will then require only a few steps).


samwaseda:
I guess now it's more or less ready. @niklassiemer How do I run Black?

pmrv:
Do I need to do anything in particular to obtain a saddle-point configuration or can I just run this with a configuration that is nearby?

niklassiemer:
> I guess now it's more or less ready. @niklassiemer How do I run Black?

just add the label (then it is run once, if you want to re-run remove and add the label)

samwaseda:
> Do I need to do anything in particular to obtain a saddle-point configuration or can I just run this with a configuration that is nearby?

Oh good point, I should have mentioned it in the DocString. To answer your question, there's nothing you need to do; just start the calculation within the harmonic distance of the saddle point. But if you already know the diffusion direction and the atom ID, you can also set them in the input. Anyway I'll modify the DocString to make this point clear.

pmrv:
If I see this right this only works with interactive jobs, right?  If I set a non-interactive reference job, I get a warning from this job and warnings from generic pyiron that lots of jobs are loaded instead of run.  Is there any fundamental reason (except efficiency) that interactive jobs are required?  In any case if non-interactive is not supported it should either throw an error or automatically enable it and show a warning that it did so.  

(I am asking this because our interactive lammps does not support MTPs yet [as it is the conda version and not our cluster build one])

pyiron_atomistics/interactive/quasi_newton.py
```diff
@@ -0,0 +1,398 @@
...
+
+    def to_hdf(self, hdf=None, group_name=None):
+        super().to_hdf(hdf=hdf, group_name=group_name)
+
+    to_hdf.__doc__ = InteractiveWrapper.to_hdf.__doc__
+
+    def from_hdf(self, hdf=None, group_name=None):
+        super().from_hdf(hdf=hdf, group_name=group_name)
+
+    from_hdf.__doc__ = InteractiveWrapper.from_hdf.__doc__
```
pmrv:
```suggestion
```
Leaving it out makes no difference, if you just call super anyway.

samwaseda: Ah you're right good job :D

pmrv:
I also find it a bit weird that I have to get the results from the initial reference job.  IS that how we generally do things with interactive wrappers?  For normal master jobs reference jobs are basically just there to specify/setup input and all results are accessible from the master job itself.

pyiron_atomistics/interactive/quasi_newton.py
samwaseda: `write_input` must be there though (which makes no sense to me but that how it works)

samwaseda:
> If I see this right this only works with interactive jobs, right?

I corrected, now it works also with static jobs.

samwaseda:
> I also find it a bit weird that I have to get the results from the initial reference job.

What kind of results do you mean?

pmrv:
> > I also find it a bit weird that I have to get the results from the initial reference job.
> 
> What kind of results do you mean?

I mean the final structure, energy, etc.  When I tested this earlier I had to get this of the reference job, is that not how it's supposed to be used?

samwaseda:
> I mean the final structure, energy, etc. When I tested this earlier I had to get this of the reference job, is that not how it's supposed to be used?

In principle `job.output` is available for the master job as well. For a non-interactive calculation it's true that we should store the data better.

pmrv:
> > If I see this right this only works with interactive jobs, right?
> 
> I corrected, now it works also with static jobs.

Still does not seem to work.  Now it does create new jobs of the same name, but in the output of the quasi-newton job I only have one structure/energy/etc.

samwaseda:
> Still does not seem to work. Now it does create new jobs of the same name, but in the output of the quasi-newton job I only have one structure/energy/etc.

It's not that it doesn't work, it was meant to work this way. It does exactly the same as SxExtOpt. If you want to change it, we have to do it in a different PR.

pmrv:
> > Still does not seem to work. Now it does create new jobs of the same name, but in the output of the quasi-newton job I only have one structure/energy/etc.
> 
> It's not that it doesn't work, it was meant to work this way. It does exactly the same as SxExtOpt. If you want to change it, we have to do it in a different PR.

Ah, this is then the final optimized structure.  I still find it a bit weird that I have to do `get_structure()` on the reference and not on the wrapper, but I can work with that.

Another question: can I get the eigenvectors of the Hessian used in the last step before convergence somehow? (To get an approximation to the diffusion path, e.g.)

samwaseda:
> Another question: can I get the eigenvectors of the Hessian used in the last step before convergence somehow? (To get an approximation to the diffusion path, e.g.)

Now it's there (`job['output/hessian']`), but I realised that `pr.load` doesn't work with `DataContainer` for the input, because apparently the child jobs are also stored in the hdf.

samwaseda:
I merge this one now because nothing depends on this one and I'd like to start working on [this one](https://github.com/pyiron/pyiron_atomistics/pull/468). If there are modifications to do, feel free to open more PR.

PR #552, opened 2022-03-01
Handling empty indices

sudarsan-surendralal:
This is necessary in case the indices are not available for trajectories.


tnecnivkcots:
Hi @sudarsan-surendralal,

I do not know, what the nb.broadcast_to method does, but I would like to make the suggestion, to put the variable ìndices` in the if statement, to check is this is None. Then we can shift the assignment from before the if statement into an else statement.

That would look like this, with one assignment less:
```
# If the indices don't exist, we construct the indices based on the indices of the initial structure
if indices is None:
    self._indices = np.broadcast_to(
        self._structure.indices,
        (len(self._positions), len(self._structure.indices)),
    )
else:
    self._indices = indices
```

I hope I can contribute something to a good readable and understandable code and want to thank you for this PR.

PR #556, opened 2022-03-06
Vasp Outcar: elegant split of electronic and ionic steps

jan-janssen:


sudarsan-surendralal (approved):
LGTM! In general, there is scope for improving these parsers (making them more elegant/faster)

PR #558, opened 2022-03-08
dont write kpoints file

Leimeroth:
Bcause it is possible to use KSPACING in INCAR and I have to calculate some data in a consistent manner with that setting so I need an option to not write the KPOINTS file.

This is an ugly workaround so I am open to other suggestions. Also it does not get stored in hdf, but the hdf storage of GenericParameters looks mysterious to me and I didn't want to put too much effort into the ugly workaround.

### DIFF
--- a/pyiron_atomistics/vasp/base.py
+++ b/pyiron_atomistics/vasp/base.py
@@ -2442,7 +2442,7 @@ class Kpoints(GenericParameters):
         )
         self._path_name = None
         self._n_path = None
-        
+
     def write_file(self, file_name, cwd=None):
         if self.file_name == False:
             return
--- a/pyiron_atomistics/vasp/base.py
+++ b/pyiron_atomistics/vasp/base.py
@@ -1856,7 +1856,10 @@ class Input:
             directory (str): The working directory for the VASP run
         """
         self.incar.write_file(file_name="INCAR", cwd=directory)
-        self.kpoints.write_file(file_name="KPOINTS", cwd=directory)
+        if "KSPACING" in self.incar.keys():
+            warnings.warn("'KSPACING' found in INCAR, no KPOINTS file written")
+        else:
+            self.kpoints.write_file(file_name="KPOINTS", cwd=directory)
         self.potcar.potcar_set_structure(structure, modified_elements)
         self.potcar.write_file(file_name="POTCAR", cwd=directory)
         # Write the species info in the POSCAR file only if there are no user defined species
@@ -2442,11 +2445,6 @@ class Kpoints(GenericParameters):
         )
         self._path_name = None
         self._n_path = None
-
-    def write_file(self, file_name, cwd=None):
-        if self.file_name == False:
-            return
-        return super().write_file(file_name, cwd)
 
     def set_kpoints_file(
         self, method=None, size_of_mesh=None, shift=None, n_path=None, path=None
--- a/pyiron_atomistics/vasp/base.py
+++ b/pyiron_atomistics/vasp/base.py
@@ -2442,6 +2442,11 @@ class Kpoints(GenericParameters):
         )
         self._path_name = None
         self._n_path = None
+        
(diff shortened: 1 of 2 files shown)


pmrv:
`Vasp.set_kpoints()` allows for `k_mesh_spacing`, but we calculate the k-points on our own in this case.  Is it possible for you to use this or do you want to use strict consistency with calculations performed outside pyiron?  If so does it make sense to drop our calculation and just use `KSPACING` from `set_kpoints`, @sudarsan-surendralal ?

Leimeroth:
>  Is it possible for you to use this or do you want to use strict consistency with calculations performed outside pyiron?
I need to keep calculations strictly consistent.

I think vasp calculates it differently from pyiron, so I think keeping the k-point calculation on the pyiron side is the better option to compare between different codes.

sudarsan-surendralal:
> > Is it possible for you to use this or do you want to use strict consistency with calculations performed outside pyiron?
> > I need to keep calculations strictly consistent.
> 
> I think vasp calculates it differently from pyiron, so I think keeping the k-point calculation on the pyiron side is the better option to compare between different codes.

Yes, you are right [(see here)](https://www.vasp.at/wiki/index.php/KSPACING) they are not 100% equivalent. Although the "pyironic" way is to do what @pmrv suggested, this won't be consistent with your other calculations. So I guess we have to use a workaround like you implemented.

pyiron_atomistics/vasp/base.py
```diff
@@ -2443,6 +2443,11 @@ def __init__(self, input_file_name=None, table_name="kpoints"):
         self._path_name = None
         self._n_path = None
 
+    def write_file(self, file_name, cwd=None):
+        if self.file_name == False:
```
sudarsan-surendralal: To maintain type consistency, I'd avoid assigning boolean values to `file_name` but rather set it to `None`

```suggestion
        if self.file_name is None:
```

sudarsan-surendralal (changes requested):
It would be nice to write a simple unittest for jobs like this to check it it works [(some examples here)](https://github.com/pyiron/pyiron_atomistics/blob/master/tests/vasp/test_vasp.py). Additionally, we should warn users when 'KSPACING' is there in the INCAR file but a KPOINTS file is also present

pyiron_atomistics/vasp/base.py
Leimeroth: The problem is that None is the default value right now, so using that would always lead to the file not being written. I could also change the default to "KPOINTS" maybe.

Leimeroth: > It would be nice to write a simple unittest for jobs like this to check it it works [(some examples here)](https://github.com/pyiron/pyiron_atomistics/blob/master/tests/vasp/test_vasp.py). Additionally, we should warn users when 'KSPACING' is there in the INCAR file but a KPOINTS file is also present

Alternatively to what I did it would also be possible to only write the KPOINTS file if "KSPACING" is not present in the INCAR and issue a warning, instead of fiddling around with the KPOINTS class itself

pmrv: I would suggest to make a change around [here](https://github.com/pyiron/pyiron_atomistics/blob/907991b31ba4543ffb90a33d957108181e520506/pyiron_atomistics/vasp/base.py#L1859) and issue a warning if no KPOINTS is written.  In the current solution its not at all clear how this forces the file not being written and will probably be forgotten next week.

pmrv (approved):
lgtm.

EDIT: I defer to @sudarsan-surendralal on the unit tests issue, not sure how and where to do that best, but I guess making a job with `run_mode='manual'` and then checking input files written should be possible.

sudarsan-surendralal (approved):
LGTM! I added a unit test for this and it works.

Leimeroth:
> LGTM! I added a unit test for this and it works.

Thanks

PR #561, opened 2022-03-15
Modify where number of bond types is chosen from

raynol-dsouza:
As discussed in the pyiron meeting yesterday (14/03/2022), I changed the variable from which the number of bond types is chosen to be written to the `structure.inp` file.

### DIFF
--- a/tests/lammps/test_base.py
+++ b/tests/lammps/test_base.py
@@ -697,16 +697,12 @@ class TestLammps(TestWithCleanProject):
         self.job.structure.bonds = [[1, 2, 1], [1, 3, 2]]
         self.job.potential = potential
         self.job.calc_static()
-        file_directory = os.path.join(
-            self.execution_path, "..", "static", "lammps_test_files"
-        )
-        self.job.restart_file_list.append(
-            os.path.join(file_directory, "dump.out")
-        )
-        self.job.restart_file_list.append(
-            os.path.join(file_directory, "log.lammps")
-        )
         self.job.run(run_mode="manual")
-        self.job.status.collect = True
-        self.job.run()
+
+        bond_str = "2 bond types\n"
+        struct_file = open(self.job.get_workdir_file('structure.inp'), 'r')
+        struct_lines = struct_file.readlines()[4]
+        struct_file.close()
+        self.assertTrue(struct_lines[-1], bond_str)
+
 
--- a/tests/lammps/test_base.py
+++ b/tests/lammps/test_base.py
@@ -697,5 +697,16 @@ class TestLammps(TestWithCleanProject):
         self.job.structure.bonds = [[1, 2, 1], [1, 3, 2]]
         self.job.potential = potential
         self.job.calc_static()
+        file_directory = os.path.join(
+            self.execution_path, "..", "static", "lammps_test_files"
+        )
+        self.job.restart_file_list.append(
+            os.path.join(file_directory, "dump.out")
+        )
+        self.job.restart_file_list.append(
+            os.path.join(file_directory, "log.lammps")
+        )
+        self.job.run(run_mode="manual")
+        self.job.status.collect = True
         self.job.run()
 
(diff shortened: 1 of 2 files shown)


raynol-dsouza:
The change seems to be breaking multiple workflows. I will re-open the PR once I have a better idea of where else this bit is used.

raynol-dsouza:
I just double checked that the notebooks that fail in the tests and my local machine are `tests_sphinx_sphinx_check_all.ipynb` and `sqs.ipynb`. My changes shouldn't be affecting these notebooks at all...

EDIT: I reverted the changes and confirmed that the change does not break the above mentioned tests.

sudarsan-surendralal (commented):
Yeah, the changes you made shouldn't break those notebooks and it has to be something else. But coming back to the bonds issue,  it'll be great if you can add the short example you showed on Monday as a unittest just to ensure that any changes in the future doesn't break your jobs. As an example see:
 https://github.com/pyiron/pyiron_atomistics/blob/54cc6e667bb83fd44a884230eea912e38296d88f/tests/lammps/test_base.py#L172

raynol-dsouza:
Thanks for the suggestion! I’ll add a test for this.

tests/lammps/test_base.py
```diff
@@ -678,3 +678,35 @@ def test_units(self):
...
+        self.job.calc_static()
+        file_directory = os.path.join(
+            self.execution_path, "..", "static", "lammps_test_files"
+        )
+        self.job.restart_file_list.append(
+            os.path.join(file_directory, "dump.out")
+        )
+        self.job.restart_file_list.append(
+            os.path.join(file_directory, "log.lammps")
+        )
```
sudarsan-surendralal: Do we really need these lines? This isn't the output of the calculations with the bonds anyway. Wouldn't it be sufficient if the we just check if the input is written correctly?

tests/lammps/test_base.py
```diff
@@ -678,3 +678,35 @@ def test_units(self):
...
+        self.job.run(run_mode="manual")
+        self.job.status.collect = True
+        self.job.run()
```
sudarsan-surendralal: Again, not sure if output parsing is necessary here. After `self.job.run(run_mode="manual")`, you could inspect the `structure.inp` file to check if the bonds are written the way you really want them to be.

sudarsan-surendralal (changes requested):
Thanks for the tests. I think a couple of changes would still be necessary

tests/lammps/test_base.py
raynol-dsouza: I tired it without these lines and the tests ended up failing. Probably because I was running the job. I will take a look at this again.

tests/lammps/test_base.py
raynol-dsouza: What exactly does `run_mode=manual` do? Just write the input files or actually run the job?

sudarsan-surendralal: It just writes the input files and does not run them.

tests/lammps/test_base.py
sudarsan-surendralal: If you just use `run_mode=manual` it should be fine

tests/lammps/test_base.py
raynol-dsouza: Cool. I will get it fixed right away!

tests/lammps/test_base.py
```diff
@@ -678,3 +678,31 @@ def test_units(self):
...
+        self.job.structure.bonds = [[1, 2, 1], [1, 3, 2]]
+        self.job.potential = potential
+        self.job.calc_static()
+        self.job.run(run_mode="manual")
+
+        bond_str = "2 bond types\n"
+        struct_file = open(self.job.get_workdir_file('structure.inp'), 'r')
+        struct_lines = struct_file.readlines()[4]
+        struct_file.close()
+        self.assertTrue(struct_lines[-1], bond_str)
```
sudarsan-surendralal: This can be simplified 

```suggestion
        self.assertTrue(self.job["structure.inp"][4][-1], bond_str)
```

sudarsan-surendralal (approved):
LGTM :)! Just a minor simplification recommendation

tests/lammps/test_base.py
raynol-dsouza: Awesome! I didn't know I could access the input files this way! :D

PR #572, opened 2022-03-24
Rename unit test workflow

pmrv:
I found it annoying that the workflow for unit tests is labeled "Python Package", so I changed it.


jan-janssen (approved):
Looks good to me but maybe also update the other pyiron packages

PR #576, opened 2022-03-30
Symmetrize vectors

samwaseda:
`symmetrize_vector` can now symmetrize several vectors at the same time


samwaseda:
Since it's a relatively minor change I'll merge it as soon as the tests pass

PR #594, opened 2022-04-21
A function to plot free energy vs volume curves for different temperatures

usaikia:


pyiron_atomistics/atomistics/master/quasi.py
```diff
@@ -173,3 +173,60 @@ def optimise_volume(self, bulk_eng):
...
+                "QHA Job must be successfully run, before calling this method."
+            )
+
+        if not job_murn.status.finished:
+            raise ValueError(
+                "Murnaghan Job must be successfully run, before calling this method."
+            )
+
+        import matplotlib
+        cmap = matplotlib.cm.get_cmap('coolwarm')
```
pmrv: Does it make sense to use a binary color map by default?  I'd rather suggest 'magma' or 'viridis'.  Best would probably be an extra argument to change the color map.

pyiron_atomistics/atomistics/master/quasi.py
```diff
@@ -173,3 +173,60 @@ def optimise_volume(self, bulk_eng):
...
+        Plot volume vs free energy curves for defined temperatures.
+
+        Args:
+            murnaghan_job: job_name or job_id of the Murnaghan job
+            temperature_start: if None, value will be used from job.input['temperature_start']
+            temperature_end: if None, value will be used from job.input['temperature_end']
+            temperature_steps: if None, value will be used from job.input['temperature_steps']
+
+        Returns:
+            Volume vs free energy curves for defined temperatures.
```
pmrv: The method is not actually returning anything so this should be blank.  It might make sense to return the current axis and allow passing one in, as in e.g. the [phonopy](https://github.com/pyiron/pyiron_atomistics/blob/4ab5279c2096ae0859aaf55f8eb20f35b5a0cf13/pyiron_atomistics/atomistics/master/phonopy.py#L499).

pmrv (approved):
Generally looks good to me. 

This is a discussion that goes beyond hits PR, but: can we couple this murnaghan job that is required more tightly to the QHA.  Right now I'd have to set these up indepently and make sure the inputs are consistent.  How are people actually running these?

pyiron_atomistics/atomistics/master/quasi.py
```diff
@@ -173,3 +173,60 @@ def optimise_volume(self, bulk_eng):
...
+            raise ValueError(
+                "QHA Job must be successfully run, before calling this method."
+            )
+
+        if not job_murn.status.finished:
+            raise ValueError(
+                "Murnaghan Job must be successfully run, before calling this method."
+            )
+
+        import matplotlib
```
jan-janssen: Should we move the import to the top of the class?

pyiron_atomistics/atomistics/master/quasi.py
usaikia: Added an  extra argument to change the color map.

pyiron_atomistics/atomistics/master/quasi.py
usaikia: current axis returned.

pyiron_atomistics/atomistics/master/quasi.py
usaikia: Moved import to the top of the class. Also, murnaghan_job option is now set to None. If no Murnaghan job is assigned by the user, it will plot free energy without total electronic energy at T=0.

pyiron_atomistics/atomistics/master/quasi.py
```diff
@@ -173,3 +174,81 @@ def optimise_volume(self, bulk_eng):
...
+            axis (matplotlib axis, optional): plot to this axis, if not given a new one is created.
+            *args: passed through to matplotlib.pyplot.plot when plotting free energies.
+            **kwargs: passed through to matplotlib.pyplot.plot when plotting free energies.
+
+        Returns:
+            matplib axis: the axis the figure has been drawn to, if axis is given the same object is returned.
+        """
+        energy_zero = 0
+        if murnaghan_job != None:
+            job_murn = project.load(murnaghan_job)
```
pmrv:
```suggestion
            job_murn = self.project.load(murnaghan_job)
```

pmrv (approved):
Looks good to me expect for the small typo.  Go ahead and merge this, if you are happy with it.

usaikia:
Thanks @pmrv . I will merge this pull request.

PR #603, opened 2022-05-03
Calphy implementation in pyiron

srmnitc:
This pull request implements calphy wrapper for pyiron. The calphy wrapper is implemented in `pyiron_atomistics/calphy`.

Calphy wrapper adds the possibility for fast free energy calculations for interatomic potentials.


srmnitc:
@pmrv I have an issue in which I would like some help:
When loading a finished job, I need to explicitely call `from_hdf`. See [here](https://github.com/srmnitc/pyiron_atomistics/blob/master/tests/calphy/test_base.py#L74). Could you maybe help me with this issue?

pmrv:
> @pmrv I have an issue in which I would like some help: When loading a finished job, I need to explicitely call `from_hdf`. See [here](https://github.com/srmnitc/pyiron_atomistics/blob/master/tests/calphy/test_base.py#L74). Could you maybe help me with this issue?

Sure, you're not supposed to instantiate a job like this directly.  So instead of
```python
cls.output_job = Calphy(
    project=ProjectHDFio(project=cls.output_project, file_name="tm_fcc"),
    job_name="tm_fcc",
)
```
you want
```python
cls.output_job = cls.output_project.load("tm_fcc")
```
If that doesn't work the project is setup wrongly.

pmrv:
Ah, I see you try to load some prre-run job.  What you probably want to do is more like [this](https://github.com/pyiron/pyiron_atomistics/blob/ce840e5434de3c0e78d575be10b4e37f9abf6826/tests/atomistics/job/test_transform_trajectory.py#L54).

Prepare the job in a project on the cluster.  `Project.pack` them stick that archive in `tests/static` and then `Project.unpack` them in the `setUp` method of your tests.

pyiron_atomistics/calphy/calphy.py
```diff
@@ -0,0 +1,34 @@
...
+        project (pyiron_atomistics.project.Project instance):  Specifies the project path among other attributes
+        job_name (str): Name of the job
+
+    Attributes:
+        input (calphy.Input instance): Instance which handles the input
+    """
+
+    def __init__(self, project, job_name):
+        super(Calphy, self).__init__(project, job_name)
+        self.__name__ = "Calphy"
```
pmrv: `GenericJob` does this automatically now.
```suggestion
```

pyiron_atomistics/calphy/calphy.py
```diff
@@ -0,0 +1,34 @@
...
+    "- Computational Materials Design (CM) Department"
+)
+__version__ = "1.0"
+__maintainer__ = "Sarath Menon"
+__email__ = "s.menon@mpie.de"
+__status__ = "production"
+__date__ = "Feb 14, 2022"
+
+
+class Calphy(CalphyBase):
```
pmrv: I know Lammps and Vasp have this pattern in our code, but imo it's bad form.  This class doesn't implement anything on top of `CalphyBase` and (I guess) we do not plan on having multiple classes derive from it, so best to keep everything in one spot.

tests/calphy/test_base.py
```diff
@@ -0,0 +1,81 @@
...
+            project=ProjectHDFio(project=cls.project, file_name="test_calphy"),
+            job_name="test_calphy",
+        )
+        filepath = os.path.join(os.path.dirname(os.path.abspath(__file__)), "../static/")
+        cls.output_project = Project(os.path.join(filepath, "calphy_test_files"))
+        cls.output_job = Calphy(
+            project=ProjectHDFio(project=cls.output_project, file_name="tm_fcc"),
+            job_name="tm_fcc",
+        )
+        #cls.job = cls.project.create_job("Calphy", "job_calphy")
```
pmrv: See my top-level comments regarding `pack`/`unpack`.

tests/calphy/test_base.py
```diff
@@ -0,0 +1,81 @@
...
+
+    @classmethod
+    def tearDownClass(cls):
+        cls.execution_path = os.path.dirname(os.path.abspath(__file__))
+        project = Project(os.path.join(cls.execution_path, "test_calphy"))
+        project.remove_jobs_silently(recursive=True)
+        project.remove(enable=True)
+        state.update()
```
pmrv:
```suggestion
    @classmethod
    def tearDownClass(cls):
        cls.execution_path = os.path.dirname(os.path.abspath(__file__))
        project = Project(os.path.join(cls.execution_path, "test_calphy"))
        project.remove_jobs_silently(recursive=True)
        project.remove(enable=True)
        state.update()
```
You'll want to derive from `pyiron_base._tests.TestWithProject`.

tests/calphy/test_base.py
```diff
@@ -0,0 +1,81 @@
...
+    
+    def test_potentials(self):
```
pmrv, line 50: Just a general comment regarding unit tests: I personally strongly prefer short docstrings and passing `msg=...` to `self.assert...`.  No need to go overboard with this, but if I come back in half a year I should be able to understand the rationale for the test and what invariant it is testing without reading all the involved code.

pyiron_atomistics/calphy/base.py
```diff
@@ -0,0 +1,602 @@
...
+    job.calc_free_energy(temperature=[1100, 1400], pressure=None, reference_phase="solid")
+    ```
+    
+    In addition the boolean option :attr:`input.npt` can be used to determine the MD ensemble. If True, temperature integration and alchemy/upsampling are carried out in the NPT ensemble. If False, the NVT ensemble is employed.
+    
+    After the calculation is over, the various output options can be accessed through `job.output`.
+    """
+    def __init__(self, project, job_name):
+        super(CalphyBase, self).__init__(project, job_name)
+        self.__name__ = "CalphyJob"
```
pmrv:
```suggestion
        self.__name__ = "CalphyJob"
```
No need, done by `GenericJob` automatically.

pyiron_atomistics/calphy/base.py
```diff
@@ -0,0 +1,602 @@
...
+        
+        Returns:
+            None
+        """
+        if self._potential_initial is not None:
+            self._potential_initial.copy_pot_files(self.working_directory)
+        if self._potential_final is not None:
+            self._potential_final.copy_pot_files(self.working_directory)
+
+    def prepare_pair_styles(self) -> tuple[list, list]:
```
pmrv: If this not something (Notebook-) users are expected to call, rename with underscore -> `_prepare_pair_styles`.

pyiron_atomistics/calphy/base.py
```diff
@@ -0,0 +1,602 @@
...
+            
+    def get_potentials(self) -> list:
+        """
+        Return the interatomic potentials
+        
+        Args:
+            None
+            
+        Returns:
+            list: list of interatomic potentials
```
pmrv:
```suggestion
            list of str: list of interatomic potentials
```

pyiron_atomistics/calphy/base.py
```diff
@@ -0,0 +1,602 @@
...
+            self.input.potential_initial_name = potential_filenames[0]
+        if len(potential_filenames) > 1:
+            potential = LammpsPotentialFile().find_by_name(potential_filenames[1])
+            self._potential_final = LammpsPotential()
+            self._potential_final.df = potential
+            self.input.potential_final_name = potential_filenames[0]
+        if len(potential_filenames) > 2:
+            raise ValueError("Maximum two potentials can be provided")
+            
+    def get_potentials(self) -> list:
```
pmrv:
```suggestion
    def get_potentials(self) -> List[str]:
```

pyiron_atomistics/calphy/base.py
```diff
@@ -0,0 +1,602 @@
+import copy
+import os
+import numpy as np
+from typing import Union
```
pmrv:
```suggestion
from typing import Union, List
```

pyiron_atomistics/calphy/base.py
```diff
@@ -0,0 +1,602 @@
...
+            filenames.append(self.input.potential_final_name)
+        self.set_potentials(filenames)
+            
+    @property
+    def potential(self):
+        return self.get_potentials()
+    
+    @potential.setter
+    def potential(self, potential_filenames):
+        self.set_potentials(potential_filenames)
```
pmrv: I guess this is in analogy to `AtomisticGenericJob`, but I think it's a bit of a false friend. I'd either rename in plural or make it operate explicitly only on the initial potential.  Otherwise users can be in a situation where they (by above analogy) expect a string back, but get a list.

pyiron_atomistics/calphy/base.py
```diff
@@ -0,0 +1,602 @@
...
+    
+    def write_input(self):
+        """
+        Write input for calphy calculation
+        
+        Args:
+            None
+        
+        Returns:
+            None
```
pmrv:
```suggestion
        
        Args:
            None
        
        Returns:
            None
```

pmrv (changes requested):
Looks generally pretty good, but I left some nits.  Especially for typing I like it to be as explicit as possible, e.g. using `List[str]` instead of just `list`. 

Only bigger concern is the naming of the `calc_*` methods.  Are all of them intended as entry points for users? If so, everythings good, if not I can see there being a bit of confusion between `calc_mod_*` and `calc_free_energy`, so a bit more explanation would be nice.

srmnitc:
> Looks generally pretty good, but I left some nits. Especially for typing I like it to be as explicit as possible, e.g. using `List[str]` instead of just `list`.
> 
> Only bigger concern is the naming of the `calc_*` methods. Are all of them intended as entry points for users? If so, everythings good, if not I can see there being a bit of confusion between `calc_mod_*` and `calc_free_energy`, so a bit more explanation would be nice.

I updated the docstrings to clarify the methods.

pyiron_atomistics/calphy/calphy.py
srmnitc: I am not sure I follow fully the changes. In principle, I could remove `calphy.py` altogether. Would this be the best idea?

pmrv: Yes, that's want I recommend.  Remove `Calphy` and rename `CalphyBase` to `Calphy`.  I'd probably then also rename the file from `base.py` to `job.py` or something similar, because words like `base`, `generic`, etc. have very little information content.

tests/calphy/test_base.py
srmnitc: changed tests

pyiron_atomistics/calphy/base.py
srmnitc: Done.

pyiron_atomistics/calphy/calphy.py
srmnitc: I renamed the files and removed base

pyiron_atomistics/calphy/base.py
srmnitc: I would like to keep the property as `potential`. The logic behind this is that if you set two potentials, the mode is `alchemy` automatically. I updated the property to return a string if only one potential was set. That should remove the confusion.

pyiron_atomistics/calphy/job.py
```diff
@@ -1,7 +1,7 @@
 import copy
```
pmrv:
```suggestion
```
Flagged by codacy.

pyiron_atomistics/calphy/job.py
```diff
@@ -120,7 +128,6 @@ class CalphyBase(GenericJob):
     """
     def __init__(self, project, job_name):
         super(CalphyBase, self).__init__(project, job_name)
```
pmrv:
```suggestion
        super().__init__(project, job_name)
```

tests/calphy/test_base.py
```diff
@@ -6,6 +6,7 @@
 import numpy as np
 import unittest
 import warnings
```
pmrv:
```suggestion
import unittest
```
Also codacy.

tests/calphy/test_base.py
```diff
@@ -71,11 +67,15 @@ def test_modes(self):
...
     
     def test_output(self):
-        self.output_job.from_hdf()
-        self.assertEqual(float(self.output_job.output.spring_constant), 1.51)    
-        self.assertEqual(self.output_job.output.energy_free[0], -4.002465158959863)
-        self.assertEqual(int(self.output_job.output.temperature[0]), 1100)
-        self.assertEqual(int(self.output_job.output.temperature[-1]), 1400)
+        filepath = os.path.join(os.path.dirname(os.path.abspath(__file__)), "../static/")
+        shutil.copy(os.path.join(filepath, "calphy_test_files/tm_fcc.tar.gz"), cwd)
+        shutil.copy(os.path.join(filepath, "calphy_test_files/export.csv"), cwd)
```
pmrv: `cwd` is not defined, probably you want something like `os.path.basename(__FILE__)`.

pmrv (approved):
I highlighted some typos and such from codacy, but otherwise looks good to me!

I'm surprised that the tests appear to run, because there's one undefined variable, please double check locally that everything is good and have a look at the coverage before merging.

tests/calphy/test_base.py
srmnitc: changed to `os.getcwd()`

srmnitc:
> I highlighted some typos and such from codacy, but otherwise looks good to me!
> 
> I'm surprised that the tests appear to run, because there's one undefined variable, please double check locally that everything is good and have a look at the coverage before merging.

I am also surprised about this. However, I have now checked the tests locally too and everything seems fine. I will go ahead and merge then. Thank you for your review and comments!

pmrv:
> > I highlighted some typos and such from codacy, but otherwise looks good to me!
> > I'm surprised that the tests appear to run, because there's one undefined variable, please double check locally that everything is good and have a look at the coverage before merging.
> 
> I am also surprised about this. However, I have now checked the tests locally too and everything seems fine. I will go ahead and merge then. Thank you for your review and comments!

Ah, I found the issue.  You'll need to place an empty `__init__.py` into `tests/calphy`, otherwise `unittests` will not recognize it as a module and won't run the tests.

You can force it to run them by doing
`python -m unittest tests/calphy/test_calphy.py`.

srmnitc:
> > > I highlighted some typos and such from codacy, but otherwise looks good to me!
> > > I'm surprised that the tests appear to run, because there's one undefined variable, please double check locally that everything is good and have a look at the coverage before merging.
> > 
> > 
> > I am also surprised about this. However, I have now checked the tests locally too and everything seems fine. I will go ahead and merge then. Thank you for your review and comments!
> 
> Ah, I found the issue. You'll need to place an empty `__init__.py` into `tests/calphy`, otherwise `unittests` will not recognize it as a module and won't run the tests.
> 
> You can force it to run them by doing `python -m unittest tests/calphy/test_calphy.py`.

Ah, it does not discover the tests; I have been running the file directly locally, which is why I did not have issues. Interesting! thanks!

srmnitc:
Ready to merge, but kept on hold due to [the license discussion](https://github.com/orgs/pyiron/teams/pyiron/discussions/116?from_comment=2#discussion-116-comment-2)

srmnitc:
> Ready to merge, but kept on hold due to [the license discussion](https://github.com/orgs/pyiron/teams/pyiron/discussions/116?from_comment=2#discussion-116-comment-2)

Released Calphy API under BSD license, removing the issues that prevent this merge.

PR #610, opened 2022-05-11
Fix murnaghan status tests

sudarsan-surendralal:
The unittests related to the Murnaghan class seems to fail randomly [example](https://github.com/pyiron/pyiron_atomistics/runs/6271548845?check_suite_focus=true). This PR aims to solve this issue


jan-janssen (approved):
Thanks for taking a look at this.

PR #626, opened 2022-05-26
Add error message when Vasp.set_occupancy_smearing is wrongly called

### LINKED ISSUE #624: `set_occupancy_smearing` raises error when method not specified
```python
job_vasp.set_occupancy_smearing(width=0.15)
```

Raised error :
```python
AttributeError: 'NoneType' object has no attribute 'lower'
```

Either the user must be able to set the width even without specifying the smearing method, or the error message must be more specific for the user to understand that this smearing method must be specified.

pmrv:
Fixes #624.

@Vishwa-adi Please a have a look if that makes it more understandable.


sudarsan-surendralal (commented):
Why not change the default of `smearing` and 'width' to `Methfessel-Paxton` and `0.1` instead?

pmrv:
> Why not change the default of `smearing` and 'width' to `Methfessel-Paxton` and `0.1` instead?

I have no opinion on what the defaults should be.  I simply didn't want to change the logic of the method.  We should probably check with the VASP job and make something consistent.

PR #629, opened 2022-05-26
Strip / in SphinxBase.__getitem__

pmrv:
If not stripped job["output/generic"] and job["output/generic/"] give different output, which is very weird.


samwaseda (approved):
Looks good to me, except for my feeling that @pmrv's vocabulary should go beyond "weird" when he describes a problem 😎

PR #631, opened 2022-05-30
add a simple implement for potential files symlink.

appassionate:
pyiron handle lammps input as potential, control parts.  Some potential files in lammps will be large, maybe we can symlink potential files instead of copying pots from resources.


jan-janssen:
Hi @appassionate, 

Thank you for your contribution, being able to link to the potential files rather than copying them is definitely a valuable feature. 

The only part we have to be careful with, is the submission to the queuing system. 

Best, 

Jan

appassionate:
> Hi @appassionate,
> 
> Thank you for your contribution, being able to link to the potential files rather than copying them is definitely a valuable feature.
> 
> The only part we have to be careful with, is the submission to the queuing system.
> 
> Best,
> 
> Jan

Thanks. I think so.
local queue in pysqa seems no problem, i guess. When remote, where the "temporary" potenrial files should be located at "ssh_remote_path" will be a problem? :)

appassionate:
@jan-janssen Having learned more about pyrion, I think the idea may be naive.
Maybe pyiron need some object/abstraction which might called "HasFile" similar to  "HasStructure" , to solve the extra file import from the file system?
It might  be supposed to solve GenericJob extra file importing when dealing with write input, I guess... Also, it is obvious that remote file domain should be considered...
I would like to close this pr if possible  : )
thanks.

PR #650, opened 2022-06-09
Deprecate structure creation methods directly on Project

pmrv:


liamhuber (approved):

niklassiemer:
I do not understand how the depreciation hurts that test?! Do we have a problem with the decorator?

pmrv:
No, someone somewhere only check the number of warnings in a unit tests.  Since those change with the PR the test break.

niklassiemer:
@pmrv I took the liberty to fix it directly here.

jan-janssen (approved):
Looks good to me

pmrv:
> @pmrv I took the liberty to fix it directly here.

Thanks!

PR #657, opened 2022-06-22
add lstrip() to lines for parsing

Leimeroth:
Some package/lammps version change or similar shifts the line
Step Temp PotEng TotEng Pxx Pxy Pxz Pyy Pyz Pzz Volume
which is used as a trigger for parsing by some whitespace, causing problems.
To generally prevent this lstrip() is applied to all lines read from log.lammps

### DIFF
--- a/pyiron_atomistics/lammps/base.py
+++ b/pyiron_atomistics/lammps/base.py
@@ -607,6 +607,7 @@ class LammpsBase(AtomisticGenericJob):
         if os.path.exists(file_name):
             with open(file_name, "r") as f:
                 f = f.readlines()
+                f = [l.lstrip() for l in f]
                 l_start = np.where([line.startswith("Step") for line in f])[0]
                 l_end = np.where([line.startswith("Loop") for line in f])[0]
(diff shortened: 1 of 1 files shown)


pmrv:
I'm a bit unhappy with this whole function (not your contribution), because this reads the whole file into memory and iterates multiple times over all the lines.  I'm just quickly checking whether I can come up with a single pass solution.

Leimeroth:
It would be possible to simply iterate over the lines directly instead of using readlines by looping:
for l in f:
and only create a df whenever a trigger was found or even read in the whole str up to the end and parse everything using pandas instead of creating 1000s of dataframes.
But I think the logfile does not really matter too much compared to reading the dump file anyway concerning time and memory consumption

Leimeroth:
Untested draft: #658

pmrv:
I have definitely produced log.lammps files in the 100's of MB and the old version uses ~1GB of ram for the parse, so I think it's ok to have an eye on it.  Thanks for you other PR, I'll benchmark it in a bit.

PR #658, opened 2022-06-22
Alternative lammps parse

Leimeroth:
Another approach to parsing without readlines() and multiple loops

EDIT (@liamhuber -- correct if I'm wrong about this, Niklas): Intended to solve #671

### DIFF
--- a/pyiron_atomistics/lammps/base.py
+++ b/pyiron_atomistics/lammps/base.py
@@ -639,14 +639,13 @@ class LammpsBase(AtomisticGenericJob):
                             read_thermo = False
                             continue
                         thermo_lines += l
-                
+
                     if l.startswith("Step"):
                         read_thermo = True
                         thermo_lines += l
 
-                dfs.append(pd.read_csv(
-                        StringIO(thermo_lines), delim_whitespace=True
-                    ))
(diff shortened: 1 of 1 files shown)


pmrv:
Based on this I've found a version that is even more efficient, will push it when I have cleaned it up.  In the meantime feel free to merge #657, so you don't need to wait for me.

jan-janssen:
Using a for loop could be an issue performance wise, is it possible to solve this with arrays as @samwaseda did before?

Leimeroth:
I am not sure what @samwaseda did, but the previous implementation read in the whole file into memory and then did multiple list comprehensions and initialized seperate StringIO and DataFrame instances for each line, so I assume this is a improvement already.
Anyway @pmrv did some tests and already found a better version so this was not merged yet, but only exists as a reminder.

Leimeroth:
@pmrv should I merge this or do you want to create a PR for your version?

jan-janssen:
> @pmrv should I merge this or do you want to create a PR for your version?

Can we still do a set of benchmarks before we merge these changes? I have the feeling we are moving back and forth with the LAMMPS parser, so I would like to have clear reasoning why we choose one implementation over another.

Leimeroth:
up

pmrv:
At this point, let's just merge this and revisit with pymatgen parser at some point in the future.

PR #660, opened 2022-06-24
bugfix for non constraint

samwaseda:
SPHInX offers the possibility to not constrain some of the magnetic moments, which could not be activated before this fix, because `SparseList` does not work in the same way as array or list.


samwaseda:
tests will follow

samwaseda:
In the end I also fixed `SparseList`

samwaseda:
I don't understand what's happening with black, but since everything else is working and I urgently need this bugfix for what I'm doing right now, I merge it. Otherwise I don't think anyone would be affected.

niklassiemer:
The black formatting workflow fails if there is nothing to do. Thus, everything is fine :)

PR #661, opened 2022-06-25
Lammps velocitiy handling

Leimeroth:
Implement writing of velocities in lammps data file. Currently only works for metal units and only for lammps, but handling of velocities is completely inconsistent and buggy anyway. I tried to raise NotImplementedError in the ase inherited set and get velocities and momenta functions, but that completely breaks the pyiron atoms class, which is very bad, because right now I can call set_velocities and it will simply do nothing when the structure is written.
I didn't test yet if velocities are correctly written, especially not for skewed structures, but the code runs.

### DIFF
--- a/pyiron_atomistics/atomistics/structure/atoms.py
+++ b/pyiron_atomistics/atomistics/structure/atoms.py
@@ -89,7 +89,7 @@ class Atoms(ASEAtoms):
         elements=None,
         dimension=None,
         species=None,
-        **qwargs
+        **qwargs,
     ):
         if symbols is not None:
             if elements is None:
@@ -228,7 +228,8 @@ class Atoms(ASEAtoms):
             self._velocities = val
         else:
             raise ValueError(
-                f"Shape of velocities and positions has to match. Velocities shape: {val.shape}, positions shape: {self.positions.shape}")
+                f"Shape of velocities and positions has to match. Velocities shape: {val.shape}, positions shape: {self.positions.shape}"
+            )
 
     @property
     def spins(self):
@@ -2139,7 +2140,7 @@ class Atoms(ASEAtoms):
                 pse=self._pse,
                 index=index,
                 atoms=self,
-                **new_dict
+                **new_dict,
             )
 
         new_array = super(Atoms, self).__getitem__(item)
@@ -2645,7 +2646,7 @@ class _CrystalStructure(Atoms):
         dimension=3,
         rel_coords=True,
         pse=None,
-        **kwargs
+        **kwargs,
     ):
 
         # print "basis0"
--- a/pyiron_atomistics/atomistics/structure/atoms.py
+++ b/pyiron_atomistics/atomistics/structure/atoms.py
@@ -215,6 +215,12 @@ class Atoms(ASEAtoms):
             self.dimension = 0
         self._visualize = Visualize(self)
         self._analyse = Analyse(self)
+        # Velocities were not handled at all during file writing
+        self._velocities = None
+
+    @property
+    def velocities(self):
+        return self._velocities
 
     @property
     def spins(self):
@@ -2056,6 +2062,46 @@ class Atoms(ASEAtoms):
             if not len(set(sum_atoms.indices)) == len(sum_atoms.species):
                 raise ValueError("Adding the atom instances went wrong!")
         return self
+
+    def get_momenta(self):
+        """
+        Velocity and momenta functions are currently not supported.
+        For lammps jobs intial velocities can be set using job.structure.velocities = array
+
+        Raises:
+            NotImplementedError: 
+        """        
+        raise NotImplementedError("Currently not used on pyiron atoms")
+    
+    def set_momenta(self):
+        """
+        Velocity and momenta functions are currently not supported.
+        For lammps jobs intial velocities can be set using job.structure.velocities = array
+
+        Raises:
+            NotImplementedError: 
+        """        
+        raise NotImplementedError("Currently not used on pyiron atoms")
+
+    def get_velocities(self):
+        """
+        Velocity and momenta functions are currently not supported.
+        For lammps jobs intial velocities can be set using job.structure.velocities = array
+
+        Raises:
+            NotImplementedError: 
+        """        
+        raise NotImplementedError("Currently not used on pyiron atoms")
+
+    def set_velocities(self):
+        """
+        Velocity and momenta functions are currently not supported.
+        For lammps jobs intial velocities can be set using job.structure.velocities = array
+
+        Raises:
+            NotImplementedError: 
+        """        
+        raise NotImplementedError("Currently not used on pyiron atoms")
 
     __iadd__ = extend
 
--- a/pyiron_atomistics/atomistics/structure/atoms.py
+++ b/pyiron_atomistics/atomistics/structure/atoms.py
@@ -2063,46 +2063,6 @@ class Atoms(ASEAtoms):
                 raise ValueError("Adding the atom instances went wrong!")
         return self
 
-    def get_momenta(self):
-        """
-        Velocity and momenta functions are currently not supported.
-        For lammps jobs intial velocities can be set using job.structure.velocities = array
-
-        Raises:
-            NotImplementedError: 
-        """        
-        raise NotImplementedError("Currently not used on pyiron atoms")
-    
-    def set_momenta(self):
-        """
-        Velocity and momenta functions are currently not supported.
-        For lammps jobs intial velocities can be set using job.structure.velocities = array
-
-        Raises:
-            NotImplementedError: 
-        """        
-        raise NotImplementedError("Currently not used on pyiron atoms")
-
-    def get_velocities(self):
-        """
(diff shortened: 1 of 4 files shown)


Leimeroth:
added unit handling and tested for a primitive Cu structure (is_skewed() returns True) so I assume this works for all structures now

liamhuber:
@Leimeroth sorry for the delay -- I moved cities and lost childcare. That's been restarted as of today though, so I should have a chance to review this sometime this week!

pyiron_atomistics/atomistics/structure/atoms.py
```diff
@@ -215,6 +215,19 @@ def __init__(
...
+    @property
+    def velocities(self):
+        return self._velocities
+
+    @velocities.setter
+    def velocities(self, val):
+        if self.positions.shape == val.shape:
+            self._velocities = val
+        else:
+            raise ValueError("Shape of velocities and positions has to match")
```
liamhuber:
```suggestion
            raise ValueError(f"Shapes did not match: Velocities {val.shape} vs positions {self.positions.shape}")
```

liamhuber (commented):
So in `structure.Atoms` it's very nice. I'd just ask to please go into the unit tests and test our new cases:
- On instantiation velocities really are none
- After being set, they exist as expected
- Setting with the wrong shape throws an exception
These are absolutely trivial tests here, but they'll make sure it *keeps* working.

In the Lammps code I didn't really like the entire job being passed as an attribute to `structure.LammpsStructure`. At the end of the day we just use it for unit conversions anyhow.

Reading it, I am actually a bit concerned that we didn't run into this problem beforehand. We got away with it because we always assume metal units, where distances are the same between pyiron and lammps, and now we care because velocities differ. But there are other Lammps units!

So I'd like to kill two birds with one stone: only pass the minimal amount of data needed from `LammpsBase` down to `LammpsStructure`, and set ourselves up to expand the use of unit conversions in the future.

To accomplish this, I suggest changing the signature of `LammpsStructure` to take the unit converter with a `None` default: `def __init__(self, input_file_name=None, bond_dict=None, unit_converter: UnitConverter=None)`, and store this as a private attribute instead of `_job`. Then over in the structure setter just directly use `self._unit_converter` instead of instantiating `uc`. (In the long run we should do similar conversions for the positions!) Then over in `LammpsBase._get_lammps_structure`, instead of setting the job attribute, just pass the unit converter in when instantiating the `LammpsStructure`.

@srmnitc the only other place I found `LammpsStructure` was over in `Calphy`. This proposed solution shouldn't effect you. But if you got the `UnitConverter` imported and used over in `calphy.job`, then we could change `LammpsStructure`'s signature again so that the unit converter is a required argument. IMO this is the right direction to go.

Leimeroth:
> In the Lammps code I didn't really like the entire job being passed as an attribute to `structure.LammpsStructure`. At the end of the day we just use it for unit conversions anyhow.
> 
> Reading it, I am actually a bit concerned that we didn't run into this problem beforehand. We got away with it because we always assume metal units, where distances are the same between pyiron and lammps, and now we care because velocities differ. But there are other Lammps units!
> 
> So I'd like to kill two birds with one stone: only pass the minimal amount of data needed from `LammpsBase` down to `LammpsStructure`, and set ourselves up to expand the use of unit conversions in the future.
> 
> To accomplish this, I suggest changing the signature of `LammpsStructure` to take the unit converter with a `None` default: `def __init__(self, input_file_name=None, bond_dict=None, unit_converter: UnitConverter=None)`, and store this as a private attribute instead of `_job`. Then over in the structure setter just directly use `self._unit_converter` instead of instantiating `uc`. (In the long run we should do similar conversions for the positions!) Then over in `LammpsBase._get_lammps_structure`, instead of setting the job attribute, just pass the unit converter in when instantiating the `LammpsStructure`.

I started to implement this, but realized that this could cause bugs because it would maybe be possible to decouple the units in the job and the structure. By back referencing the job ( which should be no problem, because it is only a pointer to the job object, i.e. no large memory or anything) it is made sure that the units are always in sync.
If for some reason I have the structure with some units as an object and then modify the units of the job they could be different leading to a whole bunch of problems. I am not really sure under which circumstances this could happen, but I imagine this being very hard to test for all weird cases one can come up with.

srmnitc:
> So in `structure.Atoms` it's very nice. I'd just ask to please go into the unit tests and test our new cases:
> 
>     * On instantiation velocities really are none
> 
>     * After being set, they exist as expected
> 
>     * Setting with the wrong shape throws an exception
>       These are absolutely trivial tests here, but they'll make sure it _keeps_ working.
> 
> 
> In the Lammps code I didn't really like the entire job being passed as an attribute to `structure.LammpsStructure`. At the end of the day we just use it for unit conversions anyhow.
> 
> Reading it, I am actually a bit concerned that we didn't run into this problem beforehand. We got away with it because we always assume metal units, where distances are the same between pyiron and lammps, and now we care because velocities differ. But there are other Lammps units!
> 
> So I'd like to kill two birds with one stone: only pass the minimal amount of data needed from `LammpsBase` down to `LammpsStructure`, and set ourselves up to expand the use of unit conversions in the future.
> 
> To accomplish this, I suggest changing the signature of `LammpsStructure` to take the unit converter with a `None` default: `def __init__(self, input_file_name=None, bond_dict=None, unit_converter: UnitConverter=None)`, and store this as a private attribute instead of `_job`. Then over in the structure setter just directly use `self._unit_converter` instead of instantiating `uc`. (In the long run we should do similar conversions for the positions!) Then over in `LammpsBase._get_lammps_structure`, instead of setting the job attribute, just pass the unit converter in when instantiating the `LammpsStructure`.
> 
> @srmnitc the only other place I found `LammpsStructure` was over in `Calphy`. This proposed solution shouldn't effect you. But if you got the `UnitConverter` imported and used over in `calphy.job`, then we could change `LammpsStructure`'s signature again so that the unit converter is a required argument. IMO this is the right direction to go.

@liamhuber I think it will not affect me. At the moment, I do not use `UnitConverter`

Leimeroth:
@liamhuber I am not sure why benchmark tests fail now, but as far as I can tell this should be unrelated to the PR

liamhuber:
> I started to implement this, but realized that this could cause bugs because it would maybe be possible to decouple the units in the job and the structure. By back referencing the job ( which should be no problem, because it is only a pointer to the job object, i.e. no large memory or anything) it is made sure that the units are always in sync.
If for some reason I have the structure with some units as an object and then modify the units of the job they could be different leading to a whole bunch of problems. I am not really sure under which circumstances this could happen, but I imagine this being very hard to test for all weird cases one can come up with.

Ah, very good point! Let's anyhow still make the `job` an optional (for now) argument. At the very least it stops `LammpsBase` from accessing a private variable of `LammpsStructure`, and like I said in the long run I think we need to make this argument mandatory.

tests/lammps/test_structure.py
```diff
@@ -0,0 +1,76 @@
...
+        # I still don't understand what's happening as deeply as I'd like (at all!) but I've been fighting with it too
+        # long, so for now I will just force the issue by redefining the project attribute(s). -Liam Huber
+        self.project = self.job.project
+        self.ref_project = self.ref.project
+
+    def tearDown(self) -> None:
+        super().tearDown()
+        self.ref_project.remove_jobs_silently(recursive=True)  # cf. comment in setUp
+
+    def test_velocity_basics(self):
```
liamhuber, line 51: Ah, I had actually thought this should belong in `tests.atomistics.structure.test_atoms`. The test itself looks good to me though!

In `tests.lammps.test_structure` I could imagine testing to make sure that `rotate_velocities` works as expected for both orthogonal and skewed cells. We should also test that the unit conversion is working, but I don't think that needs to be done here; handling units should probably be a separate PR that also converts the spatial units.

liamhuber:
> @liamhuber I am not sure why benchmark tests fail now, but as far as I can tell this should be unrelated to the PR

Yeah, the trace shows it failing in the environment setup so I agree it's probably unrelated. However, the only benchmark test existing *does* touch `Atoms`, so let's keep an eye on it. Hopefully when it reruns on the next push the env hiccup will have resolved itself -- it's fine on MacOS and Ubuntu so you should be a-ok.

niklassiemer:
It was actually failing for Mac (i restarted that one) and was only canceled for windows. The CI status dependent outcome of the benchmark tests is the reason for an own workflow. Thus, if they fail (with the usual not enough speedup erroe)- just rerun. If they persistently fail - maybe something really broke them xD

Leimeroth:
Can I merge this?

liamhuber:
> Can I merge this?

Sounds good to me. Now that the mac tests are passing I restarted the windows tests a couple days ago and they seem to have finished fine. Unless @niklassiemer wants to see the integration tests run on this, I'm happy to merge!

niklassiemer:
Its good practice to run the Integration tests before merging. However, they will latest run on merge with the master. Thus, I am fine with merging in this respect (of course if they fail on the master a rather urgent fix or revert would be required).

PR #666, opened 2022-06-29
Use lammps_create_atoms()

jan-janssen:


jan-janssen:
@srmnitc I am not sure if we fully implemented this in `pylammpsmpi` do you use it in calphy?

jan-janssen:
Be careful this is an evil pull request #666

samwaseda:
Is it a new functionality offered by LAMMPS?

jan-janssen:
> Is it a new functionality offered by LAMMPS?

I guess it was introduced in https://github.com/lammps/lammps/pull/2310 - roughly two years ago but it was new to me.

srmnitc:
@jan-janssen From [here](https://docs.lammps.org/Library_scatter.html), it seems that `scatter_atoms` can update atom properties after it has been created. So we may want to keep both.

> @srmnitc I am not sure if we fully implemented this in `pylammpsmpi` do you use it in calphy?

No its not used in calphy; the atoms are simply read in with `read_data` command

jan-janssen:
> @jan-janssen From [here](https://docs.lammps.org/Library_scatter.html), it seems that `scatter_atoms` can update atom properties after it has been created. So we may want to keep both.

Yes, we still have ` interactive_positions_setter()` that uses scatter to modify the positions, while ` interactive_structure_setter()` updates the whole structure 
https://github.com/pyiron/pyiron_atomistics/blob/lammps_create_atoms/pyiron_atomistics/lammps/interactive.py#L83

srmnitc (approved):

srmnitc:
> > @jan-janssen From [here](https://docs.lammps.org/Library_scatter.html), it seems that `scatter_atoms` can update atom properties after it has been created. So we may want to keep both.
> 
> Yes, we still have ` interactive_positions_setter()` that uses scatter to modify the positions, while ` interactive_structure_setter()` updates the whole structure https://github.com/pyiron/pyiron_atomistics/blob/lammps_create_atoms/pyiron_atomistics/lammps/interactive.py#L83

Thanks for the clarification! Everything looks good to me.

samwaseda (approved):
If it works why not

PR #669, opened 2022-07-05
Use _system_function_lst class variable to populate TableJob directly

niklassiemer:


pmrv (approved):
lgtm, thanks for taking care of this!

niklassiemer:
The only thing I do not like is to access the private variable...

pmrv:
If more pyiron modules want to provide additional table functions, we can add a class method that does the append.  For now I would wait until it is actually needed.

PR #674, opened 2022-07-07
np.matrix.trace() returns float

jan-janssen:


jan-janssen:
``` 
np.matrix.trace(np.eye(3)) == 3.0
> True
```

jan-janssen:
https://numpy.org/doc/stable/reference/generated/numpy.matrix.trace.html and https://numpy.org/doc/stable/reference/generated/numpy.trace.html

pmrv:
Even safer against numerical noise would probably be `not np.isclose(..., 3)`, but if the intent of that check is to test for the identity matrix I would actually prefer `np.isclose(prism.R, np.eye(3))`.

samwaseda:
> Even safer against numerical noise would probably be `not np.isclose(..., 3)`, but if the intent of that check is to test for the identity matrix I would actually prefer `np.isclose(prism.R, np.eye(3))`.

I was about to write exactly the same thing XD, although it should be with `.all()`: `np.isclose(prism.R, np.eye(3)).all()`

jan-janssen:
`rtol=0`, `atol=1e-08` - as a function

jan-janssen:
@samwaseda and @pmrv Any more suggestions? otherwise I would like to merge this pull request.

samwaseda:
Fine by me

PR #677, opened 2022-07-11
Calphy update

srmnitc:
Fixes/implements the following issues:

- [x] Preserve element order in pair coeff
- [x] Preserve element order in mass
- [x] Add list_potentials method

The first two points were implemented to fix the issues we noticed during the POTENTIALS workshop. Now the pair coeffs themselves need not be modified. Also the mass is picked in the same order as specified in pair_coeff.


pyiron_atomistics/calphy/job.py
```diff
@@ -334,6 +357,22 @@ def list_potentials(self):
         """
         return list(self.view_potentials()["Name"].values)
 
+    def structure_to_lammps(self, structure):
```
pmrv: Import from [here](https://github.com/pyiron/pyiron_atomistics/pull/678).

pmrv (approved):
LGTM.

pyiron_atomistics/calphy/job.py
srmnitc: Done

PR #680, opened 2022-07-13
Use JobType.register to add new job types

niklassiemer:


niklassiemer:
Well, I have after all to import the `TableJob` from atomistics... otherwise the additional functions are not added.

pyiron_atomistics/__init__.py
```diff
@@ -24,51 +24,53 @@
...
+JobType.register("pyiron_atomistics.interactive.sxextoptint", "SxExtOptInteractive")
+JobType.register("pyiron_atomistics.thermodynamics.sxphonons", "SxHarmPotTst")
+JobType.register("pyiron_atomistics.thermodynamics.sxphonons", "SxPhonons")
+JobType.register("pyiron_atomistics.thermodynamics.sxphonons", "SxUniqDispl")
+JobType.register("pyiron_atomistics.vasp.vasp", "Vasp")
+JobType.register("pyiron_atomistics.vasp.metadyn", "VaspMetadyn")
+JobType.register("pyiron_atomistics.vasp.vaspsol", "VaspSol")
+
+JobType.unregister("TableJob")
+JobType.register("pyiron_atomistics.table.datamining", "TableJob")
```
niklassiemer:
```suggestion
JobType.register("pyiron_atomistics.table.datamining", "TableJob", overwrite=True)
```

pmrv:
> Well, I have after all to import the `TableJob` from atomistics... otherwise the additional functions are not added.

Is it not possible to make it so that happens on the first import of (atomistic) `TableJob`, i.e. the first time a table is created by the user.

niklassiemer:
I just have to register the atomistics table job which is imported on first creation as usual. I only do not like this since if we add a continuum table job, or rather add some functions to the table job from continuum we have a clash between these two jobs... in that case I would like JobType to import all three modules, the base Job plus the extensions and return the final enhanced base job.

pmrv:
I've read the code again and it's actually *not* necessary to overwrite the table job in the job class dict, since 

https://github.com/pyiron/pyiron_atomistics/blob/dd6266b95ed107d51328de77cba500451af455d0/pyiron_atomistics/table/datamining.py#L57

already modifies the class (or class attribute) originally defined in base.  So as long as this file is imported the additional functions will be present.  I would recommend that we put this snippet in a function and then just call it from pyiron_atomistics __init__ to make it easier to follow, though.

niklassiemer:
Which would, however, again import the overall `TableJob`... That's what I meant - sorry for the confusion!

pmrv:
Ah, yes, there's no way to avoid importing the base `TableJob`, but imo that fine.

pyiron_atomistics/__init__.py
```diff
@@ -24,51 +24,52 @@
...
-JOB_CLASS_DICT["SxHarmPotTst"] = "pyiron_atomistics.thermodynamics.sxphonons"
-JOB_CLASS_DICT["SxPhonons"] = "pyiron_atomistics.thermodynamics.sxphonons"
-JOB_CLASS_DICT["SxUniqDispl"] = "pyiron_atomistics.thermodynamics.sxphonons"
-JOB_CLASS_DICT["TableJob"] = "pyiron_atomistics.table.datamining"
-JOB_CLASS_DICT["Vasp"] = "pyiron_atomistics.vasp.vasp"
-JOB_CLASS_DICT["VaspMetadyn"] = "pyiron_atomistics.vasp.metadyn"
-JOB_CLASS_DICT["VaspSol"] = "pyiron_atomistics.vasp.vaspsol"
+JobType.register("pyiron_atomistics.interactive.activation_relaxation_technique", "ART")
+JobType.register("pyiron_atomistics.testing.randomatomistic", "AtomisticExampleJob")
+JobType.register("pyiron_atomistics.calphy.job", "Calphy")
```
jan-janssen, line 29: I still wonder if this is any more readable than using the `JOB_CLASS_DICT`.

niklassiemer: In my opinion it is slightly more clear what is being done. `JobType.register` tells that you are registering a job type. Also the arguments of that function tell that they are `job_class_or_module_str` and `job_name`.  This, you cannot guess directly from `JOB_CLASS_DICT['job_name'] = "job.module:` but we rely on further documentation. Furthermore, changes to the `JobType` can be tracked and name clashes are reported. Once we have another `TableJob` in e.g. experimental, this framework also allows to add some hidden `_extra_require` dict to hold additional modules to be imported (i.e. please import `datamining.TableJob` from atomistics and from experimental and return the standard `TableJob` from base).

PR #700, opened 2022-08-03
Only include necessary POTCARs

pmrv:
It can happen that a structure defines more chemical elements on it than are actually present in it.  For example `Atoms(species=['Fe','Al'], indices=[1,1,1,1], ....)` defines a structure with four aluminum inside.  `StructureStorage` does this systematically to ensure that indices of structures from one container are consistent and but it can also happen after deleting atoms from a structure or something similar. 

When a Vasp job sees such a structure it starts writing out POTCARs for all elements defined instead of those that are present.  Since the POSCAR however only has one element present, the calculation ends up using the wrong pseudopotential.

I've made a quick fix by ensuring that only POTCARs of elements present in the structure are written, but it's a bit of a footgun waiting to happen again.  I'm open for other suggestions how to fix this more cleanly.

### DIFF
--- a/pyiron_atomistics/vasp/base.py
+++ b/pyiron_atomistics/vasp/base.py
@@ -109,8 +109,10 @@ class VaspBase(GenericDFTJob):
         """
         GenericDFTJob.structure.fset(self, structure)
         if structure is not None:
+            elements = structure.get_species_symbols().tolist()
+            elements = [el for el in elements if el in structure.get_chemical_symbols()]
             self._potential = VaspPotentialSetter(
-                element_lst=structure.get_species_symbols().tolist()
+                element_lst=elements
             )
 
     @property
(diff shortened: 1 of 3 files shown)


Leimeroth:
Since this also messed with my calculations (See #718) already, I assume that the species property of the structures itself is dangerous right now. Maybe something like that maps:
np.unique(self.numbers)
to symbols would be better?

Leimeroth:
bump

ligerzero-ai:
A working/hacky solution would be to convert it to a pymatgen structure to avoid the chemical API completely.

Then call this function that I had for my prior workflows:

```
 struct = pyiron_to_pymatgen(structure)
 site_element_list = [site.species_string for site in struct]
 potcar_ele_list = [site_element_list[0]]

 for i, element in enumerate(site_element_list [1:]):
     if element != site_element_list[i]:
         potcar_ele_list.append(element)
```

It's not elegant, but it works, and fixes a dangerous error that's still lurking. The only issue I have with this is that this can get expensive for large structures, if ordering is preserved (which won't be the case for most users anyway). In this case it may be necessary to import a library, maybe itertools, which could be faster (I have no idea which one is faster right now.)

I am not so familiar with ASE so if there is an equivalent way to do this it may be preferable.

pmrv:
Obsolete by #799

PR #703, opened 2022-08-05
VASP -- write kintetic energy

raynol-dsouza:
I want to save the kinetic energy data in the VASP OUTCAR file for MD runs. While trying to run MD using the Langevin thermostat, I noticed that the difference between <job.output.energy_tot> and <job.output.energy_pot> was constant for every step. Upon further digging, I found that a single value was being saved for the kinetic energy, rather than saving a list of values for each MD step. 

Seeing that in lines 2048-2052 of /vasp/base.py, <job.output.energy_pot> is calculated using this single (wrong) kinetic energy value (for MD runs), I think this needs a fix.

I may be grossly misunderstanding something here, so feel free to correct me. But if I am right, this PR should take the correct kinetic energies, and also save them to the output dictionary.

### DIFF
--- a/pyiron_atomistics/calphy/job.py
+++ b/pyiron_atomistics/calphy/job.py
@@ -148,6 +148,7 @@ class Calphy(GenericJob):
         self._data = None
         self.input._pot_dict_initial = None
         self.input._pot_dict_final = None
+        self.input._calc = None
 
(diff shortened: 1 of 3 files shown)


raynol-dsouza:
@jan-janssen @pmrv @niklassiemer Can this PR be merged?

pmrv (approved):
lgtm

PR #705, opened 2022-08-08
Lammps parse dump pandas

### LINKED ISSUE #671: Parsing large lammps dummps leads to kerneal deaths
**Summary**

When trying to parse some lammps dump files I encountered repeated kernel deaths of the jupyter notebook.
I assume the reason is that similar to the parsing of logfiles the dumpfile is read using readlines(), causing out of memory problems because the whole file, that can have the size of a few GB is read into memory as a whole, but the reason could also be different.

Solution is probably to read line by line and fill some buffer with the data, regularly writing to hdf5. But the same problem will probably occur when the job is reloaded from the hdf5 or is that handled somehow already?

Latest versions, Linux

**Expected Behavior**

Working lammps parse

**Actual Behavior**

Dying Kernel

**Steps to Reproduce**

Try to parse a large lammps dump file.

Leimeroth:
Fixes #671 for my calculations at least. Temporarily I fixed it by allocating 26GB of RAM for my notebooks, but even that was not sufficient for some of my recent calculations.
With this one of the simulations that kept killing the kernel during parsing never allocated more than 3-4 GB of RAM (according to top, didnt check in detail)

Did not check whether this gives the same numbers as the old parser yet. Is correct parsing part of unit tests?


pmrv:
For testing there's some water trajectories that are checked in a round-about way, but I would have to see if it applies here.  I would just move the old method in the unit tests and include a sample dump in the test resources, then check if new and old give the same results.

Leimeroth:
> For testing there's some water trajectories that are checked in a round-about way, but I would have to see if it applies here. I would just move the old method in the unit tests and include a sample dump in the test resources, then check if new and old give the same results.

The water tests failed and seems that everything is covered, but locally the integration tests ran now, so I assume I fixed the last problems. Adding the old method to the unit tests would require to permanently keep it in. I wouldn't mind a little more code, but maybe it would be confusing or is not wanted to not have too much old struff in there.

liamhuber (approved):
LTGM. Please take a look at #716 which makes a cosmetic change, but I am content to have this merged without that if you dislike it.

jan-janssen:
@samwaseda  Can you take a look, as I remember you implemented the current version of the LAMMPS parser and as already discussed in https://github.com/pyiron/pyiron_atomistics/pull/658 I have the feeling we are moving back and forth on this issue?

pmrv:
> > For testing there's some water trajectories that are checked in a round-about way, but I would have to see if it applies here. I would just move the old method in the unit tests and include a sample dump in the test resources, then check if new and old give the same results.
> 
> The water tests failed and seems that everything is covered, but locally the integration tests ran now, so I assume I fixed the last problems. Adding the old method to the unit tests would require to permanently keep it in. I wouldn't mind a little more code, but maybe it would be confusing or is not wanted to not have too much old struff in there.

Sorry for being late.  Imo there's no problem keeping extra code in the unit tests for a few releases.  That's what they're for, to make sure known behavior continues to work.

Leimeroth:
I did some benchmarks regarding timing and memory usage (using ipythons %%timeit and %%memit requiring the memory_profiler package) on job.collect_dump_file() and job.collect_dump_file_old().
The new version is about 1.2 times faster and more important, it requires 1/3 to 1/2 of the memory. While %%timeit only has a very low error the memory usage changes a lot more from test to test. f.e.
new:
peak memory: 673.71 MiB, increment: 340.40 MiB
peak memory: 647.71 MiB, increment: 314.33 MiB
peak memory: 616.68 MiB, increment: 283.29 MiB
peak memory: 580.42 MiB, increment: 247.03 MiB
peak memory: 598.57 MiB, increment: 265.18 MiB
peak memory: 611.34 MiB, increment: 277.95 MiB
peak memory: 673.91 MiB, increment: 340.53 MiB
peak memory: 598.67 MiB, increment: 265.28 MiB
2.96 s ± 14 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

Old:
peak memory: 1104.89 MiB, increment: 771.50 MiB
peak memory: 1097.41 MiB, increment: 673.98 MiB
peak memory: 1179.28 MiB, increment: 749.60 MiB
peak memory: 1200.61 MiB, increment: 773.00 MiB
peak memory: 1125.32 MiB, increment: 697.46 MiB
peak memory: 1201.58 MiB, increment: 773.46 MiB
peak memory: 1201.09 MiB, increment: 762.42 MiB
peak memory: 1183.31 MiB, increment: 744.39 MiB
3.57 s ± 36.8 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

Edit: Testjob
```
job = pr.create.job.Lammps("TestL")
job.structure = pr.create.structure.ase.bulk("Cu", cubic=True).repeat(15)
job.calc_md(initial_temperature=4000, n_ionic_steps=10000)
job.run()
job.decompress()
```

Leimeroth:
@liamhuber if you agree I would merge this now.

PR #710, opened 2022-08-09
Move some of the structure plotting functions from contrib

pmrv:
They were previously defined on the `TrainingContainer` over in contrib, but I have added also `distances` to plot a plain histogram of nearest neighbor distances.

Corresponding PR to contrib once this is through.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 # Unreleased
 
+- Add ploting functions from pyiron_contrib to StructureContainer ([#710](https://github.com/pyiron/pyiron_atomistics/pull/710))
+
 # pyiron_atomistics-0.2.41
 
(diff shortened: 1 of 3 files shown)


Leimeroth:
So TrainingContainers plotting interface should inherit from this than?

pmrv:
Yeah, that's how I was planning it.  These plotting functions would be purely concerned with structural data and the training container ones would then only add things related to energies, forces, etc.

Leimeroth (approved):
I like the idea. If some errors occur down the line we should fix them there.

pmrv:
I didn't want to pull in seaborn for just one plot, so I put it in an `ImportAlarm`.  @niklassiemer are we tracking our opt deps somewhere?

pmrv:
I also remembered to update the change log. :')

jan-janssen (approved):
Looks good to me

PR #713, opened 2022-08-09
Raise error on raised DeprecationWarning

niklassiemer:


niklassiemer:
Obviously, this is not working as intended...!

niklassiemer:
That is somehow interesting... This looks like the coverage of all the unittests is recorded just by being called through coverage. Do we then need to actually run the unittests in the coverage workflow?!
EDIT: The coverage changes look unrelated to any windows/linux switch and are, thus, probably completely unrelated.

PR #714, opened 2022-08-09
Remove unused deps

pmrv:
`mock` and `futures` are both py2/3 compatibility libraries, which we don't need anymore.  @niklassiemer this might require your attention on next release? :S


jan-janssen (approved):
Looks good to me

PR #715, opened 2022-08-09
Remove pointless try/except and pytables dep

pmrv:
I'm not even sure what the original code wanted to do, but in the current way there's no way `GenericParameters.get` can cause an exception from `pytables` (which we don't import) to be raised, so I've axed it.


niklassiemer (approved):
The original code looks like madness, repeat the same thing and expect something different xD your changes lgtm!

jan-janssen (approved):
Looks good to me.

PR #721, opened 2022-08-16
dont set initial temperature to 0 for NVE

Leimeroth:
Initial temperature was set to 0 for NVE simulations

### DIFF
--- a/pyiron_atomistics/lammps/control.py
+++ b/pyiron_atomistics/lammps/control.py
@@ -525,6 +525,8 @@ class LammpsControl(GenericParameters):
             if langevin:
                 warnings.warn("Temperature not set; Langevin ignored.")
             fix_ensemble_str = "all nve"
+            if initial_temperature is None:
+                initial_temperature = 0
 
         if tloop is not None:
             fix_ensemble_str += " tloop " + str(tloop)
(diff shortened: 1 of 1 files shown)


pyiron_atomistics/lammps/control.py
```diff
@@ -525,7 +525,8 @@ def calc_md(
             if langevin:
                 warnings.warn("Temperature not set; Langevin ignored.")
             fix_ensemble_str = "all nve"
-            initial_temperature = 0
+            if initial_temperature is None:
+                initial_temperature = 0
```
liamhuber:
```suggestion
            initial_temperature = 0 if initial_temperature is None else initial_temperature
```

I've been one-lining this type of check.

liamhuber (approved):
LGTM. While you're here, it might be worth quickly checking the docs for this method -- since it's not thermostatted, if you give an equilibrium initial structure and set the temperature to 300, you'll get an effective temperature of 150 because of energy equipartition. I don't think we should actually double the initial temperature since there is no necessity to start with an equilibrium structure, but since such a structure is a _common_ starting spot, it may be worth making sure the equipartition behaviour is clear in the docs.

PR #722, opened 2022-08-16
Fix Input HDF serialization in ScipyOptimizer

pmrv:
This currently blocks [this PR](https://github.com/pyiron/pyiron_base/pull/823) in pyiron_base. 

See the commit messages for detailed description of the issue, but the problem was that the input didn't write itself to the right location in HDF.  This meant also that `ScipyOptimizer` jobs could not be reloaded in the past, therefore I'm not trying to be backwards compatible here.

### DIFF
--- a/pyiron_atomistics/interactive/scipy_minimizer.py
+++ b/pyiron_atomistics/interactive/scipy_minimizer.py
@@ -4,10 +4,11 @@
 
 import numpy as np
 from pyiron_atomistics.atomistics.job.interactivewrapper import InteractiveWrapper
-from pyiron_base import DataContainer
+from pyiron_base import HasStorage
 from pyiron_atomistics.atomistics.job.interactive import GenericInteractiveOutput
 from scipy.optimize import minimize
 import scipy
+from typing import List, Iterable
 import warnings
 
 __author__ = "Osamu Waseda"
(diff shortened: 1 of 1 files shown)


pyiron_atomistics/interactive/scipy_minimizer.py
```diff
@@ -291,24 +292,84 @@ def calc_minimize(
...
+        self.storage.pressure = value
+
+    @property
+    def volume_only(self) -> bool:
+        """bool: only pressure minimization"""
+        return self.storage.volume_only
+
+    @volume_only.setter
+    def volume_only(self, value: bool):
+        self.storage.volume_only = value
```
jan-janssen: I thought when using the data container, we do not have to define the individual properties, so I am a bit surprised you still do it here manually. Maybe we can use the set attribute functionality to reduce this redundancy.

niklassiemer: Using `DataContainer` subclasses directly causes a lot of trouble if one does not pay attention to the recursive data structure (For that reason this Input subclass was not working as intended). `HasStorage`, however, has everything inside the `storage` attribute in the end and we need to pull this up one level... We had quite some discussion in https://github.com/pyiron/pyiron/issues/1412 how to handle these things nicely. However, so far no one has implemented the corresponding class.

pmrv: Another reason I did it manually is to give well defined location to add input validation.  (I had tried this with the pressure attribute, but got it wrong and I don't have time to do it properly.)

PR #724, opened 2022-08-18
Calphy update

srmnitc:


pyiron_atomistics/calphy/job.py
```diff
@@ -264,6 +281,27 @@ def _prepare_pair_styles(self) -> Tuple[List, List]:
 
         return pair_style, pair_coeff
 
+    def _get_element_list(self) -> List[str]:
+        """
+        Get masses as defined in pair style
```
pmrv: `masses` is just a typo I guess.

pyiron_atomistics/calphy/job.py
```diff
@@ -284,7 +322,13 @@ def _get_masses(self) -> List[float]:
                 index = list(elements_struct_lst).index(element_name)
                 masses.append(elements_object_lst[index].AtomicMass)
 
-        return masses
+        #this picks the actual masses, now we should pad with 1s to match length
+        length_diff = len(elements_from_pot)-len(masses)
+        #if length_diff>0:
+        #    for x in range(length_diff):
+        #        masses.append(1.0)
+
```
pmrv: Is that still necessary?

pyiron_atomistics/calphy/job.py
```diff
@@ -640,6 +699,17 @@ def run_static(self):
...
+        """
+        """
+        if not iteration_step in [0, -1]:
+            raise ValueError("get_structure for calphy can only give frame 0 or -1")
+        if iteration_step == 0:
+            return self.structure
+        elif iteration_step == -1:
+            return self.output.structure_final
+
+
```
pmrv: Implement `HasStructure` instead of providing this method directly.

pyiron_atomistics/calphy/job.py
```diff
@@ -651,34 +721,82 @@ def collect_general_output(self):
...
+            #collect ediffs and so on
             f_ediff, b_ediff, flambda, blambda = self.collect_ediff()
             self.output.forward_energy_diff = list(f_ediff)
             self.output.backward_energy_diff = list(b_ediff)
             self.output.forward_lambda = list(flambda)
             self.output.backward_lambda = list(blambda)
+
+            #get final structure
+            traj = Trajectory(os.path.join(self.working_directory, "conf.equilibration.dump"))
+            aseobj = traj[0].to_ase(species=self.calc.element)[0]
```
pmrv, line 751: What's the last `[0]` do?

pmrv: Also why the conversion `pyiron.Atoms` -> `ase.Atoms` and then back again, when `aseobj` doesn't seem to be used afterwards?

pyiron_atomistics/calphy/job.py
```diff
@@ -651,34 +721,82 @@ def collect_general_output(self):
...
+            pyiron_atoms = ase_to_pyiron(aseobj)
+            self.output.structure_final = pyiron_atoms
+
             if self.input.mode == "ts":
```
pmrv, line 755: Maybe separate methods depending on `input.mode` would be more readable, but your call.

pyiron_atomistics/calphy/job.py
```diff
@@ -651,34 +721,82 @@ def collect_general_output(self):
...
+                self.output.ts_backward_cells = bkd_cells
+
             elif self.input.mode == "pscale":
                 datfile = os.path.join(self.working_directory, "pressure_sweep.dat")
                 p, fe, ferr = np.loadtxt(datfile, unpack=True, usecols=(0, 1, 2))
                 self.output.energy_free = np.array(fe)
                 self.output.energy_free_error = np.array(ferr)
                 self.output.pressure = np.array(p)
 
+                f_ediff, b_ediff, f_vol, b_vol, f_press, b_press, flambda, blambda = self.collect_thermo(mode="ts")
```
pmrv: I assume `mode="pscale"` is intended?

pyiron_atomistics/calphy/job.py
```diff
@@ -731,10 +849,97 @@ def collect_ediff(self):
...
+        f_ediff = []
+        b_ediff = []
+        f_vol = []
+        b_vol = []
+        f_press = []
+        b_press = []
+
+        for i in range(1, self.input.n_iterations + 1):
+            fwdfilename = os.path.join(self.working_directory, "%s.forward_%d.dat" % (mode, i))
+            bkdfilename = os.path.join(self.working_directory, "%s.backward_%d.dat" % (mode, i))
```
pmrv: f-strings are preferred.

pyiron_atomistics/calphy/job.py
```diff
@@ -731,10 +849,97 @@ def collect_ediff(self):
...
+        """
+        db_dict = super(Calphy, self).db_entry()
+        if self.structure:
+            if isinstance(self.structure, Atoms):
+                parent_structure = self.structure.get_parent_basis()
+            else:
+                parent_structure = self.structure.copy()
+            db_dict["ChemicalFormula"] = parent_structure.get_chemical_formula()
+
+        db_dict["projectpath"] = self.project_hdf5.project_path
```
pmrv: Is this not provided by `super()`?

pmrv (changes requested):
See my comments, but mostly code-style.

pyiron_atomistics/calphy/job.py
srmnitc: Corrected

pyiron_atomistics/calphy/job.py
srmnitc: No, removed.

pyiron_atomistics/calphy/job.py
srmnitc: @pmrv Could you please point me to an example where it is used?

pyiron_atomistics/calphy/job.py
srmnitc: [`Trajectory`](https://pyscal.org/en/latest/examples/15_trajectory_module.html) is from pyscal to handle LAMMPS dump files. It always returns a list, therefore the last `[0]` in line 752. `Trajectory` reads dump file, returns ASE atoms, which I convert to pyiron atoms and is stored.

pyiron_atomistics/calphy/job.py
srmnitc: I'm not sure I follow, the first part is general for all modes, then the specific parts depending on the mode.

pyiron_atomistics/calphy/job.py
srmnitc: You are right, thanks! fixed.

pyiron_atomistics/calphy/job.py
srmnitc: No, the `projectpath` remained empty.

pyiron_atomistics/calphy/job.py
pmrv: Ah, my bad.  I sort of assumed it's our Trajectory and got confused.  Can you maybe rename it in the import?

pyiron_atomistics/calphy/job.py
pmrv: Sounds like a bug then.  Ping @liamhuber @niklassiemer didn't we have a similar situation somewhere the projectpath is not correctly set up depending on some configuration issue?

pyiron_atomistics/calphy/job.py
pmrv: Yeah, I meant to put the parts in the if clauses into separate methods, but as I said no strong opinions.

pyiron_atomistics/calphy/job.py
pmrv: Here are some example from [StructureContainer](https://github.com/pyiron/pyiron_atomistics/blob/833eb7b0d76590bd09058e1fd28192cba82a0d79/pyiron_atomistics/atomistics/job/structurecontainer.py#L134) and [StructureStorage](https://github.com/pyiron/pyiron_atomistics/blob/833eb7b0d76590bd09058e1fd28192cba82a0d79/pyiron_atomistics/atomistics/structure/structurestorage.py#L224).  

Here's also the [docs](https://pyiron-atomistics.readthedocs.io/en/latest/apidoc/pyiron_atomistics.atomistics.structure.has_structure.html?highlight=HasStructure#pyiron_atomistics.atomistics.structure.has_structure.HasStructure).

pyiron_atomistics/calphy/job.py
niklassiemer: # TL;DR
if `projectpath` remained empty, you probably are not in one of the defined `PROJECT_PATHS` of your settings and  `PROJECT_CHECK_ENABLED` is `False`.
The handling inside the `super()` call should be the right behavior!
```suggestion
```

# Full answer
We have the very confusing situation that our notion of 'projectpath' changes depending on where in the code we are...:

- In the database (and the database code) we have 'projectpath' being the paths defined in our settings and 'project' being the path to the project
- In the jobs we have 'projectpath' being the path to the project and 'root_path' being the corresponding path defined in the settings.

In the `db_entry` these two worlds are translated to each other:
```python
            "projectpath": self.project_hdf5.root_path,
            "project": self.project_hdf5.project_path,
```
is the definition in [`GenericJob.db_entry()`](https://github.com/pyiron/pyiron_base/blob/19d6c3c5180fee5b0c0ccf91f1ec516a200cc624/pyiron_base/jobs/job/generic.py#L1167-L1168)

Depending on the settings (`PROJECT_PATHS` and `PROJECT_CHECK_ENABLED` definitions) 
- we split the full path  `/settings/project/path/path/to/actual/project` into `/settings/project/path/` and `path/to/actual/project`,
- while we 'split' `/not/a/settings/project/path/path/to/actual/project` into `None` and `/not/a/settings/project/path/path/to/actual/project` if `PROJECT_CHECK_ENABLED=False`
- and `raise ValueError` else. 
Splitting of `root_path` is done in [database.manager.DatabaseManager.top_path](https://github.com/pyiron/pyiron_base/blob/19d6c3c5180fee5b0c0ccf91f1ec516a200cc624/pyiron_base/database/manager.py#L213-L235)

niklassiemer (changes requested):
Trust `super()` in `db_entry`

pyiron_atomistics/calphy/job.py
srmnitc: Thanks!

pyiron_atomistics/calphy/job.py
srmnitc: Renamed as PyscalTrajectory

pyiron_atomistics/calphy/job.py
srmnitc: Thanks!

srmnitc: Implemented this.

srmnitc:
@pmrv @niklassiemer I have a bit of a general question. Calphy has a class called `Calculation`, which is recreated when `from_hdf` is called. At the moment, the tests fail as calphy is not installed automatically in the test environment, and it does not understand what `Calculation` is. Installing calphy in the testing environment will then bring a host of dependencies in, including LAMMPS, which I guess is not ideal. What would be the best way to solve this situation?

pmrv:
One idea might be to look into [mocking](https://docs.python.org/3/library/unittest.mock.html?highlight=unittest%20mock#patch).  The section on patching has some info on how to fake certain objects.  I have never used it myself, but maybe we can fake the full calphy module.

pyiron_atomistics/calphy/job.py
srmnitc: changed

pyiron_atomistics/calphy/job.py
```diff
@@ -30,17 +38,19 @@
 __status__ = "development"
 __date__ = "April 1, 2022"
 
-inputdict = {
+_inputdict = {
```
niklassiemer: Thinking about it, I would move this dict to be a property of the Calphy class:
```python
    @property
    def _default_input(self):
        return {
            "mode": None, 
            ...
        }
```

pyiron_atomistics/calphy/job.py
```diff
@@ -651,33 +712,107 @@ def collect_general_output(self):
...
-            self.output.energy_free_error = self._data["results"]["error"]
-            self.output.energy_free_reference = self._data["results"][
+                self.output["spring_constant"] = self._data["average"][
+                    "spring_constant"
+                ]
+            if "density" in self._data["average"].keys():
+                self.output["atomic_density"] = self._data["average"]["density"]
+            self.output["atomic_volume"] = self._data["average"]["vol_atom"]
+
+            # main results from mode fe
```
niklassiemer, line 729: While I can read the most concepts (without deep understanding) I stumble at the abbreviations. 'fe' means free energy?

pyiron_atomistics/calphy/job.py
```diff
@@ -651,33 +712,107 @@ def collect_general_output(self):
...
+                (
+                    f_ediff,
+                    b_ediff,
+                    f_vol,
+                    b_vol,
+                    f_press,
+                    b_press,
+                    flambda,
+                    blambda,
+                ) = self.collect_thermo(mode="ts")
```
niklassiemer: 'ts' = transition state?

pyiron_atomistics/calphy/job.py
```diff
@@ -731,10 +866,104 @@ def collect_ediff(self):
 
         return f_ediff, b_ediff, flambda, blambda
 
+    def collect_thermo(self, mode="ts"):
```
niklassiemer: Is this supposed to be used by a user? If not I would turn it private

pyiron_atomistics/calphy/job.py
```diff
@@ -640,6 +694,13 @@ def run_static(self):
...
         self.run()
 
+    def _get_structure(self, frame=-1, wrap_atoms=False):
+        """ """
+        return [self.structure, self.output.structure_final][frame]
+
+    def _number_of_structures(self):
+        return 2
+
     def collect_general_output(self):
```
niklassiemer, line 709: Is this supposed to be used by a user? If not I would turn it private

pyiron_atomistics/calphy/job.py
```diff
@@ -651,33 +712,107 @@ def collect_general_output(self):
...
+                    f_ediff,
+                    b_ediff,
+                    f_vol,
+                    b_vol,
+                    f_press,
+                    b_press,
+                    flambda,
+                    blambda,
+                ) = self.collect_thermo(mode="pscale")
+                self.output["ps/forward/energy_diff"] = list(f_ediff)
```
niklassiemer, line 813: ps = pscale = pressure scale?

pyiron_atomistics/calphy/job.py
```diff
@@ -57,10 +67,19 @@
...
     },
+    "nose_hoover": {
+        "thermostat_damping": 0.1,
+        "barostat_damping": 0.1,
+    },
+    "berendsen": {
+        "thermostat_damping": 100.0,
+        "barostat_damping": 100.0,
+    },
+    "calc": {},
```
niklassiemer: Here, you mentioned you do not need this?

niklassiemer (commented):
Some nits, otherwise LGTM.

pyiron_atomistics/calphy/job.py
srmnitc: Sounds like a good idea to me, changed it.

pyiron_atomistics/calphy/job.py
srmnitc: Turned it to private, along with a couple of other methods.

pyiron_atomistics/calphy/job.py
srmnitc: Done, along with some others

pyiron_atomistics/calphy/job.py
srmnitc: Done

pyiron_atomistics/calphy/job.py
srmnitc: Yes, you are right. But I also see the problem. I updated the docstrings now to explain a bit more on where you can find the output.

pyiron_atomistics/calphy/job.py
srmnitc: Correct, also updated the docstrings,.

pyiron_atomistics/calphy/job.py
srmnitc: temperature scaling, also updated the docstrings.

srmnitc:
> Some nits, otherwise LGTM.

Great thanks @niklassiemer! If the tests pass I will go ahead and merge it.

niklassiemer:
Yes, I try to manage to get the tests passing on master, again. Then, we check here again :)

niklassiemer (approved):

PR #737, opened 2022-09-01
Reintroduce Raise error when incompatible structures are set on interactive jobs

niklassiemer:
reimplements pyiron/pyiron_atomistics#731


pmrv:
Ah, sorry for the bother.  I just kinda wish python had element-wise operators like Julia.

pmrv:
Notebook tests pass now, this was the offending one, right?  I'll merge this in a bit then.

niklassiemer:
'Notebooks for Integration Tests' were failing (therefore the integration label). Since this is ok now, I merge again :)

PR #740, opened 2022-09-01
Short cut ChemicalElement comparison if same object

pmrv:
When replacing species types in large structures a lot of comparisons between `ChemicalElements` are triggered. Previously the equality check was implemented by comparing the *full* data that describes the element.  This is quite costly and not necessary since `ChemicalElements` is now a singleton object.  I have therefore added a short circuit check when the elements compared are the same object.  Since the singleton-ness of `ChemicalElements` can be considered an implementation detail, I've still left the old implementation in place as a fallback.


niklassiemer (approved):
Assume equality for same objects seems reasonable!

PR #741, opened 2022-09-01
Never let format_black fail

pmrv:
Previously the format_black workflow fails, if the checked code conforms to black already (because git complains that there's nothing to commit). I always exit 0 now, so that this is hopefully silenced.

I'm not sure if exit and such are supported by the workflow template and haven't tested it.

### DIFF
--- a/.github/workflows/format_black.yml
+++ b/.github/workflows/format_black.yml
@@ -28,6 +28,7 @@ jobs:
         git config --local user.email "pyiron@mpie.de"
         git config --local user.name "pyiron-runner"
         git commit -m "Format black" -a
+        exit 0
     - name: push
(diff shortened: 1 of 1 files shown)


.github/workflows/format_black.yml
```diff
@@ -28,6 +28,7 @@ jobs:
         git config --local user.email "pyiron@mpie.de"
         git config --local user.name "pyiron-runner"
         git commit -m "Format black" -a
+        exit 0
```
niklassiemer:
```suggestion
        git commit -m "Format black" -a || exit 0
```

niklassiemer:
It already failed with non-zero exit code before that line. Now it always is ok to use the black workflow :)

PR #748, opened 2022-09-06
release ipywidgets again

niklassiemer:


pmrv:
Upstream merged my PR a while ago, so this should work again.

PR #757, opened 2022-09-16
Add a new structure factory for the MaterialsProject

pmrv:
Only two API calls are wrapped for now:
1. the summary search
2. get a structure by ID

More APIs to obtain structure exist (grain boundaries..), but should be implemented as needed.

The search method returns a custom object implementing HasStructure. This is done so that we can add methods here that convert the results to other convenient objects, like:
- plain list
- StructureStructure
- pandas dataframe?

If you want to test this, be aware that you need an account with the materials project and an API key.  See the docstring of how to get one.


pmrv:
This won't run yet because it depends on the `mp_api` package from the materials project that provides an interface to the REST api.  However that is not on conda only on pypi, what's our procedure for this?  Create a new conda package?

niklassiemer:
> This won't run yet because it depends on the `mp_api` package from the materials project that provides an interface to the REST api.  However that is not on conda only on pypi, what's our procedure for this?  Create a new conda package?

Yes, that's what we usually do. As long as this is plain python and this sounds like such a case this is really straight forward to do.

pmrv:
I found out you can actually put pip packages as dependencies into the `environment.yml` like [this](https://conda.io/projects/conda/en/latest/user-guide/tasks/manage-environments.html#create-env-file-manually) (see the example last line). I don't have time to maintain additional packages and don't to put it on someone else, so I'll just go with this route if no one objects.

jan-janssen:
> I found out you can actually put pip packages as dependencies into the `environment.yml` like [this](https://conda.io/projects/conda/en/latest/user-guide/tasks/manage-environments.html#create-env-file-manually) (see the example last line). I don't have time to maintain additional packages and don't to put it on someone else, so I'll just go with this route if no one objects.

As it is a pypi package, I opened a pull request to add it to conda-forge https://github.com/conda-forge/staged-recipes/pull/20471

pmrv:
Thanks, @jan-janssen !

jan-janssen (approved):
Looks good to me

PR #766, opened 2022-09-20
Find potential by structure

liamhuber:
I want to be able to see what Lammps potentials are available for a structure without first creating a job, e.g. [here](https://github.com/pyiron/ironflow/pull/62).

To accomplish this I moved the `list_potentials` and `view_potentials` methods of the Lammps job over to be functions in `pyiron_atomistics.lammps.potential` and exposed them in the init to allow `from pyiron_atomistics.lammps import view_potentials, list_potentials`.

We can discuss whether they should be even more available than that, but for now they're accessible to expert users and don't get in the way of anything else.


pyiron_atomistics/lammps/potential.py
```diff
@@ -2,7 +2,7 @@
 # Copyright (c) Max-Planck-Institut für Eisenforschung GmbH - Computational Materials Design (CM) Department
 # Distributed under the terms of "New BSD License", see the LICENSE file.
 
-from __future__ import print_function
+from __future__ import print_function, annotations
```
pmrv:
```suggestion
```

Both the `print` function and type annotations should be in all python versions we support.

pmrv (approved):
`Project` defines a method `inspect_emperical_potentials` which also allows to do this for users in notebooks, but I don't see an issue with reshuffling the methods here internally.  I'm not actually sure how this is implemented there, but I guess it just returns a lammps potential object.

liamhuber:
I think it's actually a data frame with all the values needed to construct such a potential object actually. Anyhow I also didn't see any issues with an internal reshuffle, and in my other project it is very convenient to have the functionality independent of a project instance as well. Thanks for the timely review!

pyiron_atomistics/lammps/__init__.py
```diff
@@ -0,0 +1 @@
+from .potential import view_potentials, list_potentials
```
jan-janssen: Please use absolute imports!

PR #767, opened 2022-09-20
Empty potential list is OK

liamhuber:
This is never None, just empty sometimes, who cares. The empty list was getting returned anyway, so let's cut out the unused code.


liamhuber:
This has been up for a couple days along with #768 with no complaints. Since all it does is remove code for a logical branch that's never reached, I'm just going to merge it.

Issue #752, opened 2022-09-08
Pyiron generated VASP inputs: Option to preserve atomic ordering 


ligerzero-ai:
Hi, I am just beginning to code some proof-of-concept workflows. There are a couple issues that immediately cropped up:

It is highly desirable for complex workflows that the POSCARs generated have the option to have their atomic ordering preserved, instead of rearranged. 

e.g. 
``` Fe 36 V 1 Fe 2 Al 1 Fe 31  remains that way ```
and doesn't become 
``` Fe 69 V 1 Al 1```

This is because the ordering of the atoms is important for complex derived workflows (e.g. interaction energy in co-segregation needs to be calculated by replacing a specific atom with another element, this needs to be considered with the exact same atom in a single-segregation case in the same position). 

As it stands, pyiron will rearrange the atomic ordering for single-segregation cases and this will make subsequent co-segregation workflows non-trivial, when they should be trivial. Data such as movement from original position, etc. which can be important are highly non-trivial to compare between "like" systems if you rearrange the structures on each pyiron job generation. Comparing atoms to determine which is the "same" from the original position becomes non-trivial compared to just using the "dumb" method and relying on atomic ordering. 

Then, immediately I would see that POTCAR generation, MAGMOM flag generation becomes very different too. Could someone point me to the right places so I can implement this?

liamhuber:
> It is highly desirable for complex workflows that the POSCARs generated have the option to have their atomic ordering preserved,

I'm not convinced that the VASP files themselves need to preserve order, although I agree it's absolutely critical that, e.g., `job.structure.positions` and `job.output.positions[-1]` maintain *exactly* the same ordering. If the VASP job internally re-orders these for the VASP run and then converts them back to the "true" pyiron order in the output, I don't see a problem with that. 

> As it stands, pyiron will rearrange the atomic ordering for single-segregation cases and this will make subsequent co-segregation workflows non-trivial, when they should be trivial. 

I thought this order-preservation was already the current behaviour, but I have to confess I'm not a regular VASP user so I might be very wrong about that. If your claim here is correct, then that is a straight up bug in our pyiron-VASP interface.

> e.g.
`Fe 36 V 1 Fe 2 Al 1 Fe 31` remains that way
and doesn't become
`Fe 69 V 1 Al 1`

[...]

PR #769, opened 2022-09-20
Create a setting to forbid sorting behaviour for POSCARs in pyiron

ligerzero-ai:
I have no idea if this is alright or not. This is supposed to be a part of the fix for #752. Variable names are free to be changed

I just copied the pull sudarshan did in #492 for direct positioning.

In particular, I am not sure why there are two different places where he needed to place the same setting;
E.g. Why I have to put two blocks of identical code in pyiron_atomistics/vasp/base.py - it seems very strange to me. 

I also don't understand the logic of "allow_reordering=not self.allow_structure_reordering"

And also why in the second block of code, he adjusts the setter to be _write_in_direct_coordinates opposed to the first one where he writes write_in_direct_coordinates. Someone explain to me why this is please, or if it's just an error/typo and not deliberate.

### DIFF
--- a/.ci_support/environment.yml
+++ b/.ci_support/environment.yml
@@ -17,7 +17,7 @@ dependencies:
 - pyiron_base =0.5.26
 - pymatgen =2022.10.22
 - pyscal =2.10.18
-- scipy =1.9.1
+- scipy =1.9.3
 - seekpath =2.0.1
 - scikit-learn =1.1.2
 - spglib =2.0.1
--- a/setup.py
+++ b/setup.py
@@ -42,7 +42,7 @@ setup(
         'pint==0.19.2',
         'pyiron_base==0.5.26',
         'pymatgen==2022.10.22',
-        'scipy==1.9.1',
+        'scipy==1.9.3',
         'seekpath==2.0.1',
         'scikit-learn==1.1.2',
         'spglib==2.0.1',


pyiron_atomistics/vasp/base.py
```diff
@@ -86,6 +86,7 @@ def __init__(self, project, job_name):
         self._potential = VaspPotentialSetter([])
         self._compress_by_default = True
         self.get_enmax_among_species = get_enmax_among_potentials
+        self._allow_structure_reordering = True
```
liamhuber: As far as I can tell, `allow_structure_reordering` gets passed from the job directly to `job.input` and this private variable never gets used anywhere.

```suggestion
```

pyiron_atomistics/vasp/base.py
```diff
@@ -113,6 +114,23 @@ def structure(self, structure):
                 element_lst=structure.get_species_symbols().tolist()
             )
 
+    @property
+    def allow_structure_reordering(self) -> bool:
```
liamhuber: Type hinting for the win.

pyiron_atomistics/vasp/base.py
```diff
@@ -1842,6 +1860,20 @@ def __init__(self):
...
+        """
+
+        Returns:
+            bool
+        """
+        return self.input.allow_structure_reordering
+
+    @allow_structure_reordering.setter
+    def allow_structure_reordering(self, val: bool):
+        self.input.allow_structure_reordering = val
```
liamhuber: `Input` has no attribute `input`, so I suspect this is just a copy+paste error from using the existing code at the job level. One option is to replace `self.input.allow_structure_reordering` with `self._allow_structure_reordering` in the setter and getter. The other is to remove the setter and getter entirely and just make `self.allow_structure_reordering` public in line 1863.

The only reason to keep the setter and getter is to enforce some sort of type checking, so if you choose to keep them please add that. Otherwise they're just a verbose way of doing the same thing.

Going one step further, I can imagine that we may have this sort of meta-input field throughout pyiron -- do we want to give it the same position of authority as `input.kpoints` and `input.incar`? Looking to a future where we have more of this type of flag, perhaps we want to introduce a new intermediate input field, like `input.meta`/`input.expert`/`input.code_specific`/etc? This way we tuck these not-really-input input values out of the way and don't clutter the tab completion. For instance, in VASP I expect we'll quickly get both `job.input.code_specific.allow_structure_reordering` and `job.input.code_specific.write_in_direct_coordinates`. @jan-janssen @samwaseda, thoughts?

pyiron_atomistics/vasp/structure.py
```diff
@@ -108,23 +108,31 @@ def write_poscar(structure, filename="POSCAR", write_species=True, cartesian=Tru
             selec_dyn = True
             cartesian = False
             f.write("Selective dynamics" + endline)
-        sorted_coords = list()
+        coords = list()
```
liamhuber: I like removing the `sorted` now that we have this new functionality. As per your earlier complaint, let's go one step further and just rename this to `positions` right here. Then, e.g., lines like `positions = structure.positions` becomes extremely easy to read and sensible.

pyiron_atomistics/vasp/base.py
```diff
@@ -1867,6 +1899,7 @@ def write(self, structure, modified_elements, directory=None):
             structure,
             filename=posixpath.join(directory, "POSCAR"),
             write_species=not do_not_write_species,
+            allow_reordering=not self.allow_structure_reordering
```
liamhuber: This is bizarre. I also cannot see why a `not` would be necessary here...... maybe test again?

For that matter, it would be excellent to get this new functionality covered in the test suite. If you haven't written unit tests before, it is probably most efficient to just run down another dev at the institute and sit side by side to write this one.

liamhuber (commented):
Nice, looks pretty good!

Only high-level comment is that because the merge is `ligerzero-ai:master->pyiron:master`, I can't find a way to check out the source branch --  thus, all my comments come just from reading the browser content. If your pyiron fork is public, I am probably able to set up a new remote and grab it, but it's extra steps. Since you're a pyiron member and will presumably be contributing regularly, we can streamline this process: instead of a fork just directly clone the source and check out a new branch and push it. Since the rest of the devs already have the pyiron repo set up as a remote source, we can then check out your entire branch super easily.

pyiron_atomistics/vasp/base.py
ligerzero-ai: I think that creating another flag level makes sense for expert users. Although I don't have the authority or the final say on these things. Let's just implement this for now because I am very reluctant to start large-scale segregation calculations without this flag. You could say I'm in a little bit a hurry x)

I'm fond of a name like input.options, it's an informative name and people will know where to look if they're looking for more granular control of the jobs.

pyiron_atomistics/vasp/structure.py
ligerzero-ai: Should be in the latest commit

niklassiemer:
> Nice, looks pretty good!
> 
> Only high-level comment is that because the merge is `ligerzero-ai:master->pyiron:master`, I can't find a way to check out the source branch --  thus, all my comments come just from reading the browser content. If your pyiron fork is public, I am probably able to set up a new remote and grab it, but it's extra steps. Since you're a pyiron member and will presumably be contributing regularly, we can streamline this process: instead of a fork just directly clone the source and check out a new branch and push it. Since the rest of the devs already have the pyiron repo set up as a remote source, we can then check out your entire branch super easily.

If I remember correctly, you may actually check out pull requests to the repository from within pycharm :)

liamhuber:
> If I remember correctly, you may actually check out pull requests to the repository from within pycharm :)

@niklassiemer, thanks!!! You are absolutely right. It's not under the git->branches window (which only shows stuff from origin), but a quick google showed me how to find it under git->github 😄

pyiron_atomistics/vasp/base.py
liamhuber: We're on the same page for the ideal situation then, and I appreciate your need for haste, but haste is dangerous in two ways here:

1) Once we include `vasp.input.allow_structure_reordering` in a release, we have an obligation to provide backwards compatibility to this interface for a while going forward. This is true even if we quickly switch to something like `vasp.input.options.allow_structure_reordering`. This clutters the code and adds maintenance work, so if we can avoid it by choosing a good name and design to start with that is much better.

2) On closer reading I realized your current PR doesn't do what you want. You would have found that you shipped all your jobs off to the queue and they came back reordered anyhow. The reason for this is that `Input.allow_structure_reordering` gets declared `True` in `__init__`, but never gets serialized to HDF. Our execution pattern is to serialize the local object, then re-initialize and deserialize it on the remote python instance. So over on the compute node the job would only ever see the initial value! The correct solution to this is adding tests that ensure that vasp job (de)serialization is working correctly with your new changes.

I made a new PR based off this one, #799, which introduces the input sub-domain discussed above and (I believe) resolves the serialization issues. It still needs tests. My proposal is that you track down someone at MPIE to sit with you and write the tests ASAP, and that we have a quick discussion of how to name `job.input.NAMETHISTHING.allow_structure_reordering` on Monday -- is this fast enough for your timeline?

ligerzero-ai: I think it's fine - I can hack something on a modified module that I'll use for my personal purposes - Putting the pyiron_atomistics folder in the folder where I'm working and modifying that is fine. 

For something that is more sustainable it is probably better that we write a solution that mostly everyone is happy with maintaining going forward

ligerzero-ai: I'll have a look at the tests on the weekend. But honestly the talk of serialisation and all that flies over my head - explain like I'm 5? I will probably need it explained before I know what I'm testing for.

I can probably grab Sam tomorrow to explain, since Marvin is off at a conference.

liamhuber:
> > If I remember correctly, you may actually check out pull requests to the repository from within pycharm :)
> 
> @niklassiemer, thanks!!! You are absolutely right. It's not under the git->branches window (which only shows stuff from origin), but a quick google showed me how to find it under git->github 😄

Aha, @niklassiemer @ligerzero-ai, there is still a nice reason to merge from a branch of pyiron rather than a personal fork: I checked out a new branch from this one, but when I push and create a PR there is no option to merge into this source branch! I could go over to @ligerzero-ai's repo and make a PR there, but that moves the discussion off the pyiron page.

So being able to check out PRs to the repo is rad, but not without edge case inconveniences.

pyiron_atomistics/vasp/base.py
liamhuber: > I'll have a look at the tests on the weekend. But honestly the talk of serialisation and all that flies over my head - explain like I'm 5? I will probably need it explained before I know what I'm testing for.
> 
> I can probably grab Sam tomorrow to explain, since Marvin is off at a conference.

Super.

Yeah, of course! "Serialization" here just means writing to file. Basically when you're working with the job in a notebook, it exists in RAM using your current python instance. To run it on the remote node, we start a new python instance utilizing the node's RAM and re-instantiate the job. That means that the re-instantiated job needs some way to access all the modifications you made from the clean, initialized state (settings changes, etc). These are what's read from file, i.e. "deserialization". So in this case the meat of your test might look something like this:

```python
original_job = pr.atomistics.job.Vasp('testing_serialization')
original_job.input.code_specific_options.allow_structure_reordering = False
# Maybe also play with kpoints, etc. if we don't have any tests for these either
original_job.to_hdf()

loaded_job = pr.atomistics.job.Vasp(original_job.name)
# loaded_job and original_job should now be separate python objects with different hashes
# At this point, if you print out allow_structure_reordering, it will still have the initial value of True
loaded_job.from_hdf()
self.assertEquals(
    original_job.input.code_specific_options.allow_structure_reordering,
    loaded_job.input.code_specific_options.allow_structure_reordering
)
```

I might have made some syntax errors there, and I'm assuming you're inside a `TestCase.TEST_SOMETHING` and have access to a `pr` of type `Project`, etc, but I hope the logic/intent is sufficiently clear.

samwaseda:
https://github.com/pyiron/pyiron_atomistics/blob/1201f799dbe986e082f02ea38b23fc22fa9d5f7e/pyiron_atomistics/sphinx/base.py#L155-L165

I have a bit different suggestion. In SPHInX we have `id_pyi_to_spx` and `id_spx_to_pyi` (v.s.), which gives the permutation. Rather than doing `if ... else` in different parts of the codes (which we'd need for output as well), I'd suggest to do centralise the indices like in SPHInX. Essentially it would be:

```python
@property
def id_pyi_to_dft(self):
    if self.input.option.allow_reorder:
        return self._reordered_indices
    return np.arange(len(self.structure))
```

After this, we would just need to use this whenever we write the input (`self.structure.positions[self.id_pyi_to_dft]`) and parse the output (`self.structure.positions = positions[self.id_dft_to_pyi]`)

ligerzero-ai:
https://github.com/pyiron/pyiron_atomistics/pull/799

Let's continue over there @samwaseda

samwaseda:
> #799
> 
> Let's continue over there @samwaseda

I don't understand the spin off. What's the difference?

ligerzero-ai:
that includes the stuff that includes vasp.input.options etc.

PR #775, opened 2022-09-22
Be ASE compatible wrt spins

pmrv:
ASE calculator objects (in particular GPAW) have some expectations about the spins.  Atoms should raise an AttributeError on accessing `.spins` if none are defined and `get_initial_magnetic_moments` should returns zeros.  If that is not the case then calculators may not reliably detect that no spins are set and attempt calculations involving NaNs which produce weird errors down the line.

With these two small changes the pyiron Atoms object can be used with the GPAW calculator as if it was a normal ASE Atoms.  I assume any other calculator should also be working.

With this change also the `AseJob` may be slimmed down a bit.


pmrv:
Fixes #652 .

pmrv:
FYI these changes have the side effect that to check whether spins have been set on a structure you need to use `structure.has("initial_magmoms")` instead of `not np.isnan(structure.get_initial_magmoms()).all())`, because the latter now always returns zero for unset moments.  That's in line with what ASE atoms does and third party calculators expect.

liamhuber:
> FYI these changes have the side effect that to check whether spins have been set on a structure you need to use `structure.has("initial_magmoms")` instead of `not np.isnan(structure.get_initial_magmoms()).all())`, because the latter now always returns zero for unset moments. That's in line with what ASE atoms does and third party calculators expect.

@pmrv, I was just quickly scrolling through stuff in my feed that looks interesting and I popped in here. Both of those formulations are super ugly to me... I'm a bit out of touch with `Atoms` at the moment, but do you think it would be useful to add a helper class, e.g. `AtomsHas` as `self.has = AtomsHas(self)` in `Atoms.__init__`, so that we can use autocomplete to answer questions like `structure.has.initial_magmom`, `structure.has.magmom`, `structure.has.velocities`, and maybe even `structure.has.cubic_cell` or `structure.has.free_surface`? If we really only need this sort of check for initial magnetic moments then the current `structure.has("initial_magmoms")` is totally fine, but I wanted to see what you think.

pmrv:
@samwaseda It seems that the refactorings uncovered a latent typo that I fixed in [2e88658](https://github.com/pyiron/pyiron_atomistics/pull/775/commits/2e88658b21939d056ef8a689471fd2f181021abe), but now the unit tests fail because some spins are seemingly not correctly written to the structure group.  Can you have a look?  I would like to merge this this week because this blocks #636 .

pmrv:
> > FYI these changes have the side effect that to check whether spins have been set on a structure you need to use `structure.has("initial_magmoms")` instead of `not np.isnan(structure.get_initial_magmoms()).all())`, because the latter now always returns zero for unset moments. That's in line with what ASE atoms does and third party calculators expect.
> 
> @pmrv, I was just quickly scrolling through stuff in my feed that looks interesting and I popped in here. Both of those formulations are super ugly to me... I'm a bit out of touch with `Atoms` at the moment, but do you think it would be useful to add a helper class, e.g. `AtomsHas` as `self.has = AtomsHas(self)` in `Atoms.__init__`, so that we can use autocomplete to answer questions like `structure.has.initial_magmom`, `structure.has.magmom`, `structure.has.velocities`, and maybe even `structure.has.cubic_cell` or `structure.has.free_surface`? If we really only need this sort of check for initial magnetic moments then the current `structure.has("initial_magmoms")` is totally fine, but I wanted to see what you think.

It's mostly necessary for the magnetic moments, but the `has` method comes from the ASE class, so I'd rather not overwrite it.

liamhuber:
> It's mostly necessary for the magnetic moments, but the `has` method comes from the ASE class, so I'd rather not overwrite it.

If you like the idea we don't need to overwrite it, we can overload it:

```python
from ase import Atoms
import numpy as np

class AtomsHas:
    def __init__(self, atoms: Atoms):
        self.atoms = atoms

    @property 
    def initial_magmom(self) -> bool:
        return self.structure.has("initial_magmoms")

    @property 
    def cubic_cell(self) -> bool:
        cell = self.atoms.cell.array
        return np.allclose(cell, np.eye(3) * np.diag(cell))  # Or whatever the right math is

    # etc...

    def __call__(self, string: str) -> bool:
        # Also with functools.wraps to get the docstring
        return self.atoms.has(string)
```

samwaseda:
Tests run.

pmrv:
> Tests run.

Thanks for the help!

PR #784, opened 2022-09-28
Create general input output for `AtomisticGenericJob`

samwaseda:
Centralize `to_hdf` and `from_hdf` in `AtomisticGenericJob`.

### DIFF
--- a/pyiron_atomistics/atomistics/job/atomistic.py
+++ b/pyiron_atomistics/atomistics/job/atomistic.py
@@ -135,9 +135,14 @@ class AtomisticGenericJob(GenericJobCore, HasStructure):
         self.__version__ = "0.1"
         self._structure = None
         self.input = DataContainer(table_name="input")
-        self._generic_input = GenericInput()
         self.output = GenericOutput(job=self)
         self.map_functions = MapFunctions()
+
+    @property
+    def _generic_input(self):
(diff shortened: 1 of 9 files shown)


niklassiemer:
Why in `AtomisticGenericJob` and not in `GenericJob`?

samwaseda:
> Why in `AtomisticGenericJob` and not in `GenericJob`?

Because the idea of storing everything inside `self.input` and `self.output` is currently not compatible with `ParallelMaster`. When the implementation is done for `AtomisticGenericJob` we can move forward to `GenericJob`

samwaseda:
So this story turned out to be more complicated, as `AtomisticsParallelMaster` was actually derived from `AtomisticGenericJob`... Is there a possibility to set priority for the parent classes?

pmrv:
You can use `super(Parent, self).method` to call `method` that is defined on `Parent` or any of it's super classes, effectively allowing you to skip the direct superclass(es) of `type(self)`.  But I'm not sure what you're trying to do.

I've found (older) explanation [here](https://rhettinger.wordpress.com/2011/05/26/super-considered-super/) which made me realize, we actually use it a bit wrong especially in the case of job classes.  Iirc Ray had a talk on youtube on this topic also.

PR #791, opened 2022-09-30
Trajectory: ensure remapped_indices is always an array

### LINKED ISSUE #605: error in animate_structure() when displaying multiple species
**Issue**
when I try to animate certain atoms by specifying atom_indices in animate_structure(), I run into the error "TypeError: list indices must be integers or slices, not tuple"

The code I use is
```
show_list = np.array([36, 37, 128,  206])
job.animate_structure(atom_indices = show_list)
```

The error I got is 
```
in AtomisticGenericJob.trajectory(self, stride, center_of_mass, atom_indices, snapshot_indices, overwrite_positions, overwrite_cells)
    619 else:
    620     remapped_indices = indices
    622 return Trajectory(
    623     positions[::stride, atom_indices, :],
    624     sub_struct,
    625     center_of_mass=center_of_mass,
    626     cells=cells[::stride],
--> 627     indices=remapped_indices[::stride, atom_indices],
    628 )

TypeError: list indices must be integers or slices, not tuple
```

**Observation**

My code actually works when I specify show_list as np.array([36, 37, 128]), np.array([36, 37, 206]), or np.array([36, 37, 206, 207]). In this case atom 36 & 37 are hydrogens, 128 is Mg, and 206 & 207 are O, so it seems the error only occurs when more than two types of elements are involved. I think this might be the key for reproducing this error.

pmrv:
Might fix #605.

### DIFF
--- a/pyiron_atomistics/atomistics/job/atomistic.py
+++ b/pyiron_atomistics/atomistics/job/atomistic.py
@@ -628,7 +628,7 @@ class AtomisticGenericJob(GenericJobCore, HasStructure):
                 sub_struct,
                 center_of_mass=center_of_mass,
                 cells=cells[::stride],
-                indices=remapped_indices[::stride, atom_indices],
+                indices=np.asarray(remapped_indices)[::stride, atom_indices],
             )
 
     def write_traj(


jyang2009:
Dear Marvin,

I just checked this branch. The bug seems to be fixed. Thanks a lot. I think we should merge it with the master branch. 

Best,
Jing

pmrv:
Thanks for testing!

PR #802, opened 2022-10-08
adds restart from contcar option to VASP jobs

ligerzero-ai:
Adds restart from contcar option to vasp jobs; one of the most common types of restarts performed by users in vasp


ligerzero-ai:
restart() inherits the last known structural configuration in the super() call. solution: update the docstring in the restart() in the vasp job class to say that it's essentially copying CONTCAR -> POSCAR

PR #812, opened 2022-10-11
Bugfixes in StructurePlot and error handling for Atoms.get_symmetry

pmrv:
There were some small typos in StructurePlots.  Additionally I found that for "weird" structures spglib can run into internal errors when calling get_symmetry.  This then produced unrelated errors in the Symmetry class, because we're not expecting spglib to fail.  So I've added a simple error class and `None` checks.


samwaseda:
It looks good to me. Can you maybe add a test? The symmetry search typically fails when there are two atoms on top of each other, so if you create a structure by:

```python
structure = pr.create.structure.bulk('Al', cubic=True)
structure += structure[-1]
structure.get_symmetry()
```

it should fail.

PR #815, opened 2022-10-12
Add warning that Sphinx.calc_minimize does not support pressure

pmrv:
Sphinx does not allow volume relaxations and while the code does throw an error if you pass `pressure`, the docstring doesn't mention this, so I've added a small warning to it.

@samwaseda and I also talked about this a bit from a more general perspective.  I think the current `AtomisticGenericJob.calc_minimize` is too generic.  It promises functionality that not all codes can deliver.  That makes it very difficult to program against `AtomisticGenericJob` as an interface.  He proposed to remove pressure related options from `Sphinx` jobs, but I feel that doesn't solve the problem, i.e. code written against `AtomisticGenericJob` would still fail when passed a `Sphinx` job.  A more intrusive option would be to remove the pressure related parameters from `AtomisticGenericJob`, but allow subclasses to add parameters of their own and then do that in `Vasp` and `Lammps` (the only codes I know that actually can do pressure stuff).  Thoughts?

### DIFF
--- a/pyiron_atomistics/sphinx/base.py
+++ b/pyiron_atomistics/sphinx/base.py
@@ -691,6 +691,10 @@ class SphinxBase(GenericDFTJob):
         Loads defaults for all SPHInX input groups, including a
         ricQN-based main Group.
 
+        .. warning::
+            Sphinx does not support volume minimizations!  Calling this method with `pressure` or `volume_only` results
+            in an error.
+
         Args:
             retain_electrostatic_potential:
             retain_charge_density:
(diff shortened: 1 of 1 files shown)


samwaseda:
If I remember well @liamhuber had the idea of totally separating structure optimisation and MD from the atomistic engines. And I guess at that time he called them plug-ins. This is slightly similar to the idea of wrapper, although the hierarchy probably differs a little bit.

samwaseda:
Actually this is something that we should talk about at the pyiron meetings at some point.

PR #816, opened 2022-10-12
Allow to configure strains used in Murnaghan

pmrv:
You can set a list now with strains to calculate directly on the Murnaghan object.  This allows for asymmetric and non-uniform sampling of the E-V curve.

### DIFF
--- a/pyiron_atomistics/atomistics/master/murnaghan.py
+++ b/pyiron_atomistics/atomistics/master/murnaghan.py
@@ -3,7 +3,7 @@
 # Distributed under the terms of "New BSD License", see the LICENSE file.
 
 from __future__ import print_function
-from typing import List, Optional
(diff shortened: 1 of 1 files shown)


pmrv:
Annoyingly, it *has* to be a list (or tuple).  An array won't work, because that borks `GenericParameters`.

samwaseda:
> Annoyingly, it _has_ to be a list (or tuple). An array won't work, because that borks `GenericParameters`.

Shouldn't `GenericParameters` be able to translate numpy arrays into lists? Sounds like an easy fix on `__setitem__`

jan-janssen:
This pull request seem to have broken the `QuasiHarmonicJob` https://github.com/pyiron/pyiron_atomistics/blob/main/pyiron_atomistics/atomistics/master/quasi.py

PR #819, opened 2022-10-13
Ipywidgets are only required when building notebooks

jan-janssen:
By adding `ipywidgets` as dependency in our build environments in https://github.com/pyiron/pyiron_atomistics/pull/729 it is installed for all tests, while before it was only installed when the notebooks were build - this is fixed now.


jan-janssen:
For some reason `ipywidgets` are still installed so it seems one of the requirements is installing it.

PR #820, opened 2022-10-13
Release dependence on ipywidgets

jan-janssen:
Checklist:
* [ ] Wait for nglview to get a [conda release](https://github.com/conda-forge/nglview-feedstock/pull/57) that is either ipywidgets 8-compliant, or does the pinning for us


liamhuber:
You need this pin for nglview to work. The problem is that nglview is inconsistent with the latest version of ipywidgets, and while [they have merged Marvin's patch that pins the dependency there](https://github.com/nglviewer/nglview/pull/1033), there has been no new release of nglview since then. So what happens is that conda sees our nglview dependency and installs the latest version, then sees nglview's dependency on ipywidgets and installs *the latest version of that* and voila, we have our error.

They have a [couple](https://github.com/nglviewer/nglview/issues/1035) [issues](https://github.com/nglviewer/nglview/issues/1035) on this error already. I'm don't know exactly what they're release process looks like, but I just learned now that they have a [conda feedstock PR](https://github.com/conda-forge/nglview-feedstock/pull/57) since August, but as of September apparently [no one has time to debug it](https://github.com/nglviewer/nglview/issues/1032#issuecomment-1245473411).

So for now we need the pin.

jan-janssen:
> So for now we need the pin.

That is why I keep the pull request a draft.

hainm:
> but as of September apparently https://github.com/nglviewer/nglview/issues/1032#issuecomment-1245473411.

yay. Sorry about this. We need someone volunteering working on that. 
PS: If someone would like to handle the pip release, please let me know so I will ping Alexander Rose to add your pypi account as one of the maintainers. cheers.

PR #832, opened 2022-10-22
Properly parse energy states and occupations

pmrv:
Due to a faulty regex previously we threw away the signs of both energy states and occupations parsed from sphinx logs.

### DIFF
--- a/pyiron_atomistics/sphinx/base.py
+++ b/pyiron_atomistics/sphinx/base.py
@@ -2122,9 +2122,7 @@ class _SphinxLogParser:
         return self._n_steps
 
     def _parse_band(self, term):
-        arr = np.loadtxt(
-                re.findall(term, self.log_main, re.MULTILINE)
-        )
+        arr = np.loadtxt(re.findall(term, self.log_main, re.MULTILINE))
         shape = (-1, len(self.k_points), arr.shape[-1])
         if self.spin_enabled:
             shape = (-1, 2, len(self.k_points), shape[-1])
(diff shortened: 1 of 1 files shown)


pyiron_atomistics/sphinx/base.py
```diff
@@ -2124,7 +2124,7 @@ def n_steps(self):
     def _parse_band(self, term):
         fa = re.findall(term, self.log_main, re.MULTILINE)
         arr = (
-            np.array(re.sub("[^0-9\. ]", "", "".join(fa)).split())
+            np.array(re.sub("[^-0-9\. ]", "", "".join(fa)).split())
```
niklassiemer: Now this only works on negative quantities?!

niklassiemer: wouldn't this allow, but not require a `-` sign? (Not tested!)
```suggestion
            np.array(re.sub("[^-?0-9\. ]", "", "".join(fa)).split())
```

pmrv: No, this regex matches everything `[]` that is not `^` a `-`, `0-9`, `\.` (plain dot, `\` is for escaping) or ` ` and replaces it by the empty string.  This leaves just numbers and spaces, which are then "parsed".  Because the regex previously also matched the `-` the parsed number were effectively only the absolute value.

Actually the whole regex business is too complicated, because in the sphinx log the lines are of the form `some text: number number number`.  So removing that prefix would be enough.

niklassiemer: Well in that case, it souds like an easier and more stable solution to replace everything until a colon (being greedy it will be the last colon)  by the empty string and parsing the numbers. Right now, a single `-` in the string would again flip the sign of the first number...

Issue #831, opened 2022-10-20
`TrainingContainer` cannot be pickled


samwaseda:
```python
container = pr.create.job.TrainingContainer('container')
for job in pr.iter_jobs():
    container.include_job(job)
df = container.to_pandas()
df.to_pickle('name_of_the_file.pkl')
```

This fails, because the `Atoms`-object contains a mysterious `None` towards the beginning (i.e. `df.drop(columns=['atoms'].to_pickle('blablabla.pkl')` works).

pmrv:
Quick fix:

```python
container = pr.create.job.TrainingContainer('container')
for job in pr.iter_jobs():
    container.include_job(job)
df = container.to_pandas()
df.atoms = df.atoms.map(lambda a: a.to_ase())
df.to_pickle('name_of_the_file.pkl')
```

If I recall correctly it's already not possible to pickle our atoms, see #643 

You can `pack`/`unpack` projects with the training containers inside, if you need it for the workshop.

PR #835, opened 2022-10-23
Make Atoms pickle-able

### LINKED ISSUE #643: Atoms cannot be pickled
Reminder for myself: ASE atoms can be pickled, but ours appear not to.

pmrv:
Previous problems with {,un}pickling Atoms were caused by our unclean overriding of `__getattr__`.  pickle (copy presumably as well) relies on this method to discover where the object in question defines methods that customize this behaviour  (`__reduce_ex__` and `__setstate__`).

That means first that returning `None` indiscriminately on undefined keys from `__getattr__` will break this search, because pickle will try to call `None`s.

Second `__setstate__` will be called on an object just after `__new__` (on copied and unpickled object `__init__` is not called at all).  So when `__getattr__` relies on attributes itself that are only defined in `__init__` this will lead to an infinite regression.

Both of these problems appear on the `ChemicalElements` and `PeriodicTable` classes as used inside `Atoms` and also with `Atoms` itself.  For the first two I have simply removed `__getattr__` and adjusted down-stream code because users are not (or rarely) interacting with it and the additional headache is not worth it, just to cut down some typing.  For `Atoms` accessing tags as attributes is part of the interface, so I have moved those attributes (_tag_list) that cause the regress to be initialized in `__new__`.

Fixes #643, #831.

I'm on vacation for two weeks from tomorrow, so I'll let someone else write a `pickle.loads(pickle.dumps(Atoms(...)))` unit test.

### DIFF
--- a/pyiron_atomistics/atomistics/structure/periodic_table.py
+++ b/pyiron_atomistics/atomistics/structure/periodic_table.py
@@ -153,10 +153,10 @@ class ChemicalElement(object):
         Args:
             hdf (Hdfio): Hdfio object which will be used
         """
-        with hdf.open(self.Abbreviation) as hdf_el:  # "Symbol of the chemical element"
+        with hdf.open(self["Abbreviation"]) as hdf_el:  # "Symbol of the chemical element"
             # TODO: save all parameters that are different from the parent (e.g. modified mass)
-            if self.Parent is not None:
-                self._dataset = {"Parameter": ["Parent"], "Value": [self.Parent]}
+            if self["Parent"] is not None:
+                self._dataset = {"Parameter": ["Parent"], "Value": [self["Parent"]]}
                 hdf_el["elementData"] = self._dataset
             with hdf_el.open(
                 "tagData"
@@ -251,7 +251,7 @@ class PeriodicTable(object):
                 if not new_element.tags["sub_tags"]:
                     del new_element.tags["sub_tags"]
 
-            if new_element.Parent is None:
+            if new_element["Parent"] is None:
                 if not (el in self.dataframe.index.values):
                     raise AssertionError()
                 if len(new_element.sub["tags"]) > 0:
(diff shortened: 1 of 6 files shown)


pyiron_atomistics/atomistics/structure/periodic_table.py
```diff
@@ -89,6 +84,12 @@ def __getitem__(self, item):
         if item in self.sub.index:
             return self.sub[item]
 
+    def get(self, item, default=None):
+        value = self[item]
```
niklassiemer, line 88: Shouldn't this be wrapped in try/except? If the `self[key]` raises an error I would expect `get(key, ANY)` to return `ANY`.

pyiron_atomistics/atomistics/structure/periodic_table.py
```diff
@@ -214,15 +215,18 @@ def __init__(self, file_name=None):  # PSE_dat_file = None):
...
-        return self[item]
-
     def __getitem__(self, item):
         if item in self.dataframe.columns.values:
             return self.dataframe[item]
         if item in self.dataframe.index.values:
             return self.dataframe.loc[item]
 
+    def get(self, item, default=None):
+        value = self[item]
```
niklassiemer, line 225: Same here, errors of `self[item]` should be cought if a default is provided.

niklassiemer (commented):
Looks good on the first glance! If it survives the pickle/unpickle test, I approve :)

Thanks a lot and enjoy your holiday!

niklassiemer:
Well of course we also have to sort out the elements saving/loading which is not working correctly, yet.

pyiron_atomistics/atomistics/structure/periodic_table.py
pmrv: In both `ChemicalElement` and `PeriodicTable` `__getitem__` never raises an exception, but returns `None` on undefined keys.

pmrv:
There were still places that used attribute access instead of item access which were causing the problems.  The remaining failing unit test is due to `ChemicalElement` not implementing `__setitem__`.  It also previously didn't implement `__setattr__`, but the default implementation made it seem to work.  Properties of elements modified this way would have never been saved however.  Is this a feature that is used at all?   If not I would say we just cut the test.

niklassiemer:
Well it was not working before... thus I would probably cut the test.

liamhuber:
@pmrv merging main in was not fast-forward, so I'll leave it to you.

pmrv:
Fixed elsewhere.

PR #853, opened 2022-11-02
Adding `animate` method to Trajectory

prince-mathews:
- Within the ´animate´ method inside the Trajectory class, it wasn't possible to include ´stride´, ´atom_indices´, ´snapshot_indices´ or ´repeat´ arguments.

### DIFF
--- a/pyiron_atomistics/atomistics/job/atomistic.py
+++ b/pyiron_atomistics/atomistics/job/atomistic.py
@@ -370,16 +370,8 @@ class AtomisticGenericJob(GenericJobCore, HasStructure):
         )
         if repeat is not None:
             traj = traj.transform(lambda s: s.repeat(repeat))
-        animation = nglview.show_asetraj(traj)
-        if spacefill:
-            animation.add_spacefill(radius_type="vdw", scale=0.5, radius=particle_size)
-            animation.remove_ball_and_stick()
-        else:
-            animation.add_ball_and_stick()
-        if show_cell:
-            if self.structure.cell is not None:
-                animation.add_unitcell()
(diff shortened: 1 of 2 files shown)


pmrv (approved):
Looks good to me, thanks!

pmrv:
You had also code for the creation of a `Trajectory` from a list of structures, right?  Can you add that as well as a method to `Trajectory`?  Something like

```python
class Trajectory:
  ...
  @classmethod
  def from_structure_list(cls, structures):
     # somehow get all positions, etc. from all structures
     return cls(positions, ...)
```

prince-mathews:
> You had also code for the creation of a `Trajectory` from a list of structures, right? Can you add that as well as a method to `Trajectory`? Something like
> 
> ```python
> class Trajectory:
>   ...
>   @classmethod
>   def from_structure_list(cls, structures):
>      # somehow get all positions, etc. from all structures
>      return cls(positions, ...)
> ```

Isn't this something the Trajectory class does by itself?
I don't have a code, I just provide the Trajectory class the initial structure, positions of the structures I want to see in the trajectory and the Trajectory is created by the class itself.

pmrv:
Mh, I was thinking really turning a list of `Atoms` objects into a trajectory.  But I remember now that the `Trajectory` class only works with "canonical" structures (equal number of particles everywhere).  I'll add some stuff to this branch.

pmrv:
I've added some code to move `animate` over to `HasStructure`.  That means that a lot more jobs can now be animated (StructureContainer/TrainingContainer/StructureStorage/TrainingStorage/etc.).

PR #880, opened 2022-11-23
[new] vasp options

### LINKED ISSUE #752: Pyiron generated VASP inputs: Option to preserve atomic ordering

ligerzero-ai:
Supercedes/obsoletes #799 

https://github.com/pyiron/pyiron_atomistics/tree/vasp_options rebase was being a major pain (wouldn't recognise traitlets being present on the new base version) so I will just implement on this branch with a fresh base. The issue was failing unittests because of an old version of base, which I couldn't rebase around to get it to pass.


ligerzero-ai:
Okay, ready for review (and really to be merged). Either works for me, but I'd prefer sooner rather than later.

pyiron_atomistics/table/funct.py
```diff
@@ -95,9 +95,9 @@ def get_volume_per_atom(job):
 
 
 def get_elements(job):
-    species = job['input/structure/species']
-    indices = job['input/structure/indices']
-    return {s: sum(indices==i) for i, s in enumerate(species)}
+    species = job["input/structure/species"]
```
liamhuber: This looks like black formatting on an unrelated function. It could be that black just never got run on this snippet until now, but could you please merge master into this branch to make sure we have the most up-to-date diff?

pyiron_atomistics/vasp/base.py
```diff
@@ -113,6 +128,36 @@ def structure(self, structure):
                 element_lst=structure.get_species_symbols().tolist()
             )
 
+    # DEPRECATE THIS WITH VASP_SORTER IN STRUCTURE.PY
```
liamhuber, line 131: Deprecations warnings and notes should be handled with `pyiron_base.Deprecate`/`deprecate`/`deprecate_soon`

pyiron_atomistics/vasp/base.py
```diff
@@ -113,6 +128,36 @@ def structure(self, structure):
...
+    @sorted_indices.setter
+    def sorted_indices(self, val):
+        """
+        Setter for the sorted indices
+        """
+        self._sorted_indices = val
+
+    @property
+    def _idx_pyiron_to_user(self):
+        if self.input._idx_pyiron_to_user is None:
```
liamhuber, line 151: This is a big PR so right now just consider this a note to myself, and after I've read the rest and better understand what's happening I'll try and remember to come back here and update the context.

At the moment it's a bit of a yellow flag to me that accessing a property is doing a silent and implicit variable update.

Edit: Ok, coming back at the end and I'm afraid I am still not confident I understand all the side effects that this might have. Doesn't mean it can't be done or isn't a good idea, just that I'm not going to give it my official blessing atm 😜

pyiron_atomistics/vasp/base.py
```diff
@@ -370,8 +399,10 @@ def write_input(self):
                     self.logger.info(
                         "The POSCAR file will be overwritten by the CONTCAR file specified in restart_file_list."
                     )
+        # Structure being fed into the .write() function is sorted, that's why write()'s structure input is not sorted
```
liamhuber: This is unclear. The structure being fed in is already sorted, or write sorts structures it is fed?

It looks to me like we are passing a "sorted" structure to `.write` by virtue of the fact we use the `self._idx_user_to_pyiron` map.

pyiron_atomistics/vasp/base.py
```diff
@@ -386,12 +417,10 @@ def collect_output(self):
                 self.structure = self.get_final_structure_from_file(filename="CONTCAR")
             except IOError:
                 self.structure = self.get_final_structure_from_file(filename="POSCAR")
-            self._sorted_indices = np.array(range(len(self.structure)))
+            self._idx_pyiron_to_user = np.array(range(len(self.structure)))
```
liamhuber:
```suggestion
            self._idx_pyiron_to_user = np.arange(len(self.structure), dtype=int)
```

Should be equivalent and more direct.

But also, really? Maybe I'm just being daft here, but I don't understand how [0,...,n] can always be the pyiron-to-user map. This map should be the inverse map of user-to-pyiron and that can be pretty much arbitrary depending on species order, so how is the reverse map always this simple?

Edit: Also came back to this at the end of the review. I expanded the context and see that this is happening when we didn't have a structure originally and need to read it from file. So is this just some sort of special case that only gets hit when we're importing an externally executed vasp job or something? That's a condition under which I could see setting the map like this, since pyiron doesn't have any prior idea what other order the atoms could have been in, so what's read in is good enough.

pyiron_atomistics/vasp/base.py
```diff
@@ -777,13 +817,20 @@ def get_final_structure_from_file(self, filename="CONTCAR"):
             pyiron.atomistics.structure.atoms.Atoms: The final structure
         """
         filename = posixpath.join(self.working_directory, filename)
+        # Backwards compatibility for old datasets using old vasp_sorter method
+        # DEPRECATE THIS WITH VASP_SORTER IN STRUCTURE.PY
+        try:
+            _idx_pyiron_to_user = self._idx_pyiron_to_user
+        except:
```
liamhuber:
```suggestion
        except AttributeError:
```

It's good form to be as targeted as possible when catching exceptions

pyiron_atomistics/vasp/base.py
```diff
@@ -672,8 +707,13 @@ def from_directory(self, directory):
             else:
                 raise ValueError("Unable to import job because structure not present")
             self.structure = structure
-            # Always set the sorted_indices to the original order when importing from jobs
-            self.sorted_indices = np.arange(len(self.structure), dtype=int)
+            # Always set the _idx_pyiron_to_user to the original order (unsorted) when importing from jobs
+            try:
+                self._idx_pyiron_to_user = np.arange(len(self.structure), dtype=int)
+            except:
```
liamhuber: I'm struggling a bit to see when this exception will/won't be thrown. As soon as someone's running this version of the code, every Vasp job instance should have the read-only property `self._idx_pyiron_to_user`, and thus the try will fail with an `AttributeError`. Is this correct? If it is *always* failing, what is it that we're actually trying to accomplish with the try/except and can we do it in a more meaningful way? It's possible this bit is just fine and I'm misunderstanding things, but I might need some hand-holding here.


```suggestion
            except AttributeError:
```
It's good form to be as targeted as possible when catching exceptions

pyiron_atomistics/vasp/base.py
```diff
@@ -1817,6 +1864,25 @@ def __del__(self):
...
+
+    Traits:
+        allow_structure_reordering (bool): Allows pyiron to reorder structures to minimize POTCAR sizing
+            (e.g. Fe37 P1 Fe35 -> Fe72 P1 reduces POTCAR by one Fe POTCAR filesize. (Default is True.)
+    """
+
+    def __init__(self, group_name="advanced_options"):
+        super().__init__(group_name=group_name)
+
+    allow_structure_reordering = Bool()
```
liamhuber: Just nit-picking the UX here, would `use_structure_reordering` be a more accurate name here? (I recognize with full humility I might be the one who originally suggested the current name!) To me "allow" implies that there is some user action that is now _permissible_, but I guess in this case we are turning on or off an automated process.

I guess the counter-point is that this automated process only happens some of the time, i.e. it's allowed to happen, but I still think I lean to "use" as slightly more descriptive.

pyiron_atomistics/vasp/base.py
```diff
@@ -1843,15 +1909,70 @@ def __init__(self):
...
+            self._map_pyiron_to_user_idx()
+        return self.__idx_pyiron_to_user
+
+    def _map_pyiron_to_user_idx(self):
+        """
+        This writes the indices maps for user->pyiron (_idx_user_to_pyiron), and pyiron->user (_idx_pyiron_to_user)
+        This looks in Input.options for the allow_structure_reordering boolean value, and decides to return either
+        1. Sorting map when species-based reordering is allowed (default behaviour of pyiron)
+        or
+        2. Sorting map when sorting is forbidden, in which case naive map (1:1, 2:2 etc.) is returned.
```
liamhuber:
```suggestion
        2. Sorting map when sorting is forbidden, in which case naive map (0:0, 1:1 etc.) is returned.
```

pythonic indexing

pyiron_atomistics/vasp/base.py
```diff
@@ -1843,15 +1909,70 @@ def __init__(self):
...
+        """
+        if self.options.allow_structure_reordering:
+            atom_numbers = self.structure.get_number_species_atoms()
+
+            _idx_user_to_pyiron = list()
+            for species in atom_numbers.keys():
+                indices = self.structure.select_index(species)
+                for i in indices:
+                    _idx_user_to_pyiron.append(i)
+            self.__idx_user_to_pyiron = np.array(_idx_user_to_pyiron)
```
liamhuber, line 1953:
```suggestion
            self.__idx_user_to_pyiron = np.array(
                [
                    self.structure.select_index(species) for species in atom_numbers.keys()
                ],
                dtype=int
            ).flatten()
```

So this is not a hill I want to die on, and I haven't done testing to be 100% sure it's identical -- but I don't find this formulation any harder to read (maybe even a bit easier for someone who's familiar with numpy), and I'm pretty sure it's faster. @pmrv thoughts?

pyiron_atomistics/vasp/base.py
```diff
@@ -1843,15 +1909,70 @@ def __init__(self):
...
+
+            _idx_pyiron_to_user = np.array([0] * len(_idx_user_to_pyiron))
+            for i, p in enumerate(_idx_user_to_pyiron):
+                _idx_pyiron_to_user[p] = i
```
liamhuber, line 1957: This can be done without the for-loop

```suggestion
            n = len(_idx_user_to_pyiron)
            _idx_pyiron_to_user = np.empty(n, dtype=int)
            _idx_pyiron_to_user[_idx_user_to_pyiron] = np.arange(n)
```

I don't think we need a `dtype=int` in the arange, but maybe double check `_idx_pyiron_to_user` is still int-like.

pyiron_atomistics/vasp/base.py
```diff
@@ -1843,15 +1909,70 @@ def __init__(self):
...
+
     def write(self, structure, modified_elements, directory=None):
         """
         Writes all the input files to a specified directory
 
         Args:
-            structure (atomistics.structure.atoms.Atoms instance): Structure to be written
+            structure (atomistics.structure.atoms.Atoms instance): Structure (unsorted) to be written
+
+            The structure obj being fed into this fn at .write_input() function is sorted;
```
liamhuber: Three yellow flags:
- This documentation makes assumptions about the behaviour of another function. What's more, it makes assumptions about a function _that calls this function_, which I feel is very easy to lose track of. In contrast, if some of this same discussion were in `write_input`, at least there we'd be discussing assumptions about a function we're calling.
- The bit about "if you call it manually and want to be consistent with `write_input` is good, but I see we use a private variable. Should we make `_idx_X_to_Y` public?
- I don't think we need abbreviations like "obj" and "fn" in the docs.

pyiron_atomistics/vasp/base.py
```diff
@@ -1961,19 +2085,24 @@ def structure(self, atoms):
...
-            sorted_indices = vasp_sorter(self.structure)
         files_present = os.listdir(directory)
         log_dict = dict()
         vasprun_working, outcar_working = False, False
+        # DEPRECATE THIS WITH VASP_SORTER IN STRUCTURE.PY
+        # First attempt to extract indices maps from the job
+        try:
+            # If it succeeds, just use it for output parsing
+            _idx_pyiron_to_user = self._job._idx_pyiron_to_user
+        except:
```
liamhuber: Don't catch naked exceptions, blah blah. I guess this is an `AttributeError` again, or?

pyiron_atomistics/vasp/potential.py
```diff
@@ -418,10 +418,16 @@ def _set_default_path_dict(self):
...
-        object_list = self._structure.get_species_objects()
+
+        prev_element = self._structure.elements[0]
+        object_list = [prev_element]
+        element_list = [prev_element.Abbreviation]
+        for i, element in enumerate(self._structure.elements):
+            if element != prev_element:
+                object_list.append(element)
+                element_list.append(element.Abbreviation)
+                prev_element = element
```
liamhuber, line 429:
```suggestion
        object_list = [
            e for i, e in enumerate(self._structure.elements) 
            if i==0 or self._structure.elements[i-1] != e
        ]
        element_list = [e.Abbreviation for e in object_list]
```

pyiron_atomistics/vasp/structure.py
```diff
@@ -98,33 +99,49 @@ def write_poscar(structure, filename="POSCAR", write_species=True, cartesian=Tru
         for a_i in structure.get_cell():
             x, y, z = a_i
             f.write("{0:.15f} {1:.15f} {2:.15f}".format(x, y, z) + endline)
-        atom_numbers = structure.get_number_species_atoms()
+
+        # This section generates the species string and count of the POSCAR
```
liamhuber, line 104: Instead of a comment, I suggest to refactor this out to a separate, well-named method

pyiron_atomistics/vasp/structure.py
```diff
@@ -322,6 +339,17 @@ def vasp_sorter(structure):
     """
     Routine to sort the indices of a structure as it would be when written to a POSCAR file
 
+    ######################################################################################################
```
liamhuber, line 348: `vasp_sorter` should be decorated with `pyiron_base.deprecate`

tests/vasp/test_vasp.py
```diff
@@ -448,16 +447,103 @@ def test_potentials(self):
...
+                        "Structure sorting should be True by default in job.input.options.allow_structure_reordering!")
+        # Test 2: The length of the indices maps are equal length to the structure.
+        self.assertEqual(len(job_idx_map._idx_pyiron_to_user), len(job_idx_map.structure),\
+                         "The created indices maps (job._idx_pyiron_to_user) are not equal in length to structure!")
+        self.assertEqual(len(job_idx_map._idx_user_to_pyiron), len(job_idx_map.structure),\
+                         "The created indices maps (job._idx_user_to_pyiron) are not equal in length to structure!")
+        # Test 3: The indices maps in event of sorting (due to how test is configured) must not be equal as default behaviour is to sort
+        # This causes the atom of different species in the middle to be shuffled to the ends of the file, so _idx_user_to_pyiron != _idx_pyiron_to_user
+        self.assertFalse(np.array_equal(job_idx_map._idx_pyiron_to_user, job_idx_map._idx_user_to_pyiron),\
+                         "job._idx_user_to_pyiron and job._idx_pyiron_to_user should not be equal in the sorted test")
```
liamhuber: Instead of using a large docstring, you can exploit multiple tests and `self.subTest` to structure your tests so that what is now in the docstring/comments is mostly clear from the code itself. The other advantage to breaking your tests apart is that then if one piece fails, the rest will still run -- this often lets you debug multiple problems at once instead of "run; crash; debug; run; crash a little farther on; etc...". This is true both of separate test functions, and of different clauses in `self.subTest`. The `subTest` object takes a message string as its first argument, so you can basically just move a bunch of your documentation right there. For separate test functions, the name itself should cover most of the info, but copying the relevant part of your larger docstring into the new, smaller function's docstring is ok too.

You will probably know better than I where good break points are, but at a first pass I suggest:
- `def test_sorting_maps` (1, 2, 3)
- `def test_writing` (4)
- `def test_reading` (5, 6)

And then exploiting `subTest` wherever your comments indicate an "(a)", "(b)", etc.

Concretely, breaking out part 1 might look like this:

```suggestion
def test_sorting_maps(self):
    job_idx_map = self.project.create_job("Vasp", "job_idx_maps")
    struct = self.project.create.structure.bulk("Fe") * [np.random.randint(5, 100), 1, 1]
    # Put another species in the middle somewhere in user-defined structure
    struct[3] = "Al"
    job_idx_map.structure = struct
    job_idx_map.input.incar["KSPACING"] = 0.5
        
    with self.subTest("Default behaviour should be to allow sorting"):
        self.assertTrue(job_idx_map.input.options.allow_structure_reordering,\
                        "Structure sorting should be True by default in job.input.options.allow_structure_reordering!")

    with self.subTest("Map lengths should match the structure"):
        self.assertEqual(len(job_idx_map._idx_pyiron_to_user), len(job_idx_map.structure),\
                         "The created indices maps (job._idx_pyiron_to_user) are not equal in length to structure!")
        self.assertEqual(len(job_idx_map._idx_user_to_pyiron), len(job_idx_map.structure),\
                         "The created indices maps (job._idx_user_to_pyiron) are not equal in length to structure!")

    with self.subTest("Maps should not be their own inverse when actual sorting occurs"):
        self.assertFalse(np.array_equal(job_idx_map._idx_pyiron_to_user, job_idx_map._idx_user_to_pyiron),\
                         "job._idx_user_to_pyiron and job._idx_pyiron_to_user should not be equal in the sorted test")
```

liamhuber (commented):
Nice. I covered it in red ink, but it's mostly pretty specific stuff and I don't see any major issues.

The bit about implicitly re-assigning the `input.structure` every time the map is updated is the only part that might keep me up at night, the rest mostly consistency, QoL, readability, and efficiency suggestions. I also appreciate that you took the time to patch some typos and some of the outdated syntax that was being used in these files.

Edit: Oh, and the notebooks test should be fixed before merge. A first pass there looks like maybe it's from trying to set the backend to 'retina' when the remote machine is not a macbook? I don't know why that would only crop up as a problem _now_ though.

ligerzero-ai:
> Nice. I covered it in red ink, but it's mostly pretty specific stuff and I don't see any major issues.
> 
> The bit about implicitly re-assigning the `input.structure` every time the map is updated is the only part that might keep me up at night, the rest mostly consistency, QoL, readability, and efficiency suggestions. I also appreciate that you took the time to patch some typos and some of the outdated syntax that was being used in these files.
> 
> Edit: Oh, and the notebooks test should be fixed before merge. A first pass there looks like maybe it's from trying to set the backend to 'retina' when the remote machine is not a macbook? I don't know why that would only crop up as a problem _now_ though.

Just a quickie but it appears that the latest ipython update broke the test, so it's not relevant to this PR. See convo below:

image

image

https://github.com/ipython/comm/pull/6

pyiron_atomistics/vasp/base.py
ligerzero-ai: Yep - it's just for the case where we're importing external non-pyiron jobs into pyiron. In that case we just assign a naive map.

More specifically, there is no structure in the job, or if the structure has length 0 (empty list).

pyiron_atomistics/vasp/base.py
ligerzero-ai: To be completely honest, this is only here because it failed tests.

I would prefer to have it not exist, but one of the tests fails if I remove this. Something about trying to access an attribute of an old job in the hdf5 that doesn't exist throws an error that fails tests. So if at some point if we want to remove backwards compatibility and instead get users to re-generate the hdf5s so that they are valid with this implementation, we can get rid of this try; except.

pyiron_atomistics/vasp/base.py
ligerzero-ai: Im not fussed, let's go with use_structure_reordering, since I agree that it's better understood that way.

pyiron_atomistics/vasp/base.py
ligerzero-ai: > This documentation makes assumptions about the behaviour of another function. What's more, it makes assumptions about a function that calls this function, which I feel is very easy to lose track of. In contrast, if some of this same discussion were in write_input, at least there we'd be discussing assumptions about a function we're calling.

This is mostly just me trying to not completely rewrite the entire module and trying to preserve the original as much as possible xD

If it were up to me, I would have completely rewritten large parts of this module, but this is what I have inherited, and I thought some might prefer that I leave it the way it was. (Even while being somewhat self-aware of my hubris, I still think that I could do a much better job with a complete re-write for achieving what some parts of this module is trying to do.) 

I don't like it any more than you do. The logic is mostly written in the docstring for me to understand what exactly was going on in the (somewhat convoluted) code, and for any future unlucky people that are trying to work with this module.

> The bit about "if you call it manually and want to be consistent with write_input is good, but I see we use a private variable. Should we make _idx_X_to_Y public?

it was originally public in an earlier commit, but @pmrv suggested it private since he didn't want it polluting the tab complete. So I guess we should have a throwdown xD. His complaint was mostly the naming, but I think `idx_user_to_pyiron` and vice-versa make great variable names since it makes it pretty clear what's exactly going on. 

> I don't think we need abbreviations like "obj" and "fn" in the docs.

Sure, I'll remove them and put in full words

pyiron_atomistics/vasp/base.py
ligerzero-ai: addressed above - to do with old vasp imports. So I guess AttributeError, maybe.

pyiron_atomistics/vasp/base.py
ligerzero-ai: if there's an issue with this, you may also have issues with what's going on in the sphinx module xD.

I basically ripped it out of there.

I would have preferred a different implementation myself, but I'm just trying to keep some semblance of consistency between code modules for methods which are doing the same thing. I had a lot of trouble understanding the logic, like you. It's not readable to me unless I spend a good half hour-hour on it, if I were looking at the code as an outsider. Open to changing it to something explicit and more readable. We should discuss at the meeting on Monday.

pmrv:
Looking over the old code probably a good way the sorting is done in `vasp.structure.write_poscar`.  So a start would be

1. pull the sorting code from sphinx into a generic function
2. remove the sorting from write_poscar and pass a sorted structure with the generic function
3. save the mapping on the vasp base job
4. add the options object and make the sorting optional

PR #895, opened 2022-12-06
Set initial magnetic moments

samwaseda:
This allow the user to set the magnetic moments via single value or `dict`:

Example I

```python
structure = pr.create.structure.bulk('Ni', cubic=True)
structure[-1] = 'Fe'
v_dict = {'Fe': 1, 'Ni': 2}
structure.set_initial_magnetic_moments(v_dict)
print(structure.get_initial_magnetic_moments())
```

Output: `array([2, 2, 2, 1])`

Example II

```python
structure = pr.create.structure.bulk('Ni', cubic=True)
structure[-1] = 'Fe'
structure.set_initial_magnetic_moments(1)
structure.get_initial_magnetic_moments()
```

Output `array([1, 1, 1, 1])`

Moreover, I saw that the default behaviour was setting the system non-magnetic, i.e. regardless of what was set for the magnetic moments, `structure.set_initial_magnetic_moments()` would make the structure non-magnetic. This is now disabled, i.e. if the user wants to make it non-magnetic, they are required to explicitly set `structure.set_initial_magnetic_moments(None)`

### DIFF
--- a/pyiron_atomistics/atomistics/structure/atoms.py
+++ b/pyiron_atomistics/atomistics/structure/atoms.py
@@ -2569,6 +2569,8 @@ class Atoms(ASEAtoms):
         Args:
             magmoms (None/numpy.ndarray/list/dict/float): Default value is None (non magnetic calc).
             List, dict or single value assigning magnetic moments to the structure object.
+            
+            Non-collinear calculations may be specified through using a dict (see last example)
 
         If you want to make it non-magnetic, set `None`
         >>> structure.set_initial_magnetic_moments(None)
@@ -2582,7 +2584,7 @@ class Atoms(ASEAtoms):
         >>> structure.set_initial_magnetic_moments(spin_list)
         >>> structure.get_initial_magnetic_moments()
         array([1, 2, 3, 4])
-
+        
         Example II input: dict
         Assigns species-specific magnetic moments species
         >>> from pyiron_atomistics import Project
@@ -2601,13 +2603,35 @@ class Atoms(ASEAtoms):
         >>> structure.set_initial_magnetic_moments(1)
         >>> print(structure.get_initial_magnetic_moments())
         array([1, 1, 1, 1])
+        
+        Example IV input: dict/list for non-collinear magmoms.
+        Assigns non-collinear magnetic moments to the sites in structure
+        >>> from pyiron_atomistics import Project
+        >>> structure = Project('.').create.structure.bulk('Ni', cubic=True)
+        >>> structure[-1] = 'Fe'
+        
+        Option 1: List input sets vectors for each individual atom
+        >>> non_coll_magmom_vect = [[1, 2, 3]
+                                    [2, 3, 4],
+                                    [3, 4, 5],
+                                    [4, 5, 6]]
+        >>> structure.set_initial_magnetic_moments(non_coll_magmom_vect)
+        >>> print(structure.get_initial_magnetic_moments())
+        array([[1, 2, 3], [2, 3, 4], [3, 4, 5], [4, 5, 6]])
+        
+        Option 2: Dict input sets magmom vectors for individual species:
+        >>> print(structure.get_initial_magnetic_moments())
+        >>> non_coll_spin_dict = {'Fe': [2, 3, 4], 'Ni': [1, 2, 3]}
+        >>> structure.set_initial_magnetic_moments(spin_dict)
+        >>> print(structure.get_initial_magnetic_moments()) 
+        array([[1, 2, 3], [1, 2, 3], [1, 2, 3], [2, 3, 4]])       
         """
         # pyiron part
         if magmoms is not None:
             if isinstance(magmoms, dict):
                 if set(self.get_species_symbols()) != set(magmoms.keys()):
                     raise ValueError(
-                        "Elements in structure {} not equal to those in dict {}".format(
+                        "Elements in structure {} not found in dict {}".format(
                             set(self.get_chemical_symbols()), set(magmoms.keys())
                         )
                     )
@@ -2615,7 +2639,7 @@ class Atoms(ASEAtoms):
             elif not isinstance(magmoms, (np.ndarray, Sequence)):
                 magmoms = len(self) * [magmoms]
             if len(magmoms) != len(self):
-                raise ValueError("magmons can be collinear or non-collinear.")
+                raise ValueError("magmoms can be collinear or non-collinear.")
             if "spin" not in self._tag_list._lists.keys():
                 self.add_tag(spin=None)
             for ind, spin in enumerate(magmoms):
--- a/pyiron_atomistics/atomistics/structure/atoms.py
+++ b/pyiron_atomistics/atomistics/structure/atoms.py
@@ -2569,7 +2569,7 @@ class Atoms(ASEAtoms):
         Args:
             magmoms (None/numpy.ndarray/list/dict/float): Default value is None (non magnetic calc).
             List, dict or single value assigning magnetic moments to the structure object.
-            
+
             Non-collinear calculations may be specified through using a dict/list (see last example)
 
         If you want to make it non-magnetic, set `None`
@@ -2584,7 +2584,7 @@ class Atoms(ASEAtoms):
         >>> structure.set_initial_magnetic_moments(spin_list)
         >>> structure.get_initial_magnetic_moments()
         array([1, 2, 3, 4])
-        
+
         Example II input: dict
         Assigns species-specific magnetic moments
         >>> from pyiron_atomistics import Project
@@ -2603,13 +2603,13 @@ class Atoms(ASEAtoms):
         >>> structure.set_initial_magnetic_moments(1)
         >>> print(structure.get_initial_magnetic_moments())
         array([1, 1, 1, 1])
-        
+
         Example IV input: dict/list for non-collinear magmoms.
         Assigns non-collinear magnetic moments to the sites in structure
         >>> from pyiron_atomistics import Project
         >>> structure = Project('.').create.structure.bulk('Ni', cubic=True)
         >>> structure[-1] = 'Fe'
-        
+
         Option 1: List input sets vectors for each individual site
         >>> non_coll_magmom_vect = [[1, 2, 3]
                                     [2, 3, 4],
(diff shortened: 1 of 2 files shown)


pyiron_atomistics/atomistics/structure/atoms.py
```diff
@@ -2561,22 +2561,71 @@ def get_initial_magnetic_moments(self):
...
+        >>> from pyiron_atomistics import Project
+        >>> structure = Project('.').create.structure.bulk('Ni', cubic=True)
+        >>> structure[-1] = 'Fe'
+        >>> v_dict = {'Fe': 1, 'Ni': 2}
+        >>> structure.set_initial_magnetic_moments(v_dict)
+        >>> print(structure.get_initial_magnetic_moments())
+
+        Output:
+
+        >>> array([2, 2, 2, 1])
```
pmrv:
```suggestion
        >>> structure.get_initial_magnetic_moments()
        array([2, 2, 2, 1])
```

then it should work directly with the doctest.

pyiron_atomistics/atomistics/structure/atoms.py
```diff
@@ -2561,22 +2561,71 @@ def get_initial_magnetic_moments(self):
...
+
+        If a list or an array of length 3 (e.g. [1, 2, 3]) is set and the
+        number of atoms is not 3, it will be considered as a non-collinear
+        magnetic moments, and the values will be distributed to all the
+        atoms (like in Example II)
         """
         # pyiron part
         if magmoms is not None:
+            if isinstance(magmoms, dict):
+                if set(self.get_chemical_symbols()) != set(magmoms.keys()):
```
pmrv:
```suggestion
                if set(self.get_species_symbols()) != set(magmoms.keys()):
```

pmrv: Probably `sorted` would also work, but I don't care much which you want to use.

pyiron_atomistics/atomistics/structure/atoms.py
```diff
@@ -2561,22 +2561,71 @@ def get_initial_magnetic_moments(self):
...
+                    raise ValueError(
+                        "Elements in structure {} not equal to those in dict {}".format(
+                            set(self.get_chemical_symbols()), set(magmoms.keys())
+                        )
+                    )
+                magmoms = [magmoms[c] for c in self.get_chemical_symbols()]
+            elif not hasattr(magmoms, "__len__") or (
```
pmrv:
```suggestion
            elif not isinstance(Sequence, magmoms) or (
```

plus `from collections.abc import Sequence`.

pmrv: Probably better to first handle the case that it's a sequence explicitely and then do this case rather than checking that it is not a sequence.

pyiron_atomistics/atomistics/structure/atoms.py
```diff
@@ -2561,22 +2561,71 @@ def get_initial_magnetic_moments(self):
...
+                magmoms = len(self) * [magmoms]
             if len(magmoms) != len(self):
                 raise ValueError("magmons can be collinear or non-collinear.")
             if "spin" not in self._tag_list._lists.keys():
                 self.add_tag(spin=None)
             for ind, spin in enumerate(magmoms):
-                self.spin[ind] = spin
-        self.spins = magmoms
+                self.spin[ind] = spin  # For self._tag_list.spin
+        self.spins = magmoms  # For self.array['initial_magmoms']
```
pmrv, line 2650: I find it very yikes, that we have `spin` and `spins` and then also the ASE inherited `set_initial_magnetic_moments()`.  I would prefer to call `super().set_initial_magnetic_moments` here directly because that's what the `spins` setter does.  Since this is not introduced in this PR I don't feel strongly though.

pyiron_atomistics/atomistics/structure/atoms.py
```diff
@@ -2561,22 +2561,71 @@ def get_initial_magnetic_moments(self):
             else:
                 return np.zeros(len(self))
 
-    def set_initial_magnetic_moments(self, magmoms=None):
+    def set_initial_magnetic_moments(self, magmoms):
```
pmrv: The default is also present in the ASE method, so imo we should leave it here as well.

pmrv (approved):
Some nits, but generally ok with me.

pyiron_atomistics/atomistics/structure/atoms.py
samwaseda: > Probably better to first handle the case that it's a sequence explicitely and then do this case rather than checking that it is not a sequence.

I don't really understand what you mean by this.

pyiron_atomistics/atomistics/structure/atoms.py
samwaseda: Yeah I feel the same. I also thought about changing it directly in this PR, but I was criticised for including too many changes in a PR, so I don't think I'm gonna do it.

pyiron_atomistics/atomistics/structure/atoms.py
samwaseda: Apparently `ndarray` is not`Sequence`, so I cannot really use this one here.

pmrv: Maybe just `isinstance(magmoms, (ndarray, Sequence))` then.

pmrv: > > Probably better to first handle the case that it's a sequence explicitely and then do this case rather than checking that it is not a sequence.
> 
> I don't really understand what you mean by this.

Disregard, didn't read carefully enough.

pyiron_atomistics/atomistics/structure/atoms.py
```diff
@@ -2561,22 +2561,71 @@ def get_initial_magnetic_moments(self):
...
+            if isinstance(magmoms, dict):
+                if set(self.get_chemical_symbols()) != set(magmoms.keys()):
+                    raise ValueError(
+                        "Elements in structure {} not equal to those in dict {}".format(
+                            set(self.get_chemical_symbols()), set(magmoms.keys())
+                        )
+                    )
+                magmoms = [magmoms[c] for c in self.get_chemical_symbols()]
+            elif not hasattr(magmoms, "__len__") or (
+                len(magmoms) == 3 and len(self) != 3
```
pmrv: I find this a bit tricky, because what happens if I iterate over structures and I want to set the same (vector) magmom for all of them?  If one of the structures happens to be three atoms, it'll try colinear spins and break.  You mention it in the docstring, but it seems annoying.  I'm not sure how common this case is anyway, but maybe we add a flag that forces `magmoms` to be intepreted as a vector spin?

pyiron_atomistics/atomistics/structure/atoms.py
```diff
@@ -2566,17 +2566,63 @@ def set_initial_magnetic_moments(self, magmoms=None):
         Set array of initial magnetic moments.
 
         Args:
-            magmoms (numpy.ndarray/list): List of magneric moments
+            magmoms (numpy.ndarray/list/dict/float/None): List, dict or single
+                value of magneric moments
```
ligerzero-ai:
```suggestion
                value of magnetic moments
```

pyiron_atomistics/atomistics/structure/atoms.py
```diff
@@ -2566,17 +2566,63 @@ def set_initial_magnetic_moments(self, magmoms=None):
...
+
+
+        Example I:
```
ligerzero-ai: If it takes a an array, list, dict, float, it would probably be good to explicitly show what the behaviours are in each of the cases, but I'm just mostly nitpicking. I'll write up the other examples after shower/dinner

pyiron_atomistics/atomistics/structure/atoms.py
```diff
@@ -2566,17 +2566,63 @@ def set_initial_magnetic_moments(self, magmoms=None):
...
+
+        Output:
+
+        >>> array([1, 1, 1, 1])
+
+        If you want to make it non-magnetic, set `None`:
+
+        >>> structure.set_initial_magnetic_moments(None)
+
+        If a list or an array of length 3 (e.g. [1, 2, 3]) is set and the
```
ligerzero-ai: This is a little implicit for my liking? I'd prefer to force the user to set a flag non-collinear when they wish to specify that behaviour and then their code also becomes more readable overall too.

pyiron_atomistics/atomistics/structure/atoms.py
```diff
@@ -2566,17 +2566,63 @@ def set_initial_magnetic_moments(self, magmoms=None):
...
         # pyiron part
         if magmoms is not None:
+            if isinstance(magmoms, dict):
+                if set(self.get_species_symbols()) != set(magmoms.keys()):
+                    raise ValueError(
+                        "Elements in structure {} not equal to those in dict {}".format(
+                            set(self.get_chemical_symbols()), set(magmoms.keys())
+                        )
+                    )
+                magmoms = [magmoms[c] for c in self.get_chemical_symbols()]
```
ligerzero-ai, line 2641: Note for myself to come back to later - unsure if this assumes species are ordered (selfish reasons as I'm in the middle of implementing the functionality and don't want more implicit assumptions of ordering to filter out). I (think) it might, but I need to open my IDE to see what exactly this returns.

pyiron_atomistics/atomistics/structure/atoms.py
samwaseda: XD good job

pyiron_atomistics/atomistics/structure/atoms.py
samwaseda: yeah, I was also not really sure if this should be offered or not, on the other hand, it's so extremely common to put just one value, especially if it's ferromagnetic. I can also remove it if you think it's too dangerous.

pyiron_atomistics/atomistics/structure/atoms.py
samwaseda: ok so I guess I just simply remove this possibility for non-collinear calculations. It's probably less common that people set the same vector for all atoms.

pyiron_atomistics/atomistics/structure/atoms.py
samwaseda: No at this point it's not ordered. It's also inside `atoms.py`, so there's no ordering.

pyiron_atomistics/atomistics/structure/atoms.py
samwaseda: So in the end, I removed the possibility to put a simple 3d-vector which is to be copied to all atoms. Either the user sets a full array, or a `dict` like `{"Fe": [1, 0, 0]}`

pmrv: Ok, doing it with `dict` is actually a nice compromise.

pyiron_atomistics/atomistics/structure/atoms.py
ligerzero-ai: Okay, got around to checking it. It works as an explicit array generated, so I'm satisfied.

image

ligerzero-ai (approved):
All done - let's get this merged.

PR #897, opened 2022-12-06
Generic calc_static interface

pmrv:
This is my first stab at a generic interface for the calc_modes.  Next step is to implement a similar thing for DFT codes.

### DIFF
--- a/pyiron_atomistics/lammps/base.py
+++ b/pyiron_atomistics/lammps/base.py
@@ -66,7 +66,7 @@ class DumpData:
 
 class LammpsStaticOutput(StaticWithPressureOutput):
 
-    def __init__(self, job: LammpsBase):
+    def __init__(self, job: "LammpsBase"):
         self._job = job
 
     @property
--- a/pyiron_atomistics/lammps/base.py
+++ b/pyiron_atomistics/lammps/base.py
@@ -60,8 +60,25 @@ class DumpData:
     positions: List = field(default_factory=lambda: [])
     computes: Dict = field(default_factory=lambda: {})
 
+class LammpsStaticOutput(StaticWithPressureOutput):
 
-class LammpsBase(AtomisticGenericJob):
+    def __init__(self, job: LammpsBase):
+        self._job = job
+
+    @property
+    def energy_pot(self):
+        return job.output.energy_pot[-1]
+
+    @property
+    def forces(self):
+        return job.output.forces[-1]
+
+    @property
+    def pressure(self):
+        return job.output.pressures[-1]
+
+
+class LammpsBase(CalcStatic, AtomisticGenericJob):
     """
     Class to setup and run and analyze LAMMPS simulations which is a derivative of
     atomistics.job.generic.GenericJob. The functions in these modules are written in such the function names and
@@ -805,14 +822,23 @@ class LammpsBase(AtomisticGenericJob):
 
     calc_minimize.__doc__ = LammpsControl.calc_minimize.__doc__
 
-    def calc_static(self):
-        """
+    # CalcStatic Impl
+    @property
+    def _calc_input(self):
+        return self._generic_input
 
-        Returns:
-
-        """
-        super(LammpsBase, self).calc_static()
+    def _calc_static(self):
         self.input.control.calc_static()
+
+    # CalcStatic Impl
+    def collect_static(self) -> LammpsStaticOutput:
+        # HACK: in the current architecture collection is done centrally, so we
+        # defer here to collect_output.  In the future we want to split the
+        # output parsing for each calculation mode
+        if not self.status.finished:
+            self.collect_output()
+        return LammpsStaticOutput(self)
+
 
     def calc_md(
         self,
--- a/pyiron_atomistics/lammps/base.py
+++ b/pyiron_atomistics/lammps/base.py
@@ -21,7 +21,11 @@ from pyiron_atomistics.lammps.potential import (
     view_potentials,
     list_potentials,
 )
-from pyiron_atomistics.atomistics.job.atomistic import AtomisticGenericJob
+from pyiron_atomistics.atomistics.job.atomistic import (
+        AtomisticGenericJob,
+        CalcStatic,
+        StaticWithPressureOutput
+)
 from pyiron_base import state, extract_data_from_file, deprecate
 from pyiron_atomistics.lammps.control import LammpsControl
 from pyiron_atomistics.lammps.potential import LammpsPotential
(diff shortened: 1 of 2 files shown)


liamhuber:
Instead of using properties, what if we set `StaticOutput(ABC, HasTraits)` and use traits instead? E.g.:

```python
import numpy as np
from traitlets import HasTraits, Float, Instance
from pyiron_base.storage.has_stored_traits import ABCTraitsMeta, HasStoredTraits


class StaticOutput(HasTraits, ABC, metaclass=ABCTraitsMeta):
    energy_pot = traitlets.Float()
    forces = traitlets.Instance(klass=np.ndarray)


class StaticWithPressureOutput(StaticOutput, ABC):
    pressures = traitlets.Instance(klass=np.ndarray)


class LammpsStaticOutput(StaticWithPressureOutput):
    energy_tot = traitlets.Float()  # Do we really need this? For static it should always be exactly pot?

    def __init__(self, job: LammpsBase):
        self.energy_pot = job.output.energy_pot[-1]
        self.forces = job.output.forces[-1]
        self.pressures = job.output.pressures[-1]
        self.energy_tot = job.output.energy_tot[-1]
        # self.lock()  # Should output just always be read-only?
```

I think traitlets has advantages in terms of verbosity (lack thereof), the ability to add real type checking, and some mixins (like `observe`) that may wind up being useful.

Thinking back to our at the last meeting, we discussed really splitting things into `Engines` (which don't have any storage) and `Calculators` (which do). Something else nice about this formulation is that we can trivially convert out pure-traitlets input or output classes for engines into their storable-counterpart for calculators by adding a mixin:

```python
class StorableLammpsStaticOutput(HasStoredTraits, LammpsStaticOutput):
    pass
```

(And, in fact, we can compose any number of `HasTraits` classes together with `HasStoredTraits` to the same effect).

pyiron_atomistics/lammps/base.py
```diff
@@ -60,8 +64,25 @@ class DumpData:
     positions: List = field(default_factory=lambda: [])
     computes: Dict = field(default_factory=lambda: {})
 
+class LammpsStaticOutput(StaticWithPressureOutput):
+
+    def __init__(self, job: "LammpsBase"):
```
liamhuber, line 69: Doesn't this resolve OK with the actual class if you have `from __future__ import annotations`?

pmrv: Didn't know about it, but reading it, yeah, it should.

pmrv:
I decided against using more complex objects for the output, because I want to leave a lot of freedom for implementations to as efficient as they can.  E.g. in your example the output will be copied from the HDF5 into the object and then again once the user accesses them.  That's probably ok for static calculations, but for larger scale MD or interactive jobs with multiple steps it might be problematic.  I'm still a bit struggling to put a code example together, what I want to allow is that multiple Output objects access the same underlying storage without copying anything.

I'm also not sure if the type checking, observing, etc. of traitlets gets us so much, since I would expect the output to be read-only and set by pyiron. So there would be no place a user could mess something up.

*super rough*

Say we had an interactive murnaghan type of calculation with three volumes and an internal minimization at each of them.  Say after the engine parsed each of the steps independently it put it in HDF or a flat store.

```python

store = FlattenedStorage()
# first minimization
store.add_chunk(2, energy_pot=3.3, energy_pot_per_atom=[1.7, 1.6], volume=1, step=1)
store.add_chunk(2, energy_pot=3.0, energy_pot_per_atom=[1.5, 1.5], volume=1, step=1)

# second
store.add_chunk(2, energy_pot=2.4, energy_pot_per_atom=[1.2, 1.2], volume=2, step=2)
store.add_chunk(2, energy_pot=2.2, energy_pot_per_atom=[1.1, 1.1], volume=2, step=2)
store.add_chunk(2, energy_pot=2.0, energy_pot_per_atom=[1.0, 1.0], volume=2, step=2)

# third
store.add_chunk(2, energy_pot=4, energy_pot_per_atom=[2, 2], volume=3, step=3)

class FullOutput:
  def __init__(self, store):
    self._store=store

  @property
  def energy_pot(self):
    return self._store['energy_pot']

class SubOutput:
  def __init__(self, full, step):
    ...

  @property
  def energy_pot(self):
    # doesn't actually work, but would return a sub array matched by the **kwargs
    return self._full._store.get_array("energy_pot", step=self._step)

# then you could have multiple accessors on the job
job.output # -> gives FullOutput
job.output.seps[1]  # -> gives SubOutput(job.output, step=1)
job.output.final # would give a third (not defined) output object that just contains the last output of each substep (like our interactive jobs now), so here that would be chunk 1, 4, 5
```

liamhuber:
Ah, I think we had different goals in mind then. If I understood @jan-janssen correctly, and certainly for the ironflow stuff, what we're looking for is a *transient* interface level that doesn't interact with the HDF at all -- i.e. the pyiron objects (not `PyironObject`s) are purely python and live in RAM, although the engine will write and read IO for the external code being called.

I was super misleading in my example though, because I just kept the reference to `job.output...`, when actually I was trying to define an interface for something that doesn't know about jobs. I was just making minimal changes to your example, but that was silly of me.

While I think what you're trying to build here may be different from what I'm trying to build, I like the key points of (a) multiple objects having read-only access to the same underlying serialized data (i.e. storage conservation), and (b) explicit access points to get different elements of that data (good UX).

Maybe the point of contact between the engine and job layers comes in something like this `job.store.add_chunk(bigger_data=engine.output.bigger_data)`?

----------

> I decided against using more complex objects for the output, because I want to leave a lot of freedom for implementations to as efficient as they can. ... I'm also not sure if the type checking, observing, etc. of traitlets gets us so much, since I would expect the output to be read-only and set by pyiron. 

Yes, simple objects for efficiency is a good argument. At the user-facing level I think either type-checking or enforced read-only-access becomes important, and `HasStoredTraits` comes with both of these, so maybe we can use it at the topmost layer of abstraction -- although your other point about large scale MD and copying objects is also well taken. At the dev-facing level I think the type checking may still be useful, e.g. someone writes an interface to a new code (and engine) and inherits from a particular output class, and then gets errors during development because their parser is sending data of the wrong format to the output object. 

I also still don't like the verbosity of all the abstract getters on the dev-facing side. I just poked around with the idea of using dataclasses to accomplish this, but they're pretty awkward for the sort of many-read-from-one behaviour that you are aiming for, and [making them read-only is a bit awkward](https://stackoverflow.com/questions/61237131/how-to-make-attribute-in-dataclass-read-only), so maybe the verbose property approach is as good as it gets...

PR #908, opened 2022-12-19
Fix SPHInX bandstructure restarts

freyso:
This fixes issues when calling restart_for_bandstructure from a minimization run: generic input and the main group must be cleaned up from all minimizer stuff, and then main is recreated freshly.


skatnagallu (approved):
Looks good to me. I would wait on Sam for his thumbs up.

samwaseda:
I'm not sure if I understand the changes correctly. The problem is that the final spins from the old calculation are not set as the initial spins of the new calculation?

freyso:
> I'm not sure if I understand the changes correctly. The problem is that the final spins from the old calculation are not set as the initial spins of the new calculation?

No. A band structure rerun requires to **replace** certain items of the input tree:
 - the initial density (initialGess.rho) must be read from a file, **all** other density (incl. spin) initialization must go away.
 - the content of the main group must be replaced by a "fixed-rho" scfDiag block

The main-line version fails to do this. The present changeset moves the initial spin stuff alongside the "atomicOrbitals" density initialization, and reinitializes the main group.

PR #914, opened 2022-12-22
fix issues cause by deprecated numpy float

Leimeroth:
After updating today my pyiron quitted its job because np.float is removed in numpy 1.24
This should fix it.

EDIT:
Actual code changes are only a few lines, most come from black formatter


pmrv (approved):
lgtm

PR #854, opened 2022-11-05
Add functional style methods to HasStructure

pmrv:
This adds some transform/filter methods to `HasStructure` to allow a "functional programming" style in its use.  Contingent on #853 and some tests/docs.

PR #920, opened 2023-01-06
Remove redundant property consuming huge amounts of memory in vasprun object

ligerzero-ai:
Part 1 of a series of fixes for vasp jobs consuming way too much memory


ligerzero-ai:
doc tests fail because of unrelated issue to this pull #854 rest is fine - merging as @pmrv was with me reviewing this code already

PR #925, opened 2023-01-09
Revert "Iteratively parse vasprun instead of having entire object in memory"

ligerzero-ai:
Reverts pyiron/pyiron_atomistics#922


ligerzero-ai:
Doesn't appear to fix the problem of being killed on the cmti head nodes when `vasprun().from_file()` is called, so it's still taking way too much memory.

image

PR #928, opened 2023-01-12
Pull structure sorting from sphinx module into generic dft

ligerzero-ai:


pyiron_atomistics/dft/job/generic.py
```diff
@@ -501,6 +515,54 @@ def get_density_of_states(self, sigma=0.1, shift_by_fermi_energy=True, grid=None
...
+            self._map_pyiron_to_user_idx()
+        return self._idx_user_to_pyiron
+
+    @property
+    def idx_pyiron_to_user(self):
+        if self.structure is None:
+            return None
+        if len(self._idx_pyiron_to_user) == 0:
+            self._map_pyiron_to_user_idx()
+        return self._idx_pyiron_to_user
```
pmrv, line 539: These are not strictly necessary, right? Everything that can access those could also access `job.input.idx_pyiron_to_user` if I see it correctly?

pmrv (approved):
Small nit about the top level property like last time, but looks great to me otherwise.

pyiron_atomistics/dft/job/generic.py
ligerzero-ai: I am not sure what you mean actually, I think that it needs to exist because the properties in the job class implicitly call this property upon trying to access either `idx_pyiron_to_user` or `idx_user_to_pyiron` that are attached on `job.input` attribute.

This then triggers the mapping function attached to the `DFTInput` class; contingent on detecting that there is a structure and the hidden variables `_idx_user_to_pyiron` and `_idx_pyiron_to_user` are empty.

ligerzero-ai:
> Small nit about the top level property like last time, but looks great to me otherwise.

I was just too lazy and copy pasted from the older pull - I still need to sort out the failing test and rename the property as per our discussion. It will probably become really long, since it's public and we want to avoid the nameclash with the chemical indices. I still think it's kind of annoying that we have to avoid that name in the first place. I think chemical species "indices" should give way to structure indices

pyiron_atomistics/dft/job/generic.py
ligerzero-ai: Can you show me why you think it is redundant?

ligerzero-ai: I don't usually write code that has this complex logic, so I may have missed an easier way to achieve what is being done here, happy to learn.

ligerzero-ai:
I will probably need also a little help understanding this failed unittest, I can't reproduce it locally.

image

samwaseda:
https://github.com/pyiron/pyiron_atomistics/blob/76b70d264912831c601e08858e9046bd5bcda5cf/pyiron_atomistics/sphinx/base.py#L95-L99

I guess the problem comes from the fact that `InputWriter` is not the Sphinx input, but it's actually `Group`.

Issue #696, opened 2022-08-01
`calc_minimize(pressure=[0, 0, 0])` does not work for undefined structure


samwaseda:
```python
lmp = pr.create.job.Lammps('lmp')
lmp.calc_minimize(pressure=[0, 0, 0])
```

This code raises the error:

```python-traceback
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
/tmp/ipykernel_37410/1534360834.py in <cell line: 2>()
      1 lmp = pr.create.job.Lammps('lmp_test')
----> 2 lmp.calc_minimize(pressure=[0, 0, 0])

~/dev_sam/pyiron_atomistics/pyiron_atomistics/lammps/interactive.py in calc_minimize(self, ionic_energy_tolerance, ionic_force_tolerance, e_tol, f_tol, max_iter, pressure, n_print, style)
    362                 "calc_minimize() is not implemented for the non modal interactive mode use calc_static()!"
    363             )
--> 364         super(LammpsInteractive, self).calc_minimize(
    365             ionic_energy_tolerance=ionic_energy_tolerance,
    366             ionic_force_tolerance=ionic_force_tolerance,

~/dev_sam/pyiron_atomistics/pyiron_atomistics/lammps/base.py in calc_minimize(self, ionic_energy_tolerance, ionic_force_tolerance, e_tol, f_tol, max_iter, pressure, n_print, style)
    769             n_print=n_print,
    770         )
--> 771         self.input.control.calc_minimize(
    772             ionic_energy_tolerance=ionic_energy_tolerance,
    773             ionic_force_tolerance=ionic_force_tolerance,

~/dev_sam/pyiron_atomistics/pyiron_atomistics/lammps/control.py in calc_minimize(self, ionic_energy_tolerance, ionic_force_tolerance, max_iter, pressure, n_print, style, rotation_matrix)
    248         if pressure is not None:
    249             self._force_skewed = False
--> 250             pressure = self.pressure_to_lammps(pressure, rotation_matrix)
    251             if np.isscalar(pressure):
    252                 str_press = " iso {}".format(pressure)

~/dev_sam/pyiron_atomistics/pyiron_atomistics/lammps/control.py in pressure_to_lammps(self, pressure, rotation_matrix)
    152         # Isotropic, hydrostatic pressures are rotation invariant.
    153         if not np.isclose(
--> 154             np.matrix.trace(rotation_matrix), 3
[...]

PR #933, opened 2023-01-17
:memo: Add new exception for calc_minimize.

KuhnMn:
Solution for #696.  Add a new exception to the calc_minimize method that is raised when pressure is about to be calculated while the rotation_matrix is None. This would raise an exception anyways, but with a more generic message.

### DIFF
--- a/pyiron_atomistics/lammps/control.py
+++ b/pyiron_atomistics/lammps/control.py
@@ -247,8 +247,10 @@ class LammpsControl(GenericParameters):
 
         if pressure is not None:
             if rotation_matrix is None:
-                raise ValueError("No rotation matrix given while trying to convert pressure. "
-                                 "This is most likely due to no structure being defined.")
+                raise ValueError(
+                    "No rotation matrix given while trying to convert pressure. "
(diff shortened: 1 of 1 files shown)


KuhnMn:
Hello, I am having some issues understanding the failing workflows.
From the logs, the issue in both cases seems to be that the setuptools_scm module is not found. However, I changed nothing regarding the packages or modules of the project. With this being my first contribution, I'd appreciate some help in fixing the pipeline issues.
Thank you in advance!

jan-janssen:
Hi @KuhnMn , 

Unfortunately, we currently have an issue with the `mpi_api` package `pyiron_atomistics` depends on: 
https://github.com/pyiron/pyiron_base/issues/983

We are working on fixing this as soon as possible. 

Best, 

Jan

KuhnMn:
Hi, @jan-janssen  

thank you for the clarification!

Best,

Maximilian

jan-janssen:
It is fixed

KuhnMn:
Thanks a lot, that resolved the last checks not passing!

jan-janssen (approved):
Looks good to me

Issue #932, opened 2023-01-16
Reading POSCAR or CONTCAR files with selective dynamics inverts them


Leimeroth:
When using pr.create.structure.read("POSCAR") on VASP CONTCAR or POSCAR format selective dynamic tags are inversed.
I did not trace it back to its origin yet and did not check for other codes, so I don't know whether this is a bug inherited from ase,
a matter of different definitions or some messed up stuff on pyiron/ase conversion.

ligerzero-ai:
I touched this a while back so I will take a look at it tonight.

ligerzero-ai:
Hi, I need some time over the weekend to fix this - but the current workaround suggested by me is to do: 
```
from pyiron_atomistics.atomistics.structure.atoms import pyiron_to_pymatgen, pymatgen_to_pyiron
from pymatgen.core import Structure

structure = pymatgen_to_pyiron(Structure.from_file("POSCAR"))
```
I'm fairly sure that the above will be safe (as I have tests specifically checking the sel dyn tags implemented)

ligerzero-ai:
I checked ASE, there is no problem on their `ase.io.read` function so it must be on our end.

ligerzero-ai:
Actually, I stand corrected. This is inherited from ASE. 

MWE:
```
from pyiron_atomistics.atomistics.structure.atoms import pyiron_to_pymatgen, pymatgen_to_pyiron
from pymatgen.core import Structure
from pyiron import Project

pr = Project("asdf")
pi_atoms = pr.create.structure.bulk("Fe") * [2, 2, 1]
struct = pyiron_to_pymatgen(pi_atoms)
struct.add_site_property("selective_dynamics", values=[[False, False, False], [False, False, True], [False, False, False], [True, True, False]])
struct.to(fmt="poscar", filename="test_sel_dyn.vasp")
read_struct = pr.create.structure.read("test_sel_dyn.vasp")
read_struct
```
yields:
```
Fe: [0. 0. 0.]
Fe: [ 1.435 -1.435  1.435]
Fe: [-1.435  1.435  1.435]
Fe: [0.   0.   2.87]
tags: 
    selective_dynamics: [[False, False, False] [True, True, False] [False, False, False] [False, False, True]]
pbc: [ True  True  True]
cell: 
Cell([[-2.87, 2.87, 2.87], [2.87, -2.87, 2.87], [1.435, 1.435, -1.435]])
```

[...]

PR #934, opened 2023-01-20
Fix ase selective dynamics read bug

ligerzero-ai:
#932 replaces read() in StructureFactory with pymatgen reader which handles selective dynamics properly.

### DIFF
--- a/pyiron_atomistics/atomistics/structure/factory.py
+++ b/pyiron_atomistics/atomistics/structure/factory.py
@@ -110,7 +110,7 @@ class StructureFactory(PyironFactory):
             # For stuff that pymatgen can't read it raises
             # ValueError: Unrecognized file extension!
             # backwards compatibility for notebooks.
-            # TypeError: IStructure.from_file() got an unexpected keyword argument 'format' 
+            # TypeError: IStructure.from_file() got an unexpected keyword argument 'format'
             structure = self.read_using_ase(*args, **kwargs)
         return structure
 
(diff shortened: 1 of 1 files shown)


jan-janssen:
> #932 replaces read() in StructureFactory with pymatgen reader which handles selective dynamics properly.

I am surprised, does `pymatgen` supports the same range of input/ output files like `ase` ? maybe we should add an additional parameter. https://databases.fysik.dtu.dk/ase/ase/io/io.html#ase.io.write

ligerzero-ai:
pymatgen has openbabel as an optional dependency, and with it, it supports 113 file formats used in the chemistry community https://openbabel.org/docs/2.3.0/FileFormats/Overview.html

Fairly certain that pymatgen is therefore a more extensive reader, if you include openbabel as a dependency. How to proceed? Raise an error when selective dynamics is present? Or default to the pymatgen reader, and try ase where it fails?

jan-janssen:
From my experience, `openbabel` an be pretty annoying in terms of dependencies and I am not sure how much overlap the computational chemistry and the computational materials science codes have. In particular as the computational chemistry codes need the bond definitions and so on, which at least currently is not supported in pyiron, as we internally rely on the ASE structure. 

Still I am open to add multiple read functions like `read_with_ase()` and `read_with_pymatgen()` and so on. Then we only have to clarify which one we set as default. At the same time as your changes already break the jupyter notebooks, switching from ASE to pymatgen as primary reader breaks the backwards compatibility. 

Finally, we could also consider fixing the bug on the ASE side.

ligerzero-ai:
@jan-janssen newest changes ok to merge?

jan-janssen (approved):
Looks good to me

jan-janssen:
> @jan-janssen newest changes ok to merge?

Yes

PR #945, opened 2023-01-27
Don't reconstitute job object in pyiron table func

ligerzero-ai:
get_structure() inbuilt for pyiron_tables for atomistics reconstitutes the job object in its entirety. This is extremely slow and can cause kernel crashing @aabdelkawy . The new function reads from hdf5.

### DIFF
--- a/pyiron_atomistics/table/funct.py
+++ b/pyiron_atomistics/table/funct.py
@@ -238,7 +238,8 @@ def get_equilibrium_parameters(job):
 
 
 def get_structure(job):
-    atoms = pyiron_to_ase(job.to_object().get_structure())
+    atoms = Atoms().from_hdf(job["output"])
+    # I don't understand why any of this is needed?
     atoms_dict = {
         "symbols": atoms.get_chemical_symbols(),
         "positions": atoms.get_positions().tolist(),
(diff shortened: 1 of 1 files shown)


pyiron_atomistics/table/funct.py
```diff
@@ -238,7 +238,8 @@ def get_equilibrium_parameters(job):
 
 
 def get_structure(job):
-    atoms = pyiron_to_ase(job.to_object().get_structure())
+    atoms = Atoms().from_hdf(job["output"])
```
jan-janssen, line 241: This is not the same, as the structure is not always allocated at the `job["output"]` For example for the `Murnaghan` job this does not work: 
https://github.com/pyiron/pyiron_atomistics/blob/main/pyiron_atomistics/atomistics/master/murnaghan.py#L958

jan-janssen:
> get_structure() inbuilt for pyiron_tables for atomistics reconstitutes the job object in its entirety. This is extremely slow and can cause kernel crashing @aabdelkawy . The new function reads from hdf5.

I do not understand why this pull request was merged without any review. It is very hard to maintain a stable code when pull requests are merged without review.

ligerzero-ai:
Ah, my bad, I didn't realise that. Sorry.

PR #946, opened 2023-01-27
Revert "Don't reconstitute job object in pyiron table func"

ligerzero-ai:
Reverts pyiron/pyiron_atomistics#945


ligerzero-ai:
Reverted as it apparently does not work as output does not store structure consistently across jobs

PR #966, opened 2023-02-21
[minor] Make spx base a bit more pep8 compatible

samwaseda:


samwaseda:
Black and flake8 are incompatible...

PR #978, opened 2023-03-07
Not allow empty tag in sparse list

samwaseda:
- Remove `*args` in `add_tag` in `SparseList`, because it makes no sense to make an empty sparse list, which is also not offered anywhere.
- Add `remove_tag` on the `Atoms` level, which was hidden behind `_tag_lists` before.


samwaseda:
I merge this one, because in ASE anyway empty arrays are not allowed.

Issue #718, opened 2022-08-12
Structures species does not update when changing elements


Leimeroth:
When f.e. removing an element from a structure completely using f.e.
s[s.numbers==12] = "Cu"
the structures .species attribute is not updated. This leads to problems with using interatomic potentials. Exemplary Backtrace:

File ~/git_projects/pyiron_atomistics/pyiron_atomistics/lammps/base.py:207, in LammpsBase.potential(self, potential_filename)
    205     potential_elements = list(potential["Species"])[0]
    206     if not set(structure_elements).issubset(potential_elements):
--> 207         raise ValueError(
    208             "Potential {} does not support elements "
    209             "in structure {}.".format(potential_elements, structure_elements)
    210         )
    211 self.input.potential.df = potential
    212 if "Citations" in potential.columns.values:

ValueError: Potential ['Si', 'O'] does not support elements in structure ['C' 'O' 'Si'].

after removing the "C" from the structure

This generally looks very dangerous imo, because I could imagine several situations where this not correctly being updated leads to a wrong assignment of POSCAR files or atom types in potentials.

niklassiemer:
related https://github.com/pyiron/pyiron_atomistics/pull/700

PR #981, opened 2023-03-07
Remove sparse list

### LINKED ISSUE #980: Do we need `SparseList`?
We have been using our own tags for years now, but I see the following problems:

- A lot of extra lines of code which have to be maintained
- It's not numpy, and therefore either does not support the mathematical operations or repeated in our own class
- It is actually not even stored as sparse data (as shown [here](https://github.com/pyiron/pyiron_atomistics/issues/979))
- Not clear whether we gain a lot of space by using sparse list
  - For example, if we set `structure.selective_dynamics[:] = [True, True, False]`, then it becomes a dictionary containing every atom, which makes the data larger
  - There must be a more intelligent way to compress data.

samwaseda:
I made a super audacious step to remove `SparseList` and unify everything in `structure.arrays`, which is what is supported by ASE.

Closes #980 

This one includes changes made in [this PR](https://github.com/pyiron/pyiron_atomistics/pull/978) and [this PR](https://github.com/pyiron/pyiron_atomistics/pull/982) (and therefore wouldn't be merged as long as the other one is not merged)


samwaseda:
It looks like there's a problem with the user-defined species.

samwaseda:
Integration tests failing until [this PR](https://github.com/pyiron/pyiron_atomistics/pull/985) is merged

samwaseda:
I guess now it's really ready for review.

pyiron_atomistics/atomistics/structure/atoms.py
```diff
@@ -281,12 +276,16 @@ def set_species(self, value):
...
-        self._store_elements = {el.Abbreviation: el for el in value}
+        if value is not None:
+            self._species = list(value)[:]
+
+    @property
+    def _store_elements(self) -> dict:
+        return {el.Abbreviation: el for el in self.species}
+
+    @property
+    def _species_to_index_dict(self) -> dict:
```
pmrv: This might trigger #718 as well, right? I guess if it's currently broken anyway, it's ok that this PR is as well, but it might anyway be an appropriate time to think about a solution.

pyiron_atomistics/atomistics/structure/atoms.py
```diff
@@ -1977,7 +1967,15 @@ def extend(self, other):
...
 
-        new_indices = other.indices.copy()
+        d = self._store_elements.copy()
+        d.update(other._store_elements.copy())
+        chem, new_indices = np.unique(
+            self.get_chemical_symbols().tolist()
+            + other.get_chemical_symbols().tolist(),
+            return_inverse=True,
+        )
+        new_species = [d[c] for c in chem]
```
pmrv, line 2033: It'd be nice to move this whole block into the if clause below where it is actually used.  Otherwise it just takes a while to figure out where it is needed.

pyiron_atomistics/atomistics/structure/structurestorage.py
```diff
@@ -212,9 +212,7 @@ def add_structure(self, structure, identifier=None, **arrays):
         if structure.has("initial_magmoms"):
             arrays["spins"] = structure.spins
         if "selective_dynamics" in structure.get_tags():
```
pmrv, line 214: Should this also use the `.has` syntax now?

pmrv (commented):
Am generally happy with this!  It'd be nice if @ligerzero-ai can have a look as well, since it touches the `species` problem and he thought about this already.

I'm not sure if I like the fact that this returns now numpy arrays rather that lists because of backwards compat reasons.  What does `structure.array["foo"]` return on an ASE structure?  I agree that numpy arrays are nicer to work with, but I don't want to break code by changing the return type (like you had to do in the tests).   Maybe we can override `__getattr__` to do the `tolist()` conversion and print a warning and people who want array, can go to `structure.arrays`?

ligerzero-ai:
Has @jan-janssen communicated to you that he is in the process of removing the custom Atoms object over in the structure-toolkit repository [here](https://github.com/pyiron/structuretoolkit)? Just so that we don't spend too much time working on redundant solutions. @pmrv @samwaseda Once that is implemented I don't think there will be a need for this intermediate solution, and we can rip out the old Atoms object entirely in the next major update for pyiron.

The idea is that the base object will be entirely from ASE in the structure toolkit.

edit: Just so that we're clear, I'll be happy to look over the pull, just making sure that we're all on the same page regarding what everybody is working on. I'll get around to it hopefully tomorrow or the day after.

samwaseda:
We all agree that the pyiron-`Atoms` class has to be replaced by ASE, and @jan-janssen and I are essentially coming from the opposite directions to meet somewhere in between. Here we try to guarantee backward compatibility. I agree that we shouldn’t spend too much time on this, but after the removal of the `SparseList` class, there’s hardly any difference anymore, so I would say we can afford it.

ligerzero-ai:
> We all agree that the pyiron-`Atoms` class has to be replaced by ASE, and @jan-janssen and I are essentially coming from the opposite directions to meet somewhere in between. Here we try to guarantee backward compatibility. I agree that we shouldn’t spend too much time on this, but after the removal of the `SparseList` class, there’s hardly any difference anymore, so I would say we can afford it.

all good - then I'll look over the pull as promised tomorrow or the day after. Thanks for the work! :)

pyiron_atomistics/atomistics/structure/atoms.py
samwaseda: That’s probably true, but if I understand it correctly, it’s not something that exists in ASE, right? In that case we can think about getting rid of it in the next PR, rather than fixing it in the current implementation.

pyiron_atomistics/atomistics/structure/structurestorage.py
samwaseda: That would definitely be a possibility, but since `selective_dynamics` doesn’t exist in ASE, and the same functionality seems to be implemented in a fairly different way, (`FixAtoms`?) it might also make sense to leave it like this for now.

pyiron_atomistics/atomistics/structure/atoms.py
```diff
@@ -397,6 +396,7 @@ def add_high_symmetry_path(self, path):
 
         self._high_symmetry_path.update(path)
 
+    @deprecate("Use Atoms.set_array() instead")
```
ligerzero-ai:
```suggestion
    @deprecate("Use Atoms.set_array() instead: 
    e.g. Atoms.set_array('selective_dynamics', np.repeat([[True, True, False]], len(Atoms), axis=0))
```

pyiron_atomistics/atomistics/structure/atoms.py
ligerzero-ai: +1

pyiron_atomistics/atomistics/structure/atoms.py
```diff
@@ -2266,57 +2227,6 @@ def __setitem__(self, key, value):
         # For ASE compatibility
         self.numbers = self.get_atomic_numbers()
 
-    def __imul__(self, vec):
```
ligerzero-ai, line 2323: I'm assuming this is a removal of internal multiplication method? Just removal because it wasn't used or it was used in sparselist?

tests/atomistics/structure/test_atoms.py
```diff
@@ -1825,7 +1810,7 @@ def test_pymatgen_to_pyiron_conversion(self):
         pyiron_atoms_sd = pymatgen_to_pyiron(struct_with_sd)
 
         sd_equivalent = struct_with_sd.site_properties["selective_dynamics"] == [
-            x.selective_dynamics for x in pyiron_atoms_sd
+            x.selective_dynamics.tolist() for x in pyiron_atoms_sd
```
ligerzero-ai, line 1815: I haven't run these tests myself, but just looking at the tests, is there a specific reason why some tests need to call `.tolist()` when they didn't before and why some have had `.list()` removed? I'm assuming that there is type weirdness happening with this change?

ligerzero-ai:
Overall pretty happy with it - just some small things that probably need to be explained to me very slowly like I'm 5yo xD

Good work!

pyiron_atomistics/atomistics/structure/atoms.py
samwaseda: It's actually already defined in ASE. We could have shortened it much earlier, but for `SparseList` we would have needed it anyway. Now it fully uses ASE `__imul__`.

tests/atomistics/structure/test_atoms.py
samwaseda: I guess that's closely related to @pmrv's comment on list vs numpy.

samwaseda:
> I'm not sure if I like the fact that this returns now numpy arrays rather that lists because of backwards compat reasons. What does `structure.array["foo"]` return on an ASE structure? I agree that numpy arrays are nicer to work with, but I don't want to break code by changing the return type (like you had to do in the tests). Maybe we can override `__getattr__` to do the `tolist()` conversion and print a warning and people who want array, can go to `structure.arrays`?

Yeah I totally see your point, and if it's up to us I would also love to keep using things as they were before. Here, we are not only talking about getting rid of `SparseArray` and `SparseList`, but using ASE `array`, i.e. as long as we keep using `list`, we are not fully compatible with ASE. So whether we override certain functions or not, we will anyway have to do the transition from `list` to `array` at some point. I'm not really sure if this is a major change that would make things backward incompatible, and therefore requires one more PR (i.e. in this PR changes should be fully backward compatible and another PR to do a version bump). I ended up deciding to do it in one go, because: 1, `list` and `array` behave similarly; 2, `SparseList` and `SparseArray` are not so often used (at least not beyond just setting `selective_dynamics` etc.).

pyiron_atomistics/atomistics/structure/atoms.py
samwaseda: Actually it cannot be done, because ASE-`extend` has to be called after the species are checked. I presume that `indices` will be removed in the near future anyway, so I suggest to keep this location for now.

samwaseda:
Can I merge this one? @pmrv @ligerzero-ai

pyiron_atomistics/atomistics/structure/atoms.py
```diff
@@ -2113,13 +2075,14 @@ def __getitem__(self, item):
         return new_array
 
     def __getattr__(self, item):
-        if item in self._tag_list.keys():
-            return self._tag_list._lists[item]
-        return object.__getattribute__(self, item)
+        try:
+            return self.arrays[item]
```
pmrv, line 2132:
```suggestion
            return self.arrays[item].tolist()
```

pmrv:
> > I'm not sure if I like the fact that this returns now numpy arrays rather that lists because of backwards compat reasons. What does `structure.array["foo"]` return on an ASE structure? I agree that numpy arrays are nicer to work with, but I don't want to break code by changing the return type (like you had to do in the tests). Maybe we can override `__getattr__` to do the `tolist()` conversion and print a warning and people who want array, can go to `structure.arrays`?
> 
> Yeah I totally see your point, and if it's up to us I would also love to keep using things as they were before. Here, we are not only talking about getting rid of `SparseArray` and `SparseList`, but using ASE `array`, i.e. as long as we keep using `list`, we are not fully compatible with ASE. So whether we override certain functions or not, we will anyway have to do the transition from `list` to `array` at some point. I'm not really sure if this is a major change that would make things backward incompatible, and therefore requires one more PR (i.e. in this PR changes should be fully backward compatible and another PR to do a version bump). I ended up deciding to do it in one go, because: 1, `list` and `array` behave similarly; 2, `SparseList` and `SparseArray` are not so often used (at least not beyond just setting `selective_dynamics` etc.).

The `__getattr__` is not part of the ASE interface so we can return whatever we want.  So we can stay backwards compatible with our code and be ASE compatible at the same time with the suggestion that I made above.

Otherwise lgtm.

samwaseda:
https://github.com/pyiron/pyiron_atomistics/blob/2eac446ffab0a4875cb488532273d913395e3a13/tests/atomistics/structure/test_atoms.py#L687

@pmrv as it turned out, sparse array also allows partly array-like operations. I thought about restoring this by rewriting `__getattr__`, but it's a feature that's gonna be lost anyway when we fully move to ASE, so I find it redundant, and potentially counterproductive (i.e. more and more new users might keep using it without realising it's gonna be unavailable at some point). While I know that it's something that @jan-janssen doesn't like, i.e. making two steps at once, in this case removing `SparseArray` and moving from list to array, I would still love to do so, for the following reasons:

- `__getattr__` of tags is not a feature that's used very often
- Going from `list` to `np.array` is not quite violent.

What do you think?

samwaseda:
@pmrv ping

pmrv:
Sorry, I completely missed your updated comment.  Imo go ahead and merge.  Changing the type is not perfect, but integration tests pass.

pmrv (approved):

PR #982, opened 2023-03-08
Not store useless extra species info in `Atoms`

samwaseda:
I originally made these changes in [this PR](https://github.com/pyiron/pyiron_atomistics/pull/981) by mistake (or misinterpretation of certain errors), but it appears correct to me, so I separate it here.

### DIFF
--- a/pyiron_atomistics/atomistics/structure/atoms.py
+++ b/pyiron_atomistics/atomistics/structure/atoms.py
@@ -110,8 +110,6 @@ class Atoms(ASEAtoms):
         ):
             state.logger.debug("Not supported parameter used!")
 
-        self._store_elements = dict()
-        self._species_to_index_dict = None
         self._is_scaled = False
 
         self._species = list()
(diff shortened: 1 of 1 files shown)


pyiron_atomistics/atomistics/structure/atoms.py
```diff
@@ -290,12 +288,16 @@ def set_species(self, value):
...
+        if value is not None:
+            self._species = list(value)[:]
+
+    @property
+    def _store_elements(self) -> dict:
+        return {el.Abbreviation: el for el in self.species}
+
+    @property
+    def _species_to_index_dict(self) -> dict:
+        return {el: i for i, el in enumerate(self.species)}
```
pmrv, line 300: This will change the species numbering if you add or delete atoms of different types, I think.

samwaseda: But whenever atoms are added or deleted, `species` is always set anew, so it shouldn’t matter

pmrv: I still couldn't test the code, but it sounds to me like it could trigger behavior like [this](https://github.com/pyiron/pyiron_atomistics/issues/718).

samwaseda: I guess it's more error prone if this change is **not** merged (because then there are dangling unchecked variables). But feel free to take your time.

samwaseda:
Actually this one has to be merged first. Can I? @pmrv

pmrv:
I don't have time to look again in detail this week.  If you think this will not create new failure modes, imo go ahead.

samwaseda:
> I don't have time to look again in detail this week. If you think this will not create new failure modes, imo go ahead.

Since all the tests passed and in principle I don't see anything that should break, I'm gonna merge it, because if it breaks here, it's gonna probably break anyway when we fully change to ASE, so it's better to break with this small step :D

PR #984, opened 2023-03-09
Use `set_array` instead of setting `structure.arrays` directly & change position reading in `from_hdf` when scaled positions are to be read

samwaseda:
While working on [this PR](https://github.com/pyiron/pyiron_atomistics/pull/981), I realized that we use `structure.arrays['tag'] = xyz` very often. Since there is this function `set_array` in ASE, which checks the shape of the value etc., we should also use it.

After I did the change above, I realized that ASE does not allow a redefinition of positions, which is apparently initialized to an empty list. So for `from_hdf` we have to stick to the old notation. I also changed `set_scaled_positions` in `from_hdf` because it does not work for the same reason.

### DIFF
--- a/pyiron_atomistics/atomistics/structure/atoms.py
+++ b/pyiron_atomistics/atomistics/structure/atoms.py
@@ -249,21 +249,10 @@ class Atoms(ASEAtoms):
 
     @spins.setter
     def spins(self, val):
-        if val is not None:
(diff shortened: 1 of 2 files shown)


samwaseda:
`main` started failing after merging this branch, even though the tests didn't fail here. How's that possible?

PR #990, opened 2023-03-20
[minor] make correct order of entries for symmetry

samwaseda:


samwaseda:
I also added a test for the bug I had encountered

PR #1007, opened 2023-04-17
Evaluate pub_lst only if length > 0

samwaseda:
I don't really know how this thing works, so I make a mini PR.


jan-janssen (approved):
Looks good to me

PR #1008, opened 2023-04-17
Make potential algebraic

### LINKED ISSUE #997: Ideas for `pr.create.potential`
First: Anything currently available will stay unchanged -> ✔️ backward compatibility

Currently pyiron is either limited to the existing potentials, or the user has to look up how the potential is read by LAMMPS, if they want to overlay multiple potentials. Also, I find it annoying to have to define a job and a structure to be able to look up available EAM potentials. I want to introduce `pr.create.potential` to solve these issues.

Case I: Simple EAM Al potential:

```python
job = pr.create.job.Lammps(“eam_Al”)
job.structure = pr.create.structure.bulk(“Al”)
eam = pr.create.potential.effective.EAM(“Al”)
eam.set_potential(eam.list_potentials()[0])
job.potential = eam
```

Case II: Morse potential:

```python
job = pr.create.job.Lammps(“morse_Al”)
job.structure = pr.create.structure.bulk(“Al”)
morse = pr.create.potential.effective.Morse(“Al”)
morse.set_potential(alpha=0.1, D_0=10, r_0=3) # Following LAMMPS convention; should be explained in DocStr
job.potential = morse
```

Case III: Overlay EAM and Morse

```python
job = pr.create.job.Lammps(“EAM_morse_Al_C”)
job.structure = pr.create.structure.bulk(“Al”, cubic=True)
job.structure[1] = “C”
eam = pr.create.potential.effective.EAM(“Al”)
eam.set_potential(eam.list_potentials()[0])
morse_Al_C = pr.create.potential.effective.Morse(“Al”, “C”)
morse_Al_C.set_potential(alpha=0.1, D_0=10, r_0=3)
morse_C = pr.create.potential.effective.Morse(“C”)
morse_C.set_potential(alpha=0.24, D_0=4, r_0=1.1)
job.potential = eam * morse_C * morse_Al_C # pair_style hybrid ?
job.potential = eam + morse_C + morse_Al_C # pair_style  hybrid/overlay ?
job.potential = 0.5 * eam + 0.3 * morse_Al_C + 0.5 * morse_C # pair_style  hybrid/scale ?
job.potential = eam + morse_Al_C # missing C-C interactions -> error
```

And I guess for DFT it should look more or less the same:

```python
job = pr.create.job.Sphinx(“dft_Al”)
job.structure = pr.create.structure.bulk(“Al”)
pseudo = pr.create.potential.pseudo(“Al”)
pseudo.set_potential(“GW”)
job.potential = pseudo
```

But to be honest I haven’t really thought much about DFT. Most of the work has already been done by @Ahmed-Aslam a few years ago, so I will just have to implement it in the new factory class. Maybe we can also try to use the EAM class provided by ASE. Let me know what you think.

samwaseda:
Basic ideas are discussed [here](https://github.com/pyiron/pyiron_atomistics/issues/997)

```python
from pyiron_atomistics.lammps.potentials import Library, Morse, CustomPotential


pr = Project('CUSTOM_POTENTIAL')
pr.remove_jobs(recursive=True, silently=True)
```

### Specify potential via element

```python
eam = Library("Al")
lmp = pr.create.job.Lammps('lmp_unary_eam')
lmp.potential = eam
lmp.structure = pr.create.structure.bulk('Al', cubic=True)
lmp.run()
```
### Specify potential via `name`
```python
eam = Library(name="1995--Angelo-J-E--Ni-Al-H--LAMMPS--ipr1")
lmp = pr.create.job.Lammps('lmp_unary_eam_via_name')
lmp.potential = eam
lmp.structure = pr.create.structure.bulk('Al', cubic=True)
lmp.run()
```

### Unary
```python
morse = Morse("Al", D_0=0.5, alpha=1.1, r_0=2.1, cutoff=6)
lmp = pr.create.job.Lammps('lmp_unary_morse')
lmp.potential = morse
lmp.structure = pr.create.structure.bulk('Al', cubic=True)
lmp.run()
```

### Binary

In this case **all** combinations must be defined (which makes sense...)

```python
morse_Al = Morse("Al", D_0=0.5, alpha=1.1, r_0=2.1, cutoff=6)
morse_Al_Ni = Morse("Al", "Ni", D_0=0.2, alpha=1.05, r_0=2.2, cutoff=6)
morse_Ni = Morse("Ni", D_0=0.7, alpha=1.15, r_0=2.15, cutoff=6)

lmp = pr.create.job.Lammps('lmp_binary_morse')
lmp.potential = morse_Al + morse_Al_Ni + morse_Ni
lmp.structure = pr.create.structure.bulk('Al', cubic=True)
lmp.structure[1] = "Ni"
lmp.run()
```

## Case III: Hybrid potential
E.g. EAM + Morse

### Unary overlay

```python
eam = Library("Al")
morse = Morse("Al", D_0=0.5, alpha=1.1, r_0=2.1, cutoff=6)
lmp = pr.create.job.Lammps('lmp_unary_eam_morse')
lmp.potential = eam + morse
lmp.structure = pr.create.structure.bulk('Al', cubic=True)
lmp.run()
```

### Binary overlay

Again, all combinations must be specified
```python
eam = Library("Al")
morse_Al_Ni = Morse("Al", "Ni", D_0=0.2, alpha=1.05, r_0=2.2, cutoff=6)
morse_Ni = Morse("Ni", D_0=0.7, alpha=1.15, r_0=2.15, cutoff=6)

lmp = pr.create.job.Lammps('lmp_binary_eam_morse')
lmp.potential = eam + morse_Al_Ni + morse_Ni
lmp.structure = pr.create.structure.bulk('Al', cubic=True)
lmp.structure[0] = "Ni"
lmp.run()
```

## Addendum: custom potential
```python
custom_pot = CustomPotential("lj/cut", "Al", epsilon=0.5, sigma=1, cutoff=3)
lmp = pr.create.job.Lammps('lmp_unary_custom')
lmp.potential = custom_pot
lmp.structure = pr.create.structure.bulk('Al', cubic=True)
lmp.run()
```

To do:

- [x] DocStrings
- [ ] Tests
- [ ] More potentials (not this PR)
- [ ] Creator on the project level (probably not this PR)
- [x] Specify EAM via `int` (not this PR - sorry @liamhuber) EDIT: stacked in #1012 

Closes #997


pyiron_atomistics/lammps/potentials.py
```diff
@@ -0,0 +1,414 @@
...
+            pair_style += f" {s[0]}"
+            if s[1] > 0:
+                pair_style += f" {s[1]}"
+        return pair_style + "\n"
+
+    @property
+    def pair_coeff(self) -> list:
+        """LAMMPS pair_coeff"""
+
+        class PairCoeff:
```
niklassiemer: Why do you define the `PairCoeff` inside the property? This way it is always refefinded on accessing the property.

niklassiemer (commented):
I am not really familiar with the potentials we can set in pyiron. Therefore, I just quickly skimmed over :)

pyiron_atomistics/lammps/potentials.py
samwaseda: Yeah I was not very sure about how to write it properly. Initially, there was just one property function `pair_coeff` (the one you referred to), but then I realized that it was getting too long. In my attempt to refactor the function, I decided to make a separate class for this purpose, and since it's only used for `pair_coeff`, I put it there inside, to make people easily understand that it's not to be used anywhere else. It's possible that it's not the right way to do so, but I'm not good enough in coding to know what's the right way. You can also tell me how to do it better or @pmrv or @liamhuber also has a good insight?

niklassiemer: Why not simply define it at the module level and use it in the property? If you do not want to use it outside of module, you could define it as private class `class _LammpsPairCoeff:` or the like.

samwaseda: I'm ok with that. Since the efficiency hardly plays a role here, I just thought I'd put it within the function because it's even clearer that it's used only within the function.

pyiron_atomistics/lammps/potentials.py
```diff
@@ -0,0 +1,414 @@
...
+__author__ = "Joerg Neugebauer, Sudarsan Surendralal, Jan Janssen"
+__copyright__ = (
+    "Copyright 2021, Max-Planck-Institut für Eisenforschung GmbH - "
+    "Computational Materials Design (CM) Department"
+)
+__version__ = "1.0"
+__maintainer__ = "Sudarsan Surendralal"
+__email__ = "surendralal@mpie.de"
+__status__ = "production"
+__date__ = "Sep 1, 2017"
```
liamhuber:
```suggestion
__author__ = "Sam Waseda"
__copyright__ = (
    "Copyright 2023, Max-Planck-Institut für Eisenforschung GmbH - "
    "Computational Materials Design (CM) Department"
)
__version__ = "1.0"
__maintainer__ = "Sam Waseda"
__email__ = "waseda@mpie.de"
__status__ = "production"
__date__ = "April 18, 2023"
```

pyiron_atomistics/lammps/potentials.py
```diff
@@ -0,0 +1,414 @@
...
+    @property
+    def df(self):
+        """DataFrame containing all info for each pairwise interactions"""
+        return self._df
+
+    def get_df(self, default_scale=None):
+        if default_scale is None or "scale" in self.df:
+            return self.df.copy()
+        df = self.df.copy()
+        df["scale"] = 1
```
liamhuber: The `default_scale` arg is not used unless it has its default value... Was this the intent?:

```suggestion
        df["scale"] = default_scale
```

pyiron_atomistics/lammps/potentials.py
```diff
@@ -0,0 +1,414 @@
...
+import numpy as np
+import warnings
+
+
+class LammpsPotentials:
+    def __new__(cls, *args, **kwargs):
+        obj = super().__new__(cls)
+        cls._df = None
+        return obj
+
```
liamhuber: Please consider sliding all the methods around. E.g. I find it very useful to read `__init__` right after I've read `__new__` to see why `__new__` was needed to start with (or whatever other weirdness is driving the need for this atypical dunder); `copy` holds a position of high honour being the first named method yet it calls the df getter and setter which we don't see forever; a bunch of properties get listed high up (which I like), but they mostly reference `self.df` and the `df` property is way down at the bottom; `_harmonize_args` is private and appears right near the top, but then the other private method (`_initialize_df`) doesn't appear until the very, very end. Overall it is just kind of chaos.

There is no perfect way to sort things, and not all of these are always satisfiable at once, but I try to do the following:
- Inity stuff has priority 1 (`__init__`, `__new__`, and `_initialize_df` here, probably)
- Abstract methods have priority 2 (none here)
- Then public-facing properties, then public-facing methods, then private stuff, and finally overrides of magic methods (actually, I'm pretty loose with where the dunder overrides go...)
- Methods/variables with no/fewer internal dependencies get placed higher (e.g. the `df` property before the `name` property)
- Dependencies as soon as possible after they're used (e.g. `set_df` right after `_initialize_df`)
- Group highly-interdependent/related stuff (e.g. `get_df` and `copy` right after `set_df`)

Basically all stuff to guide the reader and minimize scrolling around the file

pyiron_atomistics/lammps/potentials.py
```diff
@@ -0,0 +1,414 @@
...
+    def copy(self):
+        new_pot = LammpsPotentials()
+        new_pot.set_df(self.get_df())
+        return new_pot
+
+    @staticmethod
+    def _harmonize_args(args) -> str:
```
liamhuber: This could use a little docstring for the dev explaining why I want to always have two args even if they're the same. It might also be nice to rename the variable like `*species_symbols`. The `*` unpacking will work regardless of what you name the var, so you might as well use it to carry some information.

pyiron_atomistics/lammps/potentials.py
liamhuber: >  it's even clearer that it's used only within the function.

This is a solid justification, but I still lean to @niklassiemer's suggestion to pull it out as a private class; this way it is much easier to test independently.

pyiron_atomistics/lammps/potentials.py
```diff
@@ -0,0 +1,414 @@
...
+                        results.append("")
+                return results
+
+            @property
+            def pair_style(self):
+                """pair_style to be output only in hybrid"""
+                if self.is_hybrid:
+                    return self._pair_style
+                else:
+                    return len(self._pair_style) * [""]
```
liamhuber:
```suggestion
                return self._pair_style if self.is_hybrid else len(self._pair_style) * [""]
```

If this breaks line length limits just leave the explicit if/else clause you have; it's just hard to tell in the github editor

pyiron_atomistics/lammps/potentials.py
```diff
@@ -0,0 +1,414 @@
...
+        return self._df
+
+    def get_df(self, default_scale=None):
+        if default_scale is None or "scale" in self.df:
+            return self.df.copy()
+        df = self.df.copy()
+        df["scale"] = 1
+        return df
+
+    def __mul__(self, scale_or_potential):
```
liamhuber: I see that one forces scales to be 1 and the other doesn't, but I don't quickly have a sense of why I care -- a quick docstring(s) so I can understand the difference between `LammpsPontentials()*LammpsPotentials()` and `LammpsPontentials() + LammpsPotentials()` would be helpful, because they both just concatenate...

pyiron_atomistics/lammps/potentials.py
```diff
@@ -0,0 +1,414 @@
...
+        scale=None,
+        cutoff=None,
+    ):
+        def check_none_n_length(variable, default, length=len(pair_coeff)):
+            if variable is None:
+                variable = default
+            if not isinstance(variable, list):
+                return variable
+            if len(variable) == 1 and len(variable) < length:
+                variable = length * variable
```
liamhuber:
```suggestion
            if isinstance(variable, list) and len(variable) == 1 and len(variable) < length:
                variable = length * variable
```

liamhuber (commented):
Nice. I think I found a bug where `default_scale` is not used, but otherwise just some requests and suggestions to improve readability.

pyiron_atomistics/lammps/potentials.py
samwaseda: Yeah thanks!

pyiron_atomistics/lammps/potentials.py
samwaseda: Yeah it does.

pyiron_atomistics/lammps/potentials.py
samwaseda: Yeah that's true. The difference is explained in [the issue](https://github.com/pyiron/pyiron_atomistics/issues/997):

- `pot_1 * pot_2` -> `hybrid`
- `pot_1 + pot_2` -> `hybrid/overlay`.
- `0.5 * pot_1 + 0.9 * pot_2` -> `hybrid/scaled`

The choice of + and * for `hybrid` and `hybrid/overlay` somewhat makes sense, because something like `pot_1 * 0.3 * pot_2` should not work. This being said, I'm not particularly happy with `pot_1 * pot_2` for `hybrid`, so if you have a better idea let me know. Anyway I'm gonna add docs for now.

liamhuber:
Ah, but now you're in trouble @samwaseda! You should have just added the index choice and I never would have pulled this branch and discovered this terrible flaw 😉 

The base class seems like it's probably ok, but there is something going wrong with the fancy representation, as this raises an error when you allow the jupyter cell to render the potential but are fine when you suppress rendering with a `;`

```python
from pyiron_atomistics.lammps.potentials import EAM
EAM("Mg", "Al", "Zn")
```

```
...
File ~/work/pyiron/pyiron_atomistics/pyiron_atomistics/lammps/potentials.py:312, in LammpsPotentials._initialize_df(self, pair_style, interacting_species, pair_coeff, preset_species, model, citations, filename, name, scale, cutoff)
    310 if scale is not None:
    311     arg_dict["scale"] = scale
--> 312 self.set_df(pd.DataFrame(arg_dict))
...
ValueError: All arrays must be of the same length
```

Can you reproduce this or is it some strange flaw in my jupyter setup? I am certain I have such a ternary potential, and there were some other combos where I was able to generate the same error.

samwaseda:
> The base class seems like it's probably ok, but there is something going wrong with the fancy representation, as this raises an error when you allow the jupyter cell to render the potential but are fine when you suppress rendering with a `;`
> 
> ```python
> from pyiron_atomistics.lammps.potentials import EAM
> EAM("Mg", "Al", "Zn")
> ```

Oh no it doesn't work with MEAM...

And I totally forgot that this class sets the potential as soon as the dataframe is to be generated, which happens to be the case when `__repr__` is called. So the following lines should work:

```python
from pyiron_atomistics.lammps.potentials import EAM
eam = EAM("Mg", "Al", "Zn")
eam.view_potentials()
```

samwaseda:
Since it doesn't work with MEAM, I make it draft for now.

samwaseda:
@liamhuber now it should be working (it was a relatively simple fix)

samwaseda:
I replaced `EAM` by `Library`, because there are also other potential types (e.g. `tersoff`, `MEAM`) in the potential library.

liamhuber (approved):
Tests are still on the TODO list, but otherwise lgtm!

samwaseda:
This PR is supposed to be moved to `pyiron_contrib`, but until I figure it out I continue developing it here.

jan-janssen:
@samwaseda Why was this merged? I thought the last decision we made in the pyiron meeting was to merge it in `pyiron_contrib` rather than `pyiron_atomistics`.

samwaseda:
Because there are only two lines. They also don't change any of the current functionalities. If these two lines are not introduced, going from `pyiron_contrib` to `pyiron_atomistics` later will be more difficult.

PR #1012, opened 2023-04-18
More eam inits

liamhuber:
Allow users to pass a structure for chemical elements -- this augments the `*chemical_elements` args, and I think it will be super useful for people who want to check if there is an EAM potential available for their structure should they want to add X to it, i.e. `EAM("X", structure=my_struct)`.

Allow users to select among candidates by index. Because the heart wants what the heart wants.

### DIFF
--- a/pyiron_atomistics/lammps/potentials.py
+++ b/pyiron_atomistics/lammps/potentials.py
@@ -458,7 +458,7 @@ class Library(LammpsPotentials):
                     f"available choices."
                 )
             df = self._df_candidates.iloc[choice]
-            if len(self._df_candidates) > 1:
+            if self._choice is None and len(self._df_candidates) > 1:
                 warnings.warn(
                     f"Potential not uniquely specified - use default {df.Name}"
                 )
--- a/pyiron_atomistics/lammps/potentials.py
+++ b/pyiron_atomistics/lammps/potentials.py
@@ -13,6 +13,8 @@ __email__ = "surendralal@mpie.de"
 __status__ = "production"
 __date__ = "Sep 1, 2017"
 
+from typing import Optional
+
 import pandas as pd
 from pyiron_atomistics.lammps.potential import LammpsPotentialFile
 import numpy as np
@@ -358,7 +360,12 @@ class EAM(LammpsPotentials):
                 )
         return
 
-    def __init__(self, *chemical_elements, name=None, pair_style=None):
+    def __init__(
+            self,
+            *chemical_elements: str,
+            name: Optional[str] = None,
+            pair_style: Optional[str] = None
+    ):
         if name is not None:
             self._df_candidates = LammpsPotentialFile().find_by_name(name)
         else:
--- a/pyiron_atomistics/lammps/potentials.py
+++ b/pyiron_atomistics/lammps/potentials.py
@@ -16,6 +16,7 @@ __date__ = "Sep 1, 2017"
 from typing import Optional
 
 import pandas as pd
+from pyiron_atomistics.atomistics.structure.atoms import Atoms
 from pyiron_atomistics.lammps.potential import LammpsPotentialFile
 import numpy as np
 import warnings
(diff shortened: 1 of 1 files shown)


pyiron_atomistics/lammps/potentials.py
```diff
@@ -358,11 +361,21 @@ def _get_scale(config):
                 )
         return
 
-    def __init__(self, *chemical_elements, name=None, pair_style=None):
+    def __init__(
+            self,
+            *chemical_elements: str,
+            choice: int = 0,
```
samwaseda: Not sure if `choice` should be preset. I would write `choice: Optional[int] = None`, so that the warning can be safely silenced when `choice=0` is explicitly chosen.

pyiron_atomistics/lammps/potentials.py
```diff
@@ -373,7 +386,12 @@ def view_potentials(self):
...
-            df = self._df_candidates.iloc[0]
+            if self._choice > len(self._df_candidates):
+                raise ValueError(
+                    f"Cannot choose {self._choice} among {len(self._df_candidates)} "
+                    f"available choices."
+                )
+            df = self._df_candidates.iloc[self._choice]
             if len(self._df_candidates) > 1:
                 warnings.warn(
                     f"Potential not uniquely specified - use default {df.Name}"
```
samwaseda: In relation with my comment above, I would issue this warning only if `choice` is not explicitly set.

samwaseda (commented):
I agree with the changes. The only thing is that I see these changes are purely extensions to the existing code (and nothing replaces the current/proposed functionality). If I know @jan-janssen well, he prefers PRs to contain minimum changes, which I disagree with, but I respect his opinion, so I would propose to merge these changes to `main` once `potential` gets merged.

pyiron_atomistics/lammps/potentials.py
liamhuber: That's a good improvement. Done.

liamhuber:
> If I know @jan-janssen well, he prefers PRs to contain minimum changes, which I disagree with, but I respect his opinion, so I would propose to merge these changes to main once potential gets merged.

@samwaseda, the purpose of these minimal-change PRs is to (a) improve reviewability, and (b) have a clearer history of the changes. So, as with all [stacked PRs](https://blog.logrocket.com/using-stacked-pull-requests-in-github/), we can choose to merge its root and then immediately merge this, or we can merge this into the root and then merge the whole thing at once. Critical is only that all individual PRs are finished review and fully compatible before we start reviewing. I mildly prefer to merge this into #1008 and then merge them together into main just to make sure all the functionality arrives at once.

EDIT: But note also that GitHub will cleverly re-base this PR onto main if the `potential` branch gets merged first!

samwaseda:
> @samwaseda, the purpose of these minimal-change PRs is to (a) improve reviewability, and (b) have a clearer history of the changes. So, as with all [stacked PRs](https://blog.logrocket.com/using-stacked-pull-requests-in-github/), we can choose to merge its root and then immediately merge this, or we can merge this into the root and then merge the whole thing at once. Critical is only that all individual PRs are finished review and fully compatible before we start reviewing. I mildly prefer to merge this into #1008 and then merge them together into main just to make sure all the functionality arrives at once.

Hmmm as long as @jan-janssen is happy, I'm ok with anything, because I still think mostly the total amount to review the PRs is the same (i.e. one PR split into two just requires more or less twice the amount of time required to review each sub-PR). I just see that with stacked PRs the fact doesn't change that there's one PR that has a lot of changed lines, correct? Or is there a possibility to see something like "these lines were already reviewed in the stacked PR"?

liamhuber:
> Hmmm as long as @jan-janssen is happy...I just see that with stacked PRs the fact doesn't change that there's one PR that has a lot of changed lines, correct? Or is there a possibility to see something like "these lines were already reviewed in the stacked PR"?

I mean, I don't want to put too many words in Jan's mouth here, but I really do think the big two things are (a) making reviews more manageable and (b) having a clear and readable history. The `branch->root, root->main` and  `root->main, branch->main` paradigms both do a good job of satisfying these.

In this case, `branch->root, root->main` is probably slightly less readable than `root->main, branch->main`, as in the former case you're correct that all the line changes in `branch` get duplicated in `root`. I don't think it's disasterously bad as the overall commit history is still clear and you can pretty easily see on github that `branch` got merged into `root`. For an example, see [here on ironflow](https://github.com/pyiron/ironflow/pulls?page=3&q=is%3Apr+is%3Aclosed) looking at PRs 107 (`branch`) and 106 (`root`). The _upside_ to `branch->root, root->main` is that it's still quite clear and allows you to get the entire feature set tested before merging anything. But anyhow, as long as the comma in `root->main, branch->main` does not represent too much time, I really am 100% fine with this paradigm too!

> I still think mostly the total amount to review the PRs is the same

I strongly disagree; IMO this falls into the fundamental category of "problems that are more manageable when you break them into smaller pieces" -- a broad and important category!

PR #1025, opened 2023-04-27
Murnaghan

### LINKED ISSUE #1014: Input / Output murnaghan - possible unit confusion
When we do a murn-fit like 
```# [e0, b0, bP, v0], [e0_error, b0_error, bP_error, v0_error] = murn._fit_leastsq(volume_lst, energy_lst,fittype='vinet')```
we have the units in eV, GPa, and Ang^3. 

However, when using the vinet equation as implemented (as in ase/eos.py)
```def vinet_energy(V, E0, B0, BP, V0):
    "Vinet equation from PRB 70, 224107"
    eta = (V / V0) ** (1 / 3)
    return E0 + 2 * B0 * V0 / (BP - 1) ** 2 * (
        2 - (5 + 3 * BP * (eta - 1) - 3 * eta) * np.exp(-3 * (BP - 1) * (eta - 1) / 2)
    )
```

The `B0` in there is defined as in ase in eV/Ang^3. The result is that one has to add a unit conversion factor as in 
https://pyiron.readthedocs.io/en/latest/source/notebooks/energy_volume_curve.html (eq. [15]).

I think this unit conversion is a bit confusing and one has to be careful in distinguishing units of variables for different modules / functions....

samwaseda:
Closes #1014 

Unit conversion of bulk modulus is now included.


samwaseda:
ah obvious mistake. I close this one

PR #1026, opened 2023-04-27
Make a proper unit conversion of bulk modulus in Murnaghan

### LINKED ISSUE #1014: Input / Output murnaghan - possible unit confusion

samwaseda:
Closes https://github.com/pyiron/pyiron_atomistics/issues/1014

Unit conversion of bulk modulus is now included.

### DIFF
--- a/pyiron_atomistics/atomistics/master/murnaghan.py
+++ b/pyiron_atomistics/atomistics/master/murnaghan.py
@@ -602,17 +602,17 @@ class EnergyVolumeFit(object):
     @staticmethod
     def birchmurnaghan_energy(V, E0, B0, BP, V0):
         "BirchMurnaghan equation from PRB 70, 224107"
-        return birchmurnaghan_energy(V, E0, B0, BP, V0)
+        return birchmurnaghan_energy(V, E0, B0 / eV_div_A3_to_GPa, BP, V0)
 
     @staticmethod
     def vinet_energy(V, E0, B0, BP, V0):
         "Vinet equation from PRB 70, 224107"
-        return vinet_energy(V, E0, B0, BP, V0)
+        return vinet_energy(V, E0, B0 / eV_div_A3_to_GPa, BP, V0)
 
     @staticmethod
     def murnaghan(V, E0, B0, BP, V0):
         "From PRB 28,5480 (1983"
-        return murnaghan(V, E0, B0, BP, V0)
+        return murnaghan(V, E0, B0 / eV_div_A3_to_GPa, BP, V0)
 
     @staticmethod
     def birch(V, E0, B0, BP, V0):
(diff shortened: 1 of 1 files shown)


niklassiemer:
Did we use these functions anywhere in our code base? Not that we apply the conversion twice in that case. 

And they were used in some notebooks, weren't they?

samwaseda:
> Did we use these functions anywhere in our code base? Not that we apply the conversion twice in that case.
> 
> And they were used in some notebooks, weren't they?

Good point, and indeed I missed it inside the file.

Although I have to say it's totally unclear what was meant to be visible to the outside and what wasn't in this file. The whole thing is a mess.

jan-janssen:
This changes the data stored in the `Murnaghan` job. 
```
 def _fit_eos_general(self, vol_erg_dic=None, fittype="birchmurnaghan"):
        self._set_fit_module(vol_erg_dic=vol_erg_dic)
        fit_dict = self.fit_module.fit_eos_general(fittype=fittype)
        self.input["fit_type"] = fit_dict["fit_type"]
        self.input["fit_order"] = 0
        self._store_fit_in_hdf(fit_dict=fit_dict)
        self.fit_dict = fit_dict
        return fit_dict
```
Short - it breaks the backwards compatibility.

samwaseda:
> Short - it breaks the backwards compatibility.

I need a bit more detail to understand why

jan-janssen:
> > Short - it breaks the backwards compatibility.
> 
> I need a bit more detail to understand why

You change is what is stored in the `fit_dict` so that changes the `Murnaghan` class.

samwaseda:
> You change is what is stored in the `fit_dict` so that changes the `Murnaghan` class.

I don't understand where I changed it.

jan-janssen:
> > You change is what is stored in the `fit_dict` so that changes the `Murnaghan` class.
> 
> I don't understand where I changed it.

I see - that was my mistake - it is only set in line 528.

jan-janssen (approved):
Looks good to me

samwaseda:
Since the original problem is bypassed with [this PR](https://github.com/pyiron/pyiron/pull/1479), I close this PR.

PR #1030, opened 2023-05-05
Add an error message indicating SPHInX is not installed

samwaseda:
Based on [this discussion](https://github.com/pyiron/pyiron/pull/1479) with @liamhuber, I added an error message in SPHInX (Liam's inner voice: "we were talking about GPAW!!").

I put it in the log parser, because that's where the error is usually shown first, but to be honest I find is a bit clumsy, because it's more the symptom and not the error itself. If there's someone who has a better idea, let me know.

### DIFF
--- a/pyiron_atomistics/sphinx/base.py
+++ b/pyiron_atomistics/sphinx/base.py
@@ -1569,21 +1569,7 @@ class SphinxBase(GenericDFTJob):
 
             return len(w) == 0
 
-    def validate_ready_to_run(self):
-        """
-        Checks whether parameters are set appropriately. It does not
-        mean the simulation won't run if it returns False.
-        """
-
-        if shutil.which("sphinx") is None:
-            raise ValueError(
-                "It looks like you have not installed SPHInX on your system, "
-                "which is not included in the pyiron conda package. You can "
-                "install SPHInX via\n\nconda install -c conda-forge "
-                "sphinxdft\n\n. For more info, take a look at our page: "
-                "https://pyiron.readthedocs.io/en/latest/"
-            )
-
+    def _validate_input_ready_to_run(self):
         all_groups = {
             "job.input.pawPot": self.input.sphinx.pawPot,
             "job.input.structure": self.input.sphinx.structure,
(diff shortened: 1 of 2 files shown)


pmrv:
I solved a similar problem with [atomsk](https://github.com/pyiron/pyiron_atomistics/blob/f6007ccb38172a4cd766ff2a7e0614d9cefa6f4a/pyiron_atomistics/atomistics/structure/factories/atomsk.py#L27) using `shutil.which`.  I guess if you do that check in the sphinx top level module or in `SphinxBase.__init__` you can fail earlier.

samwaseda:
> I solved a similar problem with [atomsk](https://github.com/pyiron/pyiron_atomistics/blob/f6007ccb38172a4cd766ff2a7e0614d9cefa6f4a/pyiron_atomistics/atomistics/structure/factories/atomsk.py#L27) using `shutil.which`. I guess if you do that check in the sphinx top level module or in `SphinxBase.__init__` you can fail earlier.

!? God Marvin you totally knocked my eyes off!!!

samwaseda:
But where's `_ATOMSK_EXISTS` actually used?

pmrv:
It's used to not define the atomsk factory if the binary is not around, see [here](https://github.com/pyiron/pyiron_atomistics/blob/f6007ccb38172a4cd766ff2a7e0614d9cefa6f4a/pyiron_atomistics/atomistics/structure/factory.py#LL35C20-L35C20) and [here](https://github.com/pyiron/pyiron_atomistics/blob/f6007ccb38172a4cd766ff2a7e0614d9cefa6f4a/pyiron_atomistics/atomistics/structure/factory.py#L73).  For sphinx that's probably a bit much, because even when sphinx is not installed as an executable, users might want to analyze calculations run elsewhere.

liamhuber:
Is it not possible to raise these sorts of errors with the import guard Marvin made? I'd also suggest to directly put the "conda install -c conda-forge (package_name)" right in the error, at least for packages where we're the conda feedstock maintainters.

samwaseda:
Yeah right now I'm looking at the best possible place for this error to be raised.

This being said, it sounds also a bit like a general feature that we should have anyway. Currently SPHInX for example raises an error only when it fails to parse the output file, and it also only says that it failed to parse the output file...

samwaseda:
Actually there was a super obvious location: `validate_ready_to_run`.

pyiron_atomistics/sphinx/base.py
```diff
@@ -1827,7 +1836,7 @@ def run_addon(
                 # move output to working directory for successful runs
                 if out.returncode == 0:
                     for file in os.listdir(tempd):
-                        movefile(os.path.join(tempd, file), self.working_directory)
+                        shutil.movefile(os.path.join(tempd, file), self.working_directory)
```
jan-janssen:
```suggestion
                        shutil.move(os.path.join(tempd, file), self.working_directory)
```

samwaseda: ah brilliant

PR #1036, opened 2023-05-09
Pickle atoms

liamhuber:
I'd like to be able to pickle atoms for the contrib workflow stuff. Right now the code below throws a recursion error, with this PR it works fine:

```python
import pickle

from pyiron_atomistics.atomistics.structure.factory import StructureFactory
struct = StructureFactory().bulk("Al")

with open("my_stored_structure", mode="wb") as f:
    pickle.dump(struct, f)
```

And then in a new interpreter instance:
```python
import pickle

with open("my_stored_structure", mode="rb") as f:
    loaded_struct = pickle.load(f)

loaded_struct.plot3d()
```

@pmrv, you are my storage expert -- am I being super naive here, or was specifying `__setstate__` really all we were missing??

PR #1037, opened 2023-05-10
Just use ase

liamhuber:
Over in #1036, @jan-janssen pointed out that `ase.Atoms` are already pickle-able, so this just adds more incentive to get switched over to using those.

I suspect this is a non-trivial change, and I think we should discuss it at the pyiron meeting this week, but it's an idea I really like.

As a sort of experiment to ground the discussion a bit, here I just naively deleted everything in our `atoms.py` and `atom.py` files and replaced them with simple imports of `ase.Atoms` and `ase.Atom`, respectively.

On my local machine, this causes 67 test failures. This supports my earlier claim that such a conversion is not _totally trivial_, but I haven't gone through the tests in detail and I suspect a bunch of them are _almost trivial_. E.g. the first one I looked at imports and uses `CrystalStructure`, which is just a wrapper class around `Atoms` and the docstring just reads, I kid you not, "For historical reasons." So maybe a bunch of these tests will be easy to fix?

Anyhow, I don't really intend to merge this, I just wanted a good jumping-off point for discussion.


liamhuber:
I went through and got all the tests in `tests/structure/factories` working by adjusting the code in `pyiron_atomistics/atomistics/structure/factory.py` and the `factories/` subdirectory, along with some of the test code itself.

In the factories it was basically just a matter of getting rid of all the ase-to-pyiron conversions, making the conversions from other codes to pyiron stop half-way and just be conversions to ase, and replacing non-existent stuff like instances of the `CrystalStructure` class with things that exists like the output of the `.bulk` creator.

In the tests it was mostly transforming things to arrays (ase often gives lists instead) and directly calling analysis methods instead of calling them off the structure objects themselves.

This did serve to drive home to me how convenient our `Atoms` class was in terms of all the methods we had slapped onto it. We will need to think carefully about how to keep these at least close to as-accessible if we strip `Atoms` down to `ase.Atoms`.

niklassiemer:
We are not _forced_ to use the plain `ase.Atoms`, are we?
I think that we could after all have 'our' `Atoms` object, however, a lot closer like a tiny layer on top, e.g.
```
from ase import Atoms as _ASEAtoms
from structuretoolkit import get_neighbors

class Atoms(_ASEAtoms):
    def get_neighbors(self, *args):
        return get_neighbors(self, *args)
```
Or convenience conversions to `np.arrays` for some properties using `super`

pmrv:
> We are not _forced_ to use the plain `ase.Atoms`, are we? I think that we could after all have 'our' `Atoms` object, however, a lot closer like a tiny layer on top, e.g.
> 
> ```
> from ase import Atoms as _ASEAtoms
> from structuretoolkit import get_neighbors
> 
> class Atoms(_ASEAtoms):
>     def get_neighbors(self, *args):
>         return get_neighbors(self, *args)
> ```
> 
> Or convenience conversions to `np.arrays` for some properties using `super`

Imo that's how it should be.

liamhuber:
> We are not forced to use the plain `ase.Atoms`, are we?
I think that we could after all have 'our' `Atoms` object, however, a lot closer like a tiny layer on top

No, of course not. I do think it's a worthwhile exercise to see how much needs to be changed in order to get strictly onto `ase.Atoms` though.

I also wonder whether in tests it might be preferable to do things like `from structuretoolkit import get_neighbours; ...; get_neighbours(struct)` rather than `struct.analyze.get_neighbours()` -- assuming that we're not trying to test the `Analyze` interface, that is. With the idea being to test the smallest pieces we can at once, and not be making demands about the overall interface of `Atoms` when that's not what we're trying to test. My concern is that this might quickly get into "overkill".

> Or convenience conversions to `np.arrays` for some properties using `super`

This would absolutely be convenient, but then we really do need to think carefully whether we want it. If we strictly layer additional convenience attributes/methods on top of `ase.Atoms` I don't really see how anything can go wrong. Actually modifying returned types definitely makes so the two classes are different in ways that may be dangerous/have side effects.

PR #1040, opened 2023-05-11
fix quasiharmonic job

### LINKED ISSUE #1039: QuasiHarmonicJob is broken
```
File ~/PycharmProjects/pyiron_base/pyiron_base/jobs/job/generic.py:1226, in GenericJob._run_if_created(self)
   1217 def _run_if_created(self):
   1218     """
   1219     Internal helper function the run if created function is called when the job status is 'created'. It executes
   1220     the simulation, either in modal mode, meaning waiting for the simulation to finish, manually, or submits the
   (...)
   1224         int: Queue ID - if the job was send to the queue
   1225     """
-> 1226     return run_job_with_status_created(job=self)

File ~/PycharmProjects/pyiron_base/pyiron_base/jobs/job/runfunction.py:99, in run_job_with_status_created(job)
     97     job.run_if_manually(_manually_print=False)
     98 elif job.server.run_mode.modal:
---> 99     job.run_static()
    100 elif job.server.run_mode.srun:
    101     job.run_if_srun()

File ~/PycharmProjects/pyiron_base/pyiron_base/jobs/master/parallel.py:529, in ParallelMaster.run_static(self)
    527 self._logger.info("{} run parallel master (modal)".format(self.job_info_str))
    528 self.status.running = True
--> 529 self.submission_status.total_jobs = len(self._job_generator)
    530 self.submission_status.submitted_jobs = 0
    531 if self.job_id and not self.is_finished():

File ~/PycharmProjects/pyiron_base/pyiron_base/jobs/master/parallel.py:842, in JobGenerator.__len__(self)
    841 def __len__(self):
--> 842     return len(self.parameter_list_cached)

File ~/PycharmProjects/pyiron_base/pyiron_base/jobs/master/parallel.py:790, in JobGenerator.parameter_list_cached(self)
    787 @property
    788 def parameter_list_cached(self):
    789     if len(self._parameter_lst_cached) == 0:
--> 790         self._parameter_lst_cached = self.parameter_list
    791     return self._parameter_lst_cached

File ~/PycharmProjects/pyiron_atomistics/pyiron_atomistics/atomistics/master/murnaghan.py:342, in MurnaghanJobGenerator.parameter_list(self)
    336 """
    337 
    338 Returns:
    339     (list)
    340 """
    341 parameter_lst = []
--> 342 strains = self._master.input.get("strains")
    343 if strains is None:
    344     strains = np.linspace(
    345         -self._master.input["vol_range"],
    346         self._master.input["vol_range"],
    347         int(self._master.input["num_points"]),
    348     )

File ~/PycharmProjects/pyiron_base/pyiron_base/storage/parameters.py:378, in GenericParameters.get(self, parameter_name, default_value)
    376     return default_value
    377 else:
--> 378     raise NameError("parameter not found: " + parameter_name)

NameError: parameter not found: strains
```
This happened when the `Murnaghan` class was modified to require strains: 
https://github.com/pyiron/pyiron_atomistics/pull/816

jan-janssen:
Closes https://github.com/pyiron/pyiron_atomistics/issues/1039

### DIFF
--- a/pyiron_atomistics/atomistics/master/quasi.py
+++ b/pyiron_atomistics/atomistics/master/quasi.py
@@ -96,6 +96,14 @@ class QuasiHarmonicJob(AtomisticParallelMaster):
         self.input["temperature_end"] = 500
         self.input["temperature_steps"] = 10
         self.input["polynomial_degree"] = 3
+        self.input["strains"] = (
+            None,
+            "List of strains that should be calculated.  If given vol_range and num_points take no effect.",
+        )
+        self.input["axes"] = (
+            ["x", "y", "z"],
+            "Axes along which the strain will be applied",
+        )
         self._job_generator = MurnaghanJobGenerator(self)
 
     def collect_output(self):


pmrv:
Does it make sense for users to modify the axis/strains on the level of the quasi harmonic approximation job?  If so, I'm ok with this solution.  

I guess the awkward thing here is that the murnaghan job generator is used by two master jobs, but it must, by design, assume explicit details of the master job it is assigned.  I wasn't aware we actually use the same generator for multiple masters that are not in a inheritance relationship.  Is that the only case?

jan-janssen:
> Does it make sense for users to modify the axis/strains on the level of the quasi harmonic approximation job? If so, I'm ok with this solution.

It is mainly for compatibility - if I do not have the `axis` it fails with a very similar error. 

> Is that the only case?

Given the small number of `GenericMasters` we have I think this is fair to assume. Still this is where an IDE can be helpful to show places where a given code is used.

pmrv:
> > Does it make sense for users to modify the axis/strains on the level of the quasi harmonic approximation job? If so, I'm ok with this solution.
> 
> It is mainly for compatibility - if I do not have the `axis` it fails with a very similar error.

I see that, but the alternative fix would be to change the murnaghan generator to handle `None` or non-existing entries in its master's input.  That would leave the user input of the quasi-harmonic job unchanged.  So it comes down to which interface we want and that's why I asked this.

jan-janssen:
> I see that, but the alternative fix would be to change the murnaghan generator to handle `None` or non-existing entries in its master's input. That would leave the user input of the quasi-harmonic job unchanged. So it comes down to which interface we want and that's why I asked this.

The Quasi-harmonic Job requires an `Murnaghan` job as input to calculate the free energies as discussed in https://pyiron.org/phasediagram-workshop-2020/Exercise4.html . So consequently it should support the same functionality. I am not sure if we need the functionality for the `Murnaghan` job as I primarily use uniform displacements, but I am open to allow non-uniform displacements. Still if we do, then we should also make sure that the features remain consistent over the whole pyiron project. Consequently, the alternative to me would be reverting https://github.com/pyiron/pyiron_atomistics/pull/816

pmrv (approved):

PR #1053, opened 2023-05-23
Remove Atoms().visualize

jan-janssen:
As plot3d() is now part of the structure toolkit, there is no longer a need for a separate visualize class


jan-janssen:
Waiting for https://github.com/pyiron/pyiron_atomistics/pull/1056

PR #1104, opened 2023-07-06
Separate `get_structure_group` to fix SxExtOpt bug when `selective_dynamics` is on

samwaseda:
SxExtOpt did not work when `selective_dynamics` was on anymore. This was because `write_structure` does not exist anymore in SPHInX.


samwaseda:
I merge this one because it was a rather minor issue

PR #1105, opened 2023-07-06
Remove SSA which is not used and not maintained

samwaseda:
There is currently anyway no possibility to combine SSA and SxExtOpt.


jan-janssen:
Does this affect the functionality we published in https://github.com/pyiron/spin_space_averaging ?

samwaseda:
> Does this affect the functionality we published in https://github.com/pyiron/spin_space_averaging ?

No because it’s not based on SxExtOpt

PR #1106, opened 2023-07-06
Allow some Murnaghan children to fail

pmrv:
I sometimes don't care if actually all the strains in a Murnaghan calculation work out, so I added an extra flag to allow some failures.  By default it's still zero.  I would have preferred to make this general for all masters, but since the collection logic is not central I can't really do that easily.

### DIFF
--- a/pyiron_atomistics/atomistics/master/murnaghan.py
+++ b/pyiron_atomistics/atomistics/master/murnaghan.py
@@ -832,9 +832,13 @@ class Murnaghan(AtomisticParallelMaster):
                 err_lst.append(np.var(energy))
                 vol_lst.append(volume)
                 id_lst.append(job_id)
-            aborted_children = self.input.get('allow_aborted') - allowed_aborted_children
+            aborted_children = (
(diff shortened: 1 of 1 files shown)


samwaseda:
In the input doc it's written aborted or not converged, but as far as I can see it checks only for aborted.

pmrv:
Right, thanks.  At first I also checked for not converged children but then I realized that murnaghan handles those gracefully already.

pyiron_atomistics/atomistics/master/murnaghan.py
```diff
@@ -679,6 +679,10 @@ def __init__(self, project, job_name):
             None,
             "List of strains that should be calculated.  If given vol_range and num_points take no effect.",
         )
+        self.input["allow_aborted"] = (
+            0,
+            "The number of child jobs that are allowed to fail or not converge, before this job is considered aborted or not converged.",
```
pmrv:
```suggestion
            "The number of child jobs that are allowed to abort, before the whole job is considered aborted.",
```

PR #1107, opened 2023-07-07
Lammps: Add method to save per atom kinetic energy

### LINKED ISSUE #1102: Output the kinetic energy of each atom from LAMMPS calculations using pyiron
**Output the kinetic energy of each atom from LAMMPS calculations using pyiron**

I am looking into the energy of each atom in a structure from the LAMMPS output. One can obtain this by

compute pe all ke/atom
dump [....] x y z c_ke

**We can use job.input.control.energy_pot_per_atom() to output the potential energy of each atoms. Can we also output the kinetic energy of each atom using pyiron?  Thanks a lot in advance.**

pmrv:
This is simply analogous to `energy_pot_per_atom`.  An alternative (with less storage req) might be to compute the kinetic energy after the simulation from the velocities (which we save anyway).  Not really knowing my way around lammps however, I'd prefer the pedestrian way.

@lfzhu-phys Can you give this a try?  I'm afraid it would take some time for this to be available by default on the cluster, because we still have conda problems there.  Do you know how to set up a custom branch of pyiron?

Closes #1102

### DIFF
--- a/pyiron_atomistics/lammps/control.py
+++ b/pyiron_atomistics/lammps/control.py
@@ -764,11 +764,20 @@ class LammpsControl(GenericParameters):
 
     def energy_pot_per_atom(self):
         """
-        Enable the output of atomic energies.  This will add an additional key 'energy_pot_per_atom' to the HDF output.
+        Enable the output of per atom potential energies.  This will add an additional key 'energy_pot_per_atom' to the HDF output.
         """
         if self["compute___energy_pot_per_atom"] is None:
             self["compute___energy_pot_per_atom"] = "all pe/atom"
             self["dump___1"] += " c_energy_pot_per_atom"
+            self["dump_modify___1"] = self["dump_modify___1"][:-1] + ' %20.15g"'
+
+    def energy_kin_per_atom(self):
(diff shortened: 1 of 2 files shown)


lfzhu-phys:
@pmrv, thanks, I will try to compute the kinetic energy from velocities.
-->Do you know how to set up a custom branch of pyiron?
Sorry, no:(

pmrv:
> @pmrv, thanks, I will try to compute the kinetic energy from velocities. -->Do you know how to set up a custom branch of pyiron? Sorry, no:(

Ah, schade.  For a stop gap you should be able to put

```python
job.input.control["compute___energy_kin_per_atom"] = "all ke/atom"
job.input.control["dump___1"] += " c_energy_kin_per_atom"
job.input.control["dump_modify___1"] = self["dump_modify___1"][:-1] + ' %20.15g"'
```

into your notebook before you run the simulations.  The output will appear automatically under `job['output/generic/energy_kin_per_atom']`.

lfzhu-phys:
@pmrv Great, that's super helpful:) Thanks a lot!

PR #1121, opened 2023-07-25
Murnaghan: Round children names

pmrv:
When the job name of the murnaghan is long; using ~10 digits for the strain can blow the job name limit. So I added some rounding here.

The [PR](https://github.com/pyiron/pyiron_base/pull/1173) helps, but doesn't allow to influence the rounding.  Since `Murnaghan` job knows how much to safely round, this PR takes advantage of this here.

Alternatively one could drop the murnaghan job name from the names of the children.  It's redundant, but I thought it might be a larger change, in that users will see their jobs changing.  I feel ambivalent about this though.

### DIFF
--- a/pyiron_atomistics/atomistics/master/murnaghan.py
+++ b/pyiron_atomistics/atomistics/master/murnaghan.py
@@ -19,6 +19,7 @@ import numpy as np
 from pyiron_atomistics.atomistics.master.parallel import AtomisticParallelMaster
 from pyiron_atomistics.atomistics.structure.atoms import Atoms, ase_to_pyiron
 from pyiron_base import JobGenerator
+from pyiron_base.jobs.job.util import _get_safe_job_name
 
 __author__ = "Joerg Neugebauer, Jan Janssen"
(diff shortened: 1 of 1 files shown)


samwaseda:
I'm actually not sure if we should define `job_name` in the first place in `Murnaghan`. At least I don't see why we shouldn't simply use the default `job_name` from `JobGenerator` everywhere.

pmrv:
Am open to that as well.  I kind of like seeing the strain in the job name, though, because you tell quickly which ones e.g. failed.

samwaseda:
But is it really possible to see the problem by looking at the strain value?

pmrv:
Not often, but I just find it generally informative.

samwaseda:
Ok then at least I would like to make the job name function accessible from outside, because I might want to get the data based on the strain values.

pmrv:
> Ok then at least I would like to make the job name function accessible from outside, because I might want to get the data based on the strain values.

Not sure what you mean.  So that I can set a function from outside that determines the child names?

PR #1126, opened 2023-08-02
[minor] Refactor potential setter

samwaseda:
I refactored the potential setter because it was too long.


samwaseda:
What I did must have been purely refactoring, so if the tests pass I'll merge it.

PR #1130, opened 2023-08-17
Refactor the lammps collect_output() parser

jan-janssen:
With this new refactoring we can now run LAMMPS inside functions: 
```python
import subprocess
from pyiron_base.jobs.job.util import _copy_restart_files


def initialize_job(job): 
    job.validate_ready_to_run()
    job.project_hdf5.create_working_directory()
    job.write_input()
    _copy_restart_files(job=job)

def execute_subprocess(job):
    executable, shell = job.executable.get_input_for_subprocess_call(
        cores=job.server.cores, threads=job.server.threads, gpus=job.server.gpus
    )
    out = subprocess.run(
        executable,
        cwd=job.project_hdf5.working_directory,
        shell=shell,
        stdout=subprocess.PIPE,
        stderr=subprocess.STDOUT,
        universal_newlines=True,
        check=True,
    ).stdout

def execute_job(job): 
    initialize_job(job=job)
    execute_subprocess(job=job)
    return job.collect_output_parser()


pr = Project("test")
job = pr.create.job.Lammps(job_name="lmp")
job.structure = pr.create.structure.ase.bulk("Al", cubic=True)
output_dict = execute_job(job=job)
```
Input files are written, the executable is called and the output is parsed. But there is no connection to the pyiron database or the HDF5 storage.

### DIFF
--- a/pyiron_atomistics/lammps/base.py
+++ b/pyiron_atomistics/lammps/base.py
@@ -454,19 +454,26 @@ class LammpsBase(AtomisticGenericJob):
             }
         }
 
-    def collect_output_parser(self):
-        dump_h5_file_name = self.job_file_name(
-            file_name="dump.h5", cwd=self.working_directory
+    def collect_output_parser(
+        self,
+        cwd,
+        dump_h5_file_name="dump.h5",
+        dump_out_file_name="dump.out",
+        log_lammps_file_name="log.lammps",
+    ):
+        # Parse output files
+        dump_h5_full_file_name = self.job_file_name(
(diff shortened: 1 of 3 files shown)


jan-janssen:
I still have to work on refactoring the output so it becomes a nice dictionary which represents the structure of the HDF5 file. Still the general principle remains the same. The new `collect_output_parser()` function parses the output files and returns a dictionary with the output extracted from the HDF5 file. So in most cases this output dictionary only has to be saved to the HDF5 file.

PR #1131, opened 2023-08-17
Refactor the VASP collect_output() parser

jan-janssen:
Example:
```python
import subprocess
from pyiron_atomistics import Project
from pyiron_base.jobs.job.util import _copy_restart_files

def initialize_job(job): 
    job.validate_ready_to_run()
    job.project_hdf5.create_working_directory()
    job.write_input()
    _copy_restart_files(job=job)

def execute_subprocess(job):
    executable, shell = job.executable.get_input_for_subprocess_call(
        cores=job.server.cores, threads=job.server.threads, gpus=job.server.gpus
    )
    out = subprocess.run(
        executable,
        cwd=job.project_hdf5.working_directory,
        shell=shell,
        stdout=subprocess.PIPE,
        stderr=subprocess.STDOUT,
        universal_newlines=True,
        check=True,
    ).stdout

def execute_job(job): 
    initialize_job(job=job)
    execute_subprocess(job=job)
    return job.collect_output_parser(cwd=job.working_directory)


pr = Project("test")
job = pr.create.job.Vasp("vasp")
job.structure = pr.create.structure.ase.bulk("Al", cubic=True)
output_dict = execute_job(job)
output_dict
```

PR #1139, opened 2023-08-22
Use HasHDF in various Vasp objects

pmrv:


pmrv:
This does nothing but rewriting the `{to,from}_hdf` from #1131 to the "new" `_{to,from}_hdf` from `HasHDF`.

PR #1151, opened 2023-09-05
Only structuretoolkit depends on pymatgen not pyiron_atomistics

jan-janssen:


jan-janssen:
The corresponding tests are transferred to `structuretoolkit` https://github.com/pyiron/structuretoolkit/pull/63

PR #1159, opened 2023-09-07
SPHINX eigen values parser 

ahmedabdelkawy:
1. With the current Pyiron sphinx parser, the eigenvalues from the eps.*.dat files are not parsed

- in pyiron_atomistics/pyiron_atomistics/sphinx/base.py The collect function gives priority to the final eigen values in the log file ```self.collect_sphinx_log(file_name="sphinx.log", cwd=directory)``` before  ``` self.collect_eps_dat()``` 

```python
        def collect(self, directory=os.getcwd()):
        """
        The collect function, collects all the output from a SPHInX simulation.

        Args:
            directory (str): the directory to collect the output from.
        """
        self.collect_energy_struct(file_name="energy-structOpt.dat", cwd=directory)
        self.collect_sphinx_log(file_name="sphinx.log", cwd=directory)
        self.collect_energy_dat(file_name="energy.dat", cwd=directory)
        self.collect_residue_dat(file_name="residue.dat", cwd=directory)
        if self._job._spin_enabled:
            self.collect_eps_dat(file_name="eps.dat", cwd=directory)
        else:
            self.collect_eps_dat(
                file_name=[f"eps.{i}.dat" for i in [0, 1]], cwd=directory
            )
        self.collect_spins_dat(file_name="spins.dat", cwd=directory)
        self.collect_relaxed_hist(file_name="relaxHist.sx", cwd=directory)
        self.collect_electrostatic_potential(file_name="vElStat-eV.sxb", cwd=directory)
        self.collect_charge_density(file_name="rho.sxb", cwd=directory)
```
- This is then followed by 

``` python
    def collect_eps_dat(self, file_name="eps.dat", cwd=None):
        """

        Args:
            file_name:
            cwd:

        Returns:

        """
        if isinstance(file_name, str):
            file_name = [file_name]
        values = []
        for f in file_name:
            file_tmp = posixpath.join(cwd, f)
            if not os.path.isfile(file_tmp):
                return
            values.append(np.loadtxt(file_tmp)[..., 1:])
        values = np.stack(values, axis=0)
        if "bands_eigen_values" not in self.generic.dft.list_nodes():
            self.generic.dft.bands_eigen_values = values.reshape((-1,) + values.shape)
``` 
- Which does not import values from the eps.*.dat files because it already imported the values from the log file

2. The problem with importing eigenvalues from the logfile is that pyiron expects to have the eigenvalues of all kpoints from the 1st spin channel (in case of spin-polarized calculations) followed by the eigenvalues at all kpoint from the second spin channel. Which happens here:
``` python
    def _parse_band(self, term):
        arr = np.loadtxt(re.findall(term, self.log_main, re.MULTILINE))
        shape = (-1, len(self.k_points), arr.shape[-1])
        if self.spin_enabled:
            shape = (-1, 2, len(self.k_points), shape[-1])
        return arr.reshape(shape)
``` 
- However, in sphinx, this is true from the eps.*.dat but from the log file, the eigenvalues are printed for spin up and then spin down for each kpoint in alternating fashion.

3. This appears as a problem only in the case of ferromagnetic systems (metals) where the 2 spin channels are different (while in nonmagnetic systems you have one spin channel and in antiferromagnetic systems you two identical but opposite dos)

4. I did not change the current parser priority. However, I changed the order of the parsed eigenvalues to match what pyiron expects: arr = np.vstack((arr[::2], arr[1::2]))


pyiron_atomistics/sphinx/base.py
```diff
@@ -2286,6 +2286,7 @@ def _parse_band(self, term):
         arr = np.loadtxt(re.findall(term, self.log_main, re.MULTILINE))
         shape = (-1, len(self.k_points), arr.shape[-1])
         if self.spin_enabled:
+            arr = np.vstack((arr[::2], arr[1::2]))
```
freyso, line 2289: Please add a comment in the code why this is needed. Otherwise, this will cause a lot of headache for later devels.

samwaseda:
I'm actually a bit concerned that it may or may not work depending on the SPHInX version? I'm saying this because it was correct in an earlier version and stopped being so with the current version. Maybe we need a version check inside the parser.

freyso:
Sphinx has not changed here, I believe the loop order in the writing routine dates back from 2002.
 
I suspect rather that the pyiron parser has never has been validated in this point.
 
But please do so this time, Ahmed!
 

     

-------------------------------------------------
Stay up to date and follow us on LinkedIn, Twitter and YouTube.

Max-Planck-Institut für Eisenforschung GmbH
Max-Planck-Straße 1
D-40237 Düsseldorf
 
Handelsregister B 2533 
Amtsgericht Düsseldorf
 
Geschäftsführung
Prof. Dr. Gerhard Dehm
Prof. Dr. Jörg Neugebauer
Prof. Dr. Dierk Raabe
Dr. Kai de Weldige
 
Ust.-Id.-Nr.: DE 11 93 58 514 
Steuernummer: 105 5891 1000


Please consider that invitations and e-mails of our institute are 
only valid if they end with ***@***.*** 
If you are not sure of the validity please contact ***@***.***

Bitte beachten Sie, dass Einladungen zu Veranstaltungen und E-Mails
aus unserem Haus nur mit der Endung ***@***.*** gültig sind. 
In Zweifelsfällen wenden Sie sich bitte an ***@***.***
-------------------------------------------------

--EdoMail64fb3410_7a6d7337_241f
Content-Type: text/html; charset="utf-8"
Content-Transfer-Encoding: quoted-printable
Content-Disposition: inline

<div id=3D=22edo-message=22><div><div style=3D=22font-size:16px=22>Sphinx=
 has not changed here, I believe the loop order in the writing routine da=
tes back from 2002.</div><div style=3D=22font-size:16px=22>I suspect rath=
er that the pyiron parser has never has been validated in this point.</di=
v><div style=3D=22font-size:16px=22>But please do so this time, Ahmed=21<=
/div><div style=3D=22font-size:16px=22><br></div></div></div>
--EdoMail64fb3410_7a6d7337_241f--

pmrv:
Ninja comment, but it sounds like this can be more readably solved by `arr.reshape(-1, 2)` or something related. /smokebomb

Issue #1184, opened 2023-10-10
Atomistics causing downstream tests to fail because periodic table can't be found


liamhuber:
In [pyiron_workflow](https://github.com/pyiron/pyiron_workflow/actions/runs/6462643394/job/17544577877) and [pyiron_contrib](https://github.com/pyiron/pyiron_contrib/actions/runs/6462625756/job/17544534657) we now get errors ending like:

```
File "/usr/share/miniconda3/envs/my-env/lib/python3.11/site-packages/pyiron_atomistics/atomistics/structure/factory.py", line 167, in bulk
    return self.ase.bulk(
           ^^^^^^^^^^^^^^
  File "/usr/share/miniconda3/envs/my-env/lib/python3.11/site-packages/pyiron_atomistics/atomistics/structure/factories/ase.py", line 49, in wrapper
    return func(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^
  File "/usr/share/miniconda3/envs/my-env/lib/python3.11/site-packages/pyiron_atomistics/atomistics/structure/factories/ase.py", line 60, in bulk
    return ase_to_pyiron(ase_bulk(*args, **kwargs))
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/share/miniconda3/envs/my-env/lib/python3.11/site-packages/pyiron_atomistics/atomistics/structure/atoms.py", line 3111, in ase_to_pyiron
    pyiron_atoms = Atoms(
                   ^^^^^^
  File "/usr/share/miniconda3/envs/my-env/lib/python3.11/site-packages/pyiron_atomistics/atomistics/structure/atoms.py", line 122, in __init__
    self._pse = PeriodicTable()
                ^^^^^^^^^^^^^^^
  File "/usr/share/miniconda3/envs/my-env/lib/python3.11/site-packages/pyiron_atomistics/atomistics/structure/periodic_table.py", line 218, in __init__
    self.dataframe = self._get_periodic_table_df(file_name)
                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/share/miniconda3/envs/my-env/lib/python3.11/site-packages/pyiron_atomistics/atomistics/structure/periodic_table.py", line 419, in _get_periodic_table_df
    raise ValueError("Was not able to locate a periodic table. ")
ValueError: Was not able to locate a periodic table.
```

[...]

PR #1185, opened 2023-10-10
Depend on pyiron data

liamhuber:
The basic functionality of `pyiron_atomistics` should work out of the box when we use the package as a dependency in downstream packages. Currently, you can't instantiate `Atoms` without also specifying a periodic table in your resources directory, cf. #1184.

Here, I include `pyiron-data` as a dependency, which resolves the issue above and I feel rather strongly about it (or getting some other equivalent solution). 

I also took a knife to `tests/static` directory and cut out everything that vaguely looked like it might come in the [`pyiron-data`](https://github.com/pyiron/pyiron-resources) dependency. I don't feel strongly about this part at all, I might have broken things. This was just a "what the heck" change. I tried to do this in tiny commits so we can revert any individual ones that wind up breaking things.


setup.py
```diff
@@ -54,6 +54,7 @@
         'phonopy==2.20.0',
         'pint==0.22',
         'pyiron_base==0.6.7',
+        'pyiron-data==0.0.22',
```
jan-janssen: There is no `pyiron-data` package on PYPI.

liamhuber: Right, I guess it's not exactly .py code is it

liamhuber:
Errors fall into two categories:

Vasp totally failing to find the potentials

```
ERROR: setUpClass (atomic.structure.test_potentials.TestVaspPotential)
  File "/home/runner/work/pyiron_atomistics/pyiron_atomistics/tests/atomic/structure/test_potentials.py", line 74, in setUpClass
    cls.vasp = VaspPotential()
  File "/home/runner/work/pyiron_atomistics/pyiron_atomistics/pyiron_atomistics/vasp/potential.py", line 237, in __init__
    self.pbe = VaspPotentialFile(xc="PBE", selected_atoms=selected_atoms)
  File "/home/runner/work/pyiron_atomistics/pyiron_atomistics/pyiron_atomistics/vasp/potential.py", line 161, in __init__
    potential_df = self._get_potential_df(
  File "/home/runner/work/pyiron_atomistics/pyiron_atomistics/pyiron_atomistics/atomistics/job/potentials.py", line 154, in _get_potential_df
    raise ValueError("Was not able to locate the potential files.")
ValueError: Was not able to locate the potential files.
```

I don't know off the top of my head how we're handling resource paths or why vasp is breaking like this (other resources seem to still be found...)

And sphinx finding the wrong potentials this one time:
```
FAIL: test_potential (sphinx.test_base.TestSphinx)
  File "/home/runner/work/pyiron_atomistics/pyiron_atomistics/tests/sphinx/test_base.py", line 127, in test_potential
    self.assertEqual(['Fe_GGA'], self.sphinx_2_5.list_potentials())
AssertionError: Lists differ: ['Fe_GGA'] != ['Ni_GGA', 'Fe_GGA']

First differing element 0:
'Fe_GGA'
'Ni_GGA'

Second list contains 1 additional elements.
First extra element 1:
'Fe_GGA'

- ['Fe_GGA']
+ ['Ni_GGA', 'Fe_GGA']
```

The sphinx test failure looks super familiar. Also, indeed, the purged potential file had only Fe, while [pyiron resources has many](https://github.com/pyiron/pyiron-resources/blob/main/sphinx/potentials/potentials_sphinx.csv). I haven't looked at the test yet, but I suspect this is reasonable and understandable.

liamhuber:
Discovered in the sphinx test file:

```python
        # Thoughts by C. Freysoldt, 2022-10-24:
        #   - I think this is unhealthy, as the test environment is NOT controlled by the test
        #   - there is a discrepancy between mybinder/pyiron_atomistics environment and the
        #     github/CI environment.
```

Indeed @freyso, indeed! 😂 😭

liamhuber:
> The sphinx test failure looks super familiar. Also, indeed, the purged potential file had only Fe, while [pyiron resources has many](https://github.com/pyiron/pyiron-resources/blob/main/sphinx/potentials/potentials_sphinx.csv). I haven't looked at the test yet, but I suspect this is reasonable and understandable.

Indeed, the failing line is `self.assertEqual(['Fe_GGA'], self.sphinx_2_5.list_potentials())`, and 
```python
cls.sphinx_2_5.structure = Atoms(
            elements=["Fe", "Ni"],
            scaled_positions=[[0.0, 0.0, 0.0], [0.5, 0.5, 0.5]],
            cell=2.83 * np.eye(3),
        )
```
So I think it _should_ show both Fe and Ni, and the only reason it didn't is that before our very limited sphinx potentials resources didn't have _any_ potential for Ni. So this failure makes perfect sense. I'm going to stack a PR on here resolving this sphinx test.

liamhuber:
> I don't know off the top of my head how we're handling resource paths or why vasp is breaking like this (other resources seem to still be found...)

Indeed, both sphinx and vasp are relying on [the same static method](https://github.com/pyiron/pyiron_atomistics/blob/5a0686dd6168f6d22013fcbcfd240fcadb665a82/pyiron_atomistics/atomistics/job/potentials.py#L104) too look for their potentials, and the potential path [in pyiron-data](https://github.com/pyiron/pyiron-resources/tree/main/vasp/potentials) looks to me like the same one that was previously used in `tests/static/`, so I don't know why vasp is failing to find its potentials when sphinx succeeds. 

Ah, nevermind, I think it's because the vasp test [is overriding the resource paths entirely](https://github.com/pyiron/pyiron_atomistics/blob/5a0686dd6168f6d22013fcbcfd240fcadb665a82/tests/vasp/test_potential.py#L15) instead of extending it. I'm not 100% sure the potential csv will play nicely with actual potential files in a different location but following the same resource pattern, but let's try it and find out.

liamhuber:
Backwards compatibility failure:

```
Running tests/backwards/test_sphinx_load.py
/usr/share/miniconda3/envs/my-env/lib/python3.10/site-packages/pandas/core/computation/expressions.py:21: UserWarning: Pandas requires version '2.8.0' or newer of 'numexpr' (version '2.7.3' currently installed).
  from pandas.core.computation.check import NUMEXPR_INSTALLED

  0%|          | 0/2 [00:00<?, ?it/s]
  0%|          | 0/2 [00:00<?, ?it/s]
  File "/usr/share/miniconda3/envs/my-env/lib/python3.10/site-packages/pyiron_atomistics/sphinx/base.py", line 2454, in collect_sphinx_log
    self._spx_log_parser = _SphinxLogParser(log_file)
  File "/usr/share/miniconda3/envs/my-env/lib/python3.10/site-packages/pyiron_atomistics/sphinx/base.py", line 2136, in __init__
    self._check_enter_scf()
  File "/usr/share/miniconda3/envs/my-env/lib/python3.10/site-packages/pyiron_atomistics/sphinx/base.py", line 2166, in _check_enter_scf
    raise AssertionError("Log file created but first scf loop not reached")
AssertionError: Log file created but first scf loop not reached
```

@samwaseda is that meaningful to you?

liamhuber:
Oh my goodness, they're failing because `pyiron-data` [doesn't actually include the vasp stuff](https://github.com/conda-forge/pyiron-data-feedstock/blob/main/recipe/build.sh).

Well, I still think we ought to be appending to rather than replacing the state resources, but at least it makes sense why it continues to find the vasp potentials file. I'll revert that commit.

liamhuber:
```
FAIL: test_list_potentials (vasp.test_vasp.TestVasp)
  File "/home/runner/work/pyiron_atomistics/pyiron_atomistics/tests/vasp/test_vasp.py", line 70, in test_list_potentials
    self.assertEqual(sorted([
AssertionError: Lists differ: ['Fe', 'Fe_GW', 'Fe_pv', 'Fe_sv', 'Fe_sv_GW', 'O', [53 chars]_GW'] != ['Fe', 'Fe', 'Fe_GW', 'Fe_GW', 'Fe_pv', 'Fe_pv', 'F[162 chars]_GW']

First differing element 1:
'Fe_GW'
'Fe'

Second list contains 13 additional elements.
First extra element 13:
'O_GW'

  ['Fe',
+  'Fe',
+  'Fe_GW',
   'Fe_GW',
   'Fe_pv',
+  'Fe_pv',
+  'Fe_sv',
   'Fe_sv',
   'Fe_sv_GW',
+  'Fe_sv_GW',
+  'O',
   'O',
   'O_GW',
+  'O_GW',
+  'O_GW_new',
   'O_GW_new',
   'O_h',
+  'O_h',
+  'O_s',
   'O_s',
   'O_s_GW',
+  'O_s_GW',
   'Se',
+  'Se',
+  'Se_GW',
   'Se_GW']
```

Somehow the potentials are getting double counted

liamhuber:
> Somehow the potentials are getting double counted

Yes, when a path appears twice in the resource paths, the things found there show up twice here.

I can also revert the append-instead-of-overwrite commits, since [resource paths are getting the system paths added back anyhow](https://github.com/pyiron/pyiron_base/blob/be280c42b00bd21cf3290f247c4847351fafb12f/pyiron_base/state/settings.py#L168C32-L168C32), and that's what I was worried about overwriting.

liamhuber:
Ok, super, the only failing tests now are the sphinx one that is resolved in the stacked #1186, and the backwards-compatibility sphinx failure I don't understand. Ball is in your court for both of these, @samwaseda.

More generally, I do feel that the vasp tests still suffer from the same malaise that [Christoph commented on](https://github.com/pyiron/pyiron_atomistics/pull/1185#issuecomment-1756159701) where they rely on particular environment factors which they neither directly control nor which are controlled by a dependency -- rather we just need to make sure files in `tests/static/vasp` stay up-to-date with what's here. Since there are legal reasons for not distributing vasp stuff in `pyiron-data`, I guess I am content for now to just suffer this badness.

Needs #1186, then merging the combination of this and that should close #1184 once a new version is released.

niklassiemer:
Just one grain of salt from my side to this. I am not really worried about having static test-resources for things like `list_potentials`. In my perspective it 'only' needs better documentation and probably activating these test resources only for the relevant tests. I.e. putting the test suite in charge to change the environment where needed. Otherwise we might not be able to test things (we cannot include real VASP potentials due to license in pyiron-data) or we get failures while our code does the right thing since someone (nicely) added a new potential and thus it fails having the wrong entires... 
Since the resources are completely up to the (super)user to define and we only offer a sensible default, I see this rather like (meta)data to be used while testing.
We could think about having a test-resources repo to be checked out prior to testing in all relevant repos?

liamhuber:
> Just one grain of salt from my side to this. I am not really worried about having static test-resources for things like `list_potentials`. In my perspective it 'only' needs better documentation and probably activating these test resources only for the relevant tests. I.e. putting the test suite in charge to change the environment where needed. Otherwise we might not be able to test things (we cannot include real VASP potentials due to license in pyiron-data) or we get failures while our code does the right thing since someone (nicely) added a new potential and thus it fails having the wrong entires...
> Since the resources are completely up to the (super)user to define and we only offer a sensible default, I see this rather like (meta)data to be used while testing.
> We could think about having a test-resources repo to be checked out prior to testing in all relevant repos?

Agreed, for the sake of the test suite I'm also less concerned. The big problem for me is that without this PR, if you conda install `pyiron_atomistics`, you can't even make a structure!!! 

Since we need to include some sort of standard resources in the conda dependencies anyway, I think those resources (`pyiron-data`) can be the test suite we checkout for most purposes. For the rest (e.g. vasp) I'm fine with both a second repo we check out or just doing it in a bespoke way right in the repo if there's only one repo that needs it. 

I take Christoph's point well, though, that the tests don't _control_ these resources, so it's up to us to keep things working together. Your comment about better documenting this synchronization is a good one! E.g., over in #1186 I rely on sphinx potentials having the element symbol in their title, so I just explicitly state this assumption in the test message.

niklassiemer:
> Agreed, for the sake of the test suite I'm also less concerned. The big problem for me is that without this PR, if you conda install `pyiron_atomistics`, you can't even make a structure!!! 
> 
I absolutely agree, that's a no-go! 
> Since we need to include some sort of standard resources in the conda dependencies anyway, I think those resources (`pyiron-data`) can be the test suite we checkout for most purposes. For the rest (e.g. vasp) I'm fine with both a second repo we check out or just doing it in a bespoke way right in the repo if there's only one repo that needs it. 
> 
Fine with that as long as we do make sure that we use a consistent e.g. potential (by name, not by first element) to check numeric values.
> I take Christoph's point well, though, that the tests don't _control_ these resources, so it's up to us to keep things working together. Your comment about better documenting this synchronization is a good one! E.g., over in #1186 I rely on sphinx potentials having the element symbol in their title, so I just explicitly state this assumption in the test message.

liamhuber:
Should be resolved by #1191

PR #1186, opened 2023-10-10
Update the sphinx potential tests

liamhuber:
To rely on the sphinx potentials on pyiron-data, which is now a conda-dependency.


liamhuber:
The branch this is based on got closed, but it might be worth rebasing onto main and working on. With the new sphinx PRs coming up, it reminded me to bump this issue with the test at you @samwaseda

liamhuber:
This would rely on `pyiron-data` being in the dependencies used for the unit tests, but there we have only `environment.yml` and `pyiron-data` appears only in `environement-notebooks.yml`. I still like the idea of reducing the size of the `static` directory, but relying on `pyiron-data` comes with its own pain points and at a minimum requires changes to the CI. 

@samwaseda, I'm going to ping you again so you can take a peek at the changes once more to see if there's anything you want to incorporate into future versions of the sphinx tests, but otherwise this PR is done IMO.

PR #1195, opened 2023-10-26
Calphy: Set npt input flag

pmrv:
Currently calphy complains if this value is set to None.

A work around is to set the flag manually after you call `calc_*`.

```python
job.calc_free_energy(...)
job.input.npt = True # if you set a pressure else False
```


srmnitc:
I applied a patch for this issue on the condaforge package with the workshop in mind. Its also fixed in calphy, but I am still waiting for a release because it will have breaking changes. Nevertheless, I agree that we merge this in. A bit of extra protection would definitely help.

srmnitc (approved):

PR #1200, opened 2023-11-08
Spare a poor phd student from the horrible numpy ragged array problem

samwaseda:


ahmedabdelkawy:
Code to reproduce problem:

```
from pyiron import Project
pr = Project('fe')

vasp_job = pr.create.job.Vasp('Fe_magnetic', delete_existing_job=True)
vasp_job.structure = pr.create.structure.bulk('Fe', 'bcc', a=2.8, cubic=True)
vasp_job.set_encut(450)
vasp_job.set_kpoints([10,10,10])
vasp_job.input.incar["ISPIN"]=2 
vasp_job.calc_minimize()
vasp_job.input.incar["ISIF"] = "3"
vasp_job.input.incar["MAGMOM"]= "5 5"
vasp_job.input.incar["NCORE"] = '1'
vasp_job.executable = '5.4.4'
vasp_job.server.cores = 5
vasp_job.run(delete_existing_job=True, run_mode='queue')
```

Ouput OUTCAR
[OUTCAR.txt](https://github.com/pyiron/pyiron_atomistics/files/13310931/OUTCAR.txt) (remove.txt)

- Parsing the magnetization from minimization which generates different number of magnetization values for each electronic step per each ionic step is the problem. 
- make sure to use python/latest kernel to reproduce the problem (numpy = 1.26.0).

ahmedabdelkawy:
Two quantities (maybe more) are parsed in outcar.py that have different row sizes (number of electronic steps) for each column (number of ionic steps): scf_energy_free and magnetization. Both in outcar.py are parsed as a list of arrays with different shapes. Something then happens in base.py that makes scf_energy_free work (saved as a numpy array of dtype=object) while magnetization does not. The current fix now explicitly saves the magnetization as an object which is required by numpy. Since these quantities are not yet used in any post-processing, it is okay to keep them as objects.

PR #1210, opened 2023-11-15
Patch emmet missing custodian

### LINKED ISSUE #1207: [conda package] `emmet` cannot find `custodian` when `pyiron_atomistics` is imported
The daily workflows for `pyiron_atomistics` (I also look at `pyiron_workflow`, `pyiron` and `pyiron_base`) are failing with errors like `from custodian.vasp.jobs import VaspJob; ModuleNotFoundError: No module named 'custodian'`.

Following the suggestion of @liamhuber , I started to hunt this down. Please feel free to point out anything that screams "well that's not really how things work" to your face, as I really want to get a detailed understanding.

- I checked CI and (if applicable) notebook runs for `pyiron`, `pyiron_base`, `pyiron_atomistics` and `pyiron_workflow`. Errors are indeed consistently caused by `emmet` failing to find `custodian`.
- I looked at the recipe's [meta.yaml](https://github.com/conda-forge/emmet-core-feedstock/blob/main/recipe/meta.yaml) at [emmet-core-feedstock](https://github.com/conda-forge/emmet-core-feedstock/tree/main). No mention of `custodian` there.
- I looked at [this log](https://github.com/pyiron/pyiron_atomistics/actions/runs/6847656793/job/18616330651#step:9:37) from `pyiron_atomistics`. Indeed, they import via `from custodian.vasp.jobs ...` and thus depend implicitly on `custodian`.
- In the very same log, you can see that [`emmet-core` version `0.72.20`](https://github.com/pyiron/pyiron_atomistics/actions/runs/6847656793/job/18616330651#step:7:91) is installed. From the error mentioned before, you can see the already mentioned dependency on `emmet.core` via `mp-api`. `mp-api` is an [explicit dependency of the `pyiron_atomistics` package](https://github.com/conda-forge/pyiron_atomistics-feedstock/blob/60677585ab97f4fc04dab27d96232fdfe5dbbabe/recipe/meta.yaml#L29). In the environemnt of the failed run, [`mp-api =0.37.5`](https://github.com/pyiron/pyiron_atomistics/actions/runs/6847656793/job/18616330651#step:7:159) is installed. In [mp-api's feedstock](https://github.com/conda-forge/mp-api-feedstock/blob/7f4e88a0d7ea101ca96a37d4b48f3c7fa9493b2f/recipe/meta.yaml#L32) you can see the explicit dependence on `emmet-core` but, of course, non on `custodian`.

So - if I'm understanding all of this correctly - the solution would be to put `custodian` as an explicit dependency in the `meta.yaml` in `emmet-core-feedstock`, right? At this point I just don't know **where** to raise an issue for this: in the feedstock repo? In the [`emmet-core` repo](https://github.com/materialsproject/emmet)?

mbruns91:
The materialsproject's `emmet-core`, on which we depend via `mp-api`, is missing to declare its dependency on `custodian`. This causes our CI and notebook runs to fail, with error messages like `from custodian.vasp.jobs import VaspJob; ModuleNotFoundError: No module named 'custodian'`. This patch provides a hotfix for this.

Maybe this shouldn't be merged, as I will also raise an issue on [`emmet`'s](https://github.com/materialsproject/emmet) repo, as it's a cleaner solution if they declare this dependency explicitly.

### DIFF
--- a/.ci_support/environment.yml
+++ b/.ci_support/environment.yml
@@ -7,6 +7,7 @@ dependencies:
 - coveralls
 - coverage
 - codacy-coverage
+- custodian =2023.10.9
 - defusedxml =0.7.1
 - h5py =3.10.0
 - matplotlib-base =3.8.1
--- a/binder/environment.yml
+++ b/binder/environment.yml
@@ -17,3 +17,4 @@ dependencies:
 - pyiron-data >=0.0.22
 - sqsgenerator
 - structuretoolkit
+- custodian
--- a/fd
+++ b/fd
@@ -1,0 +1,1 @@
+/home/bruns/.local/bin/fd
--- a/fd
+++ b/fd
@@ -1,1 +1,0 @@
-/home/bruns/.local/bin/fd
--- a/pyiron_atomistics/vasp/base.py
+++ b/pyiron_atomistics/vasp/base.py
@@ -2019,8 +2019,12 @@ class Output:
                 "n_elect"
             ]
             if len(self.outcar.parse_dict["magnetization"]) > 0:
-                magnetization = np.array(self.outcar.parse_dict["magnetization"], dtype=object)
-                final_magmoms = np.array(self.outcar.parse_dict["final_magmoms"], dtype=object)
+                magnetization = np.array(
+                    self.outcar.parse_dict["magnetization"], dtype=object
+                )
+                final_magmoms = np.array(
+                    self.outcar.parse_dict["final_magmoms"], dtype=object
+                )
                 # magnetization[sorted_indices] = magnetization.copy()
                 if len(final_magmoms) != 0:
                     if len(final_magmoms.shape) == 3:
--- a/setup.py
+++ b/setup.py
@@ -45,6 +45,7 @@ setup(
         'ase==3.22.1',
         'atomistics==0.0.6',
         'defusedxml==0.7.1',
+        'custodian==2023.10.09',
         'h5py==3.10.0',
         'matplotlib==3.8.1',
         'mendeleev==0.14.0',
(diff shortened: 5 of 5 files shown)


niklassiemer (approved):
LGTM, will fix the unittests for this repo :) for the other repos, however, this will probably not yet fix the issue since they rely on the conda package if I recall correctly... thus, the feedstock would need a new build...

liamhuber (approved):
LGTM too.

> for the other repos, however, this will probably not yet fix the issue since they rely on the conda package if I recall correctly... thus, the feedstock would need a new build...

Yes, for sure. The feedstock will need a new release no matter what for the other repos to start working -- the only question is whether we want to make _one_ new release (waiting for `emmet-core` to get a fixed release and then depend on that), or _two_ new releases (merge and release this now, then again when `emmet-core` is fixed).

On the one hand, the materialsproject folks were super responsive the only other time I interacted with them, but on the other hand I'm _suuuper_ tired of the tests failing... How about this: raise the issue on materialsproject/emmet, and if they fix it before Monday then we can stop worrying and make a single new version, but if they still haven't gotten to it then we can use the regular meeting to make sure the rest of the team is OK with a hotfix adding such an indirect dependency

mbruns91:
Sorry, I was kind of busy today due to unforeseen circumstances. To accelerate this a bit, I opened an [issue](https://github.com/materialsproject/emmet/issues/890) in the `emmet-core` repo and also made a [PR in the feedstock](https://github.com/conda-forge/emmet-core-feedstock/pull/149).

mbruns91:
This got resolved via https://github.com/materialsproject/emmet/pull/884 and our CI seems to be running well again.

PR #1211, opened 2023-11-16
use trusted publisher action

jan-janssen:


jan-janssen:
Waiting for `pyiron_base` update.

PR #1212, opened 2023-11-16
Update interactive.py

samwaseda:
The problem is still there, but I don't think it's worth issuing the warning every time.


jan-janssen (approved):
Looks good to me

PR #1218, opened 2023-11-23
Give installation warning when no library potential found

samwaseda:


niklassiemer (commented):
Thanks for looking into this! 
However, If I now use some weird structure with a mixture of rare earth elements or the like, I probably find myself without a valid potential?!

samwaseda:
> Thanks for looking into this! However, If I now use some weird structure with a mixture of rare earth elements or the like, I probably find myself without a valid potential?!

Yeah that's true. In that. case `job.list_potentials()` will not show any potential. Do you think we need a warning there as well?

samwaseda:
This can probably be merged, right?

jan-janssen (approved):
Looks good to me

PR #1222, opened 2023-11-27
Spx log parser

samwaseda:
This is the first one in the series of changes to be made in order for the Sphinx parsers to become independent.

### DIFF
--- a/pyiron_atomistics/sphinx/output_parser.py
+++ b/pyiron_atomistics/sphinx/output_parser.py
@@ -211,9 +211,5 @@ class SphinxLogParser:
         return convergence
 
     def get_fermi(self):
-        return np.array(
-            [
-                float(line.split()[2])
-                for line in re.findall("Fermi energy:.*$", self.log_main, re.MULTILINE)
-            ]
-        )
+        pattern = r"Fermi energy:\s+(\d+\.\d+)\s+eV"
+        return np.array(re.findall(pattern, self.log_main)).astype(float)
--- a/pyiron_atomistics/sphinx/output_parser.py
+++ b/pyiron_atomistics/sphinx/output_parser.py
@@ -141,7 +141,11 @@ class SphinxLogParser:
             )
         return self._n_atoms
 
-    def get_forces(self, spx_to_pyi=None):
+    def get_forces(self, index_permutation=None):
+        """
+        Args:
+            index_permutation (numpy.ndarray): Indices for the permutation
+        """
         str_fl = "([-+]?[0-9]*\.?[0-9]+(?:[eE][-+]?[0-9]+)?)"
         pattern = r"Atom: (\d+)\t{" + ",".join(3 * [str_fl]) + r"\}"
         arr = np.array(re.findall(pattern, self.log_file))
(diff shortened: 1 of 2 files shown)


pyiron_atomistics/sphinx/output_parser.py
```diff
@@ -0,0 +1,207 @@
+import numpy as np
+import re
+import scipy.constants
+from pyiron_base import state
```
jan-janssen: Can the `logger` be an input of the `SphinxLogParser` class, so that the `output_parser` module no longer depends on `pyiron_base` or `pyiron_atomistics`?

samwaseda: That's an excellent point. Since it's not particularly important I can also think of using `warnings`. What do you prefer?

jan-janssen: I would prefer `warnings` that seems to be simpler from my perspective.

samwaseda:
@skatnagallu This is the first parser that I would like to hand over to you. Since the class does not return a dictionary, as @jan-janssen and I had agreed in a recent workshop, but instead returns a full class, we might want to make changes in the future, but otherwise I would in principle leave the parser to you

jan-janssen:
> @skatnagallu This is the first parser that I would like to hand over to you. Since the class does not return a dictionary, as @jan-janssen and I had agreed in a recent workshop, but instead returns a full class, we might want to make changes in the future, but otherwise I would in principle leave the parser to you

@samwaseda I would prefer to do the migration to the dictionary now, otherwise changes in the class might directly break the code in `pyiron_atomistics`. Having a well defined dictionary as interface seems to be more clean to me.

samwaseda:
> @samwaseda I would prefer to do the migration to the dictionary now, otherwise changes in the class might directly break the code in `pyiron_atomistics`. Having a well defined dictionary as interface seems to be more clean to me.

Definitely! But I would still separate the two PRs

jan-janssen:
> > @samwaseda I would prefer to do the migration to the dictionary now, otherwise changes in the class might directly break the code in `pyiron_atomistics`. Having a well defined dictionary as interface seems to be more clean to me.
> 
> Definitely! But I would still separate the two PRs

Ok, that is fine with me as well.

samwaseda:
Maybe I'm gonna rephrase my comment: I feel a strong responsibility to separate the Sphinx parsers from pyiron, because I don't think anyone can / wants to clean up the mess there. But as soon as it's cleanly separated, I don't feel a strong responsibility to make the change from class properties & getter functions to a dictionary, because we are on an equal footing after the separation. This being said, if you guys have the feeling that I should do it, I can definitely take care of it, since it should be done relatively soon in the future.

skatnagallu:
> @skatnagallu This is the first parser that I would like to hand over to you. Since the class does not return a dictionary, as @jan-janssen and I had agreed in a recent workshop, but instead returns a full class, we might want to make changes in the future, but otherwise I would in principle leave the parser to you

Sounds good. I am not sure if I understand what the dictionary should be? So the entire parsed file should be output as a dictionary with corresponding properties?  I can do this in a separate PR.

samwaseda:
> Sounds good. I am not sure if I understand what the dictionary should be? So the entire parsed file should be output as a dictionary with corresponding properties? I can do this in a separate PR.

Yep that’s a very good point. I have the feeling that in the case of `SphinxLogParser` there are different possibilities, so I guess it’s worth talking about it with everyone who uses SPHInX. For a general idea, I will suggest it in the upcoming PR’s, after this one is merged.

samwaseda:
Can I merge this one?

skatnagallu:
yes, I think so.

samwaseda:
> @samwaseda I would prefer to do the migration to the dictionary now, otherwise changes in the class might directly break the code in `pyiron_atomistics`. Having a well defined dictionary as interface seems to be more clean to me.

I ended up doing it [here](https://github.com/pyiron/pyiron_atomistics/pull/1229)

PR #1223, opened 2023-11-28
SPHInX: Separate other output parsers

samwaseda:


samwaseda:
For my part I think I separated all the output parsers now. I realized that there are a few functions that @sudarsan-surendralal implemented, and I'm not really sure who use them and how well they are tested.

samwaseda:
so it seems to be ok. Can I merge this PR?

PR #1224, opened 2023-11-29
Parser for waves.sxb files from SPHInX

skatnagallu:
Adding a parser for wavefunction files generated by SPHInX


pyiron_atomistics/sphinx/output_parser.py
```diff
@@ -386,3 +386,89 @@ def get_convergence(self):
...
+        filename: file name
+        """
+        self.wfile = h5py.File(filename)
+        self._eps = None
+        self._read()
+
+    # Internal: check that wfile is set
+    def _check_loaded(self):
+        if not isinstance(self.wfile, h5py.File):
+            raise "No waves file loaded"
```
samwaseda:
```suggestion
            raise FileNotFoundError("No waves file loaded")
```

pyiron_atomistics/sphinx/output_parser.py
```diff
@@ -386,3 +386,89 @@ def get_convergence(self):
...
+
+    def _read(self):
+        self._check_loaded()
+        # load various dimensions
+        self.mesh = self.wfile["meshDim"][:]
+        self.Nx, self.Ny, self.Nz = self.mesh
```
samwaseda:
```suggestion
        self.Nx, self.Ny, self.Nz = self.wfile["meshDim"][:]
```

pyiron_atomistics/sphinx/output_parser.py
```diff
@@ -386,3 +386,89 @@ def get_convergence(self):
...
+
+        self.n_states = self.wfile["nPerK"][0]
+        self.n_spin = self.wfile["nSpin"].shape[0]
+        self.k_weights = self.wfile["kWeights"][:]
```
samwaseda: I think it's better to make properties, i.e.:

```python
@property
def n_states(self):
    return self.wfile["nPerK"][0]
```

pyiron_atomistics/sphinx/output_parser.py
```diff
@@ -386,3 +386,89 @@ def get_convergence(self):
...
+        # load the fft_idx to map from condensed psi to FFT mesh
+        # (different mapping per k)
+        self._fft_idx = []
+        self._n_gk = self.wfile["nGk"][:]
+        off = 0
+        for ngk in self._n_gk:
+            self._fft_idx.append(self.wfile["fftIdx"][off : off + ngk])
+            off += ngk
+
+        self.k_vec = self.wfile["kVec"][:]
```
samwaseda:
```suggestion
    @property
    def k_vec(self):
        return self.wfile["kVec"][:]
```

Essentially the same as above

samwaseda (commented):
Is there a particular reason that the class should not require a file when created? I think it's easier to force the user to set it and remove `_check_loaded`

pyiron_atomistics/sphinx/output_parser.py
skatnagallu: There is another [repo](https://github.com/eisenforschung/EXTRA_FIM/blob/main/EXTRA_FIM/main.py), based on sphinx wavefunctions, which uses some of these properties. Maybe in future when we make changes to that, I can change it.

samwaseda: Ah you mean `self.mesh` is also used elsewhere? Then I think it should also be a property

skatnagallu: yep ok.

skatnagallu:
> Is there a particular reason that the class should not require a file when created? I think it's easier to force the user to set it and remove `_check_loaded`

I made changes according to your comment. Have a look.

pyiron_atomistics/sphinx/output_parser.py
```diff
@@ -391,17 +391,25 @@ def get_fermi(self):
...
-            self.load (fname)
-    
+    def __init__(self, file_name="waves.sxb", cwd="."):
+        """
+        Args:
+            file_name (str): file name
+            cwd (str): directory path
+        """
+        if cwd is None:
+            cwd = "."
```
samwaseda: `if cwd is None` etc. is included in the other parsers for the backward compatibility. Here in this class I don’t think it’s required, because it’s new

skatnagallu: coolio

pyiron_atomistics/sphinx/output_parser.py
```diff
@@ -166,7 +169,7 @@ def collect_relaxed_hist(file_name="relaxHist.sx", cwd=None, index_permutation=N
     check_permutation(index_permutation)
     if cwd is None:
         cwd = "."
-    with open(file_name, "r") as f:
+    with open(file_name, "r") as f:eng
```
samwaseda: What is this `eng`?

pyiron_atomistics/sphinx/output_parser.py
```diff
@@ -386,3 +389,103 @@ def get_convergence(self):
...
+        """ Load a waves.sxb file (HDF5 format)
+        
+            filename: file name
+        """
+        self.wfile = h5py.File (filename)
+        self._eps = None
+        self._read ()
+
+    def _read(self):
+        self._check_loaded ()
```
samwaseda: This does not exist anymore right?

samwaseda: Actually I see that there’s still `_check_loaded` everywhere

pyiron_atomistics/sphinx/output_parser.py
```diff
@@ -386,3 +389,103 @@ def get_convergence(self):
...
+        # load various dimensions
+        self.Nx, self.Ny, self.Nz = self.wfile['meshDim'][:]
```
samwaseda: It depends on how often these functions are called, but I might prefer these ones to be properties as well, i.e.:

```python
@property
def Nx(self):
    return self.wfile[“meshDim”][0]
```

The reason is simple: it prevents overwriting. It might appear unlikely, but it sometimes happens, and when it happens it’s usually super difficult to find it. So if we can avoid it, then it’s always better to avoid it.

pyiron_atomistics/sphinx/output_parser.py
```diff
@@ -386,3 +389,103 @@ def get_convergence(self):
...
+        # load the fft_idx to map from condensed psi to FFT mesh
+        # (different mapping per k)
+        self._fft_idx=[]
+        self._n_gk = self.wfile['nGk'][:]
```
samwaseda: Same here, it should be a property

pyiron_atomistics/sphinx/output_parser.py
```diff
@@ -404,8 +404,6 @@ def __init__(self, file_name="waves.sxb", cwd="."):
             file_name (str): file name
             cwd (str): directory path
         """
-        if cwd is None:
-            cwd = "."
         path = Path(cwd) / Path(file_name)
```
freyso: This creates a problem if the filename is an absolute path.
I suggest something like

```
__init__(... cwd=None)
path = (cwd is None) ? file_name : Path(cwd) / Path(file_name)

```
if you feel you need the cwd (I see the convenience). Wouldn't call it cwd though, because typical use cases will not be "current working directory" (which would be superfluous to add), but job directory.

pmrv: Agree, even if `cwd` is given the code should check whether `file_name` is absolute.  See `Path.is_absolute()`.

pyiron_atomistics/sphinx/output_parser.py
skatnagallu: accident. Deleted it.

pyiron_atomistics/sphinx/output_parser.py
skatnagallu: yes. removed it.

pyiron_atomistics/sphinx/output_parser.py
skatnagallu: Agree. So now I think the code checks, if the `file_name` is an absolute path and ignores `cwd` if `True`.

pyiron_atomistics/sphinx/output_parser.py
```diff
@@ -386,3 +389,116 @@ def get_convergence(self):
...
+        return self._eps.T #change
+
+    # Define as separate method and speed it up with numba
+    @staticmethod
+    @numba.jit
+    def _fillin(res,psire,psiim,fft_idx):
+        """Distributes condensed psi (real, imag) on full FFT mesh"""
+        rflat=res.flat
+        for ig in range(fft_idx.shape[0]):
+            rflat[fft_idx[ig]] = complex(psire[ig], psiim[ig])
```
pmrv: Is this really necessary to pull in `numba` here?  Afaik we don't use it elsewhere.

Could you try something like this

```python
res.flat[fft_idx] = psire +1j * psiim
```

pmrv: If this doesn't work directly numpy.put and friends may be helpful.

pyiron_atomistics/sphinx/output_parser.py
```diff
@@ -2,6 +2,9 @@
 import re
 import scipy.constants
 from pathlib import Path
+import numba
+import netCDF4
```
pmrv:
```suggestion
```
Not used.

pyiron_atomistics/sphinx/output_parser.py
skatnagallu: This was @freyso 's call.

pyiron_atomistics/sphinx/output_parser.py
```diff
@@ -386,3 +389,116 @@ def get_convergence(self):
...
+        
+        
+
+    def load(self, filename):
+        """ Load a waves.sxb file (HDF5 format)
+        
+            filename: file name
+        """
+        self.wfile = h5py.File (filename)
+        self._eps = None
```
samwaseda: In principle all variables should be defined in `__init__` (e.g. cf. [this page](https://stackoverflow.com/questions/20661448/should-all-member-variables-be-initialized-in-init)). so `self._eps = None` should also be there. Then maybe we don't need `load` since it has only one line if it's anyway used only once.

pyiron_atomistics/sphinx/output_parser.py
skatnagallu: In principle `res.flat[fft_idx] = psire +1j * psiim` might work.

pyiron_atomistics/sphinx/output_parser.py
samwaseda: That's a good point. Since `cwd` is always set in pyiron it won't be an urgent problem but I have to change this elsewhere as well.

pyiron_atomistics/sphinx/output_parser.py
freyso: yes, this works at the expense of creating two extra copies of the data (namely first multiplying with 1j, which returns a complex vector where all real values are zero, and then adding psire (yielding a second vector), which is then fed into the FFT mesh. This might be fine here, but since I come from HPC, it never occurred to me to go this way, because the loop form doesn't need these superfluous vectors.
If we agree that performance doesn't matter, change it to the proposed version for the sake of readability.

samwaseda: for loops tend to be orders of magnitude slower than `numpy` operations, so probably it's not only readable, but also faster.

freyso: that's what the numba.jit is for. python loops are horribly slow, compiled loops are blazingly fast.

samwaseda: ah ok I see, then I might even prefer to remove both numba and the for loop XD but you are right that we should probably first check the performance

pmrv: Agree with the performance concerns, but I'd really want to avoid pulling in extra dependencies for one method, especially something that couples tightly to numpy.  Making an extra copy will still be much faster than a python loop though.

Anyway, I found this snippet that supposedly makes a complex array without the extra copy [here](https://stackoverflow.com/questions/61356763/numpy-efficient-way-to-create-a-complex-array-from-two-real-arrays).

```np.stack((a, b), axis=-1).view(np.complex).squeeze(-1)```

I have not checked this.

Comment (freyso) [reply pyiron_atomistics/sphinx/output_parser.py]: no need to check performance, the numpy version runs slower by at most a factor 3 in most situations, and that is acceptable because whatever you do next with the psi will be slower anyway. Probably the actual speed loss is less, as the indexed assignment is the actual bottleneck. As I said, it didn't occur to me to do it the numpy way, but I entirely agree that the numpy version is more readable, so kick out numba & the loop. pyiron is not about HPC.

Comment (freyso) [reply pyiron_atomistics/sphinx/output_parser.py]: as for @pmrv 's version: while avoiding one of the two superfluous copies, it is far less readable.

File: pyiron_atomistics/sphinx/output_parser.py [hunk 14]
```diff
@@ -386,3 +388,108 @@ def get_convergence(self):
...
+    # Define as separate method and speed it up with numba
+    @staticmethod
+    @numba.jit
+    def _fillin(res,psire,psiim,fft_idx):
+        """Distributes condensed psi (real, imag) on full FFT mesh"""
+        rflat=res.flat
+        for ig in range(fft_idx.shape[0]):
+            rflat[fft_idx[ig]] = complex(psire[ig], psiim[ig])
+
+    def get_psi_rec(self,i, ispin, ik):
```
samwaseda: I'm generally very uncomfortable with the way this function is written. Is there maybe a possibility to slice `wfile` at the beginning to avoid the offsets? @freyso

pyiron_atomistics/sphinx/output_parser.py
freyso: and then merge the reorder code into the read function, because the extra reorder function was only needed to encapsulate the loop for numba.jit

pyiron_atomistics/sphinx/output_parser.py
freyso: I don't think so. Data reading is meant to be done on demand, as wave function files can become dozens of GB large. The use cases I know of (so far: Shalini's FIM-EXTRA) process a single psi at a time.

samwaseda: Which part is done on demand? `self.wfile[f"psi-{ik+1}.re"]`?

freyso: as  far as I understand the hdf5 python interface, wfile["varname"] does not read anything, it returns an abstract handle for the data inside the file. Only when I index the handle, the data is read into a numpy array.

pyiron_atomistics/sphinx/output_parser.py
skatnagallu: Just did a quick test the `numba `JIT version takes 2.5 s and `numpy `version takes 0.1 s. So I will remove the `numba `implementation

skatnagallu: also, I stuck with this `res.flat[fft_idx] = psire +1j * psiim` as I felt this was readable to me.

pyiron_atomistics/sphinx/output_parser.py
```diff
@@ -386,3 +387,97 @@ def get_convergence(self):
...
+        return self.wfile["nGk"][:]
+
+    @property
+    def _fft_idx(self):
+        fft_idx = []
+        off = 0
+        for ngk in self._n_gk:
+            fft_idx.append(self.wfile["fftIdx"][off : off + ngk])
+            off += ngk
+        return fft_idx
```
samwaseda:
```suggestion
    def _get_fft_idx(self, ik):
        i_start = 0
        if ik > 0:
            i_start = np.sum(self._n_gk[ik - 1])
        i_end = np.sum(self._n_gk[ik])
        return self.wfile["fftIdx"][i_start : i_end])
```

or something like this because only one element of the current list (i.e. `self._fft_idx[ik]` in line 477) is used and appending all the elements is too inefficient

pyiron_atomistics/sphinx/output_parser.py
```diff
@@ -386,3 +387,97 @@ def get_convergence(self):
...
+        if ispin < 0 or ispin >= self.n_spin:
+            raise IndexError(f"ispin={ispin} fails 0 <= ispin < n_spin={self.n_spin}")
+        if ik < 0 or ik >= self.nk:
+            raise IndexError(f"ik={ik} fails 0 <= ik < nk={self.nk}")
+
+        res = np.zeros(shape=self.mesh, dtype=np.complex128)
+        off = self._n_gk[ik] * (i + ispin * self.n_states)
+        psire = self.wfile[f"psi-{ik+1}.re"][off : off + self._n_gk[ik]]
+        psiim = self.wfile[f"psi-{ik+1}.im"][off : off + self._n_gk[ik]]
+        res.flat[self._fft_idx[ik]] = psire + 1j * psiim
```
samwaseda:
```suggestion
        res.flat[self._get_fft_idx(ik)] = psire + 1j * psiim
```

pyiron_atomistics/sphinx/output_parser.py
```diff
@@ -386,3 +387,97 @@ def get_convergence(self):
...
+    def get_psi_rec(self, i, ispin, ik):
+        """Loads a single wavefunction on full FFT mesh"""
+        if i < 0 or i >= self.n_states:
+            raise IndexError(f"i={i} fails 0 <= i < n_states={self.n_states}")
+        if ispin < 0 or ispin >= self.n_spin:
+            raise IndexError(f"ispin={ispin} fails 0 <= ispin < n_spin={self.n_spin}")
+        if ik < 0 or ik >= self.nk:
+            raise IndexError(f"ik={ik} fails 0 <= ik < nk={self.nk}")
+
+        res = np.zeros(shape=self.mesh, dtype=np.complex128)
```
samwaseda, line 473:
```suggestion
        res = np.zeros(shape=self.mesh, dtype=np.complex128).flat
```

pyiron_atomistics/sphinx/output_parser.py
```diff
@@ -386,3 +387,97 @@ def get_convergence(self):
...
+        off = self._n_gk[ik] * (i + ispin * self.n_states)
+        psire = self.wfile[f"psi-{ik+1}.re"][off : off + self._n_gk[ik]]
+        psiim = self.wfile[f"psi-{ik+1}.im"][off : off + self._n_gk[ik]]
```
samwaseda, line 512: The indices are not the same as the ones used in `_fft_idx`? Can't we simply use `self.wfile["fftIdx"][off : off + ngk]` instead of `self._fft_idx[ik]`? @freyso

samwaseda:
Sorry I made super confusing comments now. What I want is:

1. If `self.wfile["fftIdx"][off : off + self._n_gk[ik]]` is the same as `self._fft_idx[ik]` in `def get_psi_rec`, then I think `self.wfile["fftIdx"][off : off + self._n_gk[ik]]` should be used
2. If not, then `def _fft_idx` should be a function that takes the index `ik`, so that we don't need to make a full array

pyiron_atomistics/sphinx/output_parser.py
skatnagallu: I think they are.

freyso: Sorry for the late response - they are not the same.
There is 1 fft_idx per k-point, stored contiguously in the file. The original version loaded this into a python list of numpy vectors. Very efficient appending, to answer the concern above. This was loaded once from file and then reused for all psi (which are many per k-point, and `off` refers to the index within the k-specific psi-variable inside the file).
The current version is wrong due to index problems of the fft_idx.

pyiron_atomistics/sphinx/output_parser.py
```diff
@@ -417,3 +417,89 @@ def results(self):
...
+        res.flat[self.wfile["fftIdx"][off : off + self._n_gk[ik]]] = psire + 1j * psiim
```
freyso: this won't work, as fftIdx for a specific k inside the global fft_idx var in the hd5 file has different offsets than psi for a specific ispin, istate inside the k-specific psi-variable inside hd5 file.
I also would guess that keeping fft_idx in memory instead of reloading for each new psi is better. If not keeping all fftIdx at all time, at least caching would be desirable (there is a caching function decorator).

pyiron_atomistics/sphinx/output_parser.py
samwaseda: ok if the data is stored it might be efficient, but that's not what's happening in the current implementation, so we have to change `_fft_idx` in the way that it's stored in a variable.

pyiron_atomistics/sphinx/output_parser.py
skatnagallu: I have reverted some changes. please have a look.

pyiron_atomistics/sphinx/output_parser.py
```diff
@@ -401,8 +403,7 @@ def get_convergence(self):
 
     def get_fermi(self):
         pattern = r"Fermi energy:\s+(\d+\.\d+)\s+eV"
-        return np.array(re.findall(pattern, self.log_main)).astype(float)
-
+
```
freyso: Is this deletion on purpose?

pyiron_atomistics/sphinx/output_parser.py
```diff
@@ -417,3 +418,98 @@ def results(self):
...
+            self.wfile = h5py.File(Path(file_name))
+        else:
+            path = Path(cwd) / Path(file_name)
+            self.wfile = h5py.File(path)
+
+    @property
+    def _n_gk(self):
+        return self.wfile['nGk'][:]
+    
+    @property
```
freyso:
```suggestion
    @cached_property
```

pyiron_atomistics/sphinx/output_parser.py
```diff
@@ -417,3 +418,98 @@ def results(self):
...
+    def mesh(self):
+        return self.wfile["meshDim"][:]
+
+    @property
+    def Nx(self):
+        return self.wfile['meshDim'][0]
```
freyso: Get via mesh (notably  if cached, as suggested above)
```suggestion
        return mesh[0]
```

pyiron_atomistics/sphinx/output_parser.py
```diff
@@ -417,3 +418,98 @@ def results(self):
...
+
+    @property
+    def Ny(self):
+        return self.wfile['meshDim'][1]
```
freyso: Get via mesh (notably  if cached, as suggested above)
```suggestion
        return mesh[1]
```

pyiron_atomistics/sphinx/output_parser.py
```diff
@@ -417,3 +418,98 @@ def results(self):
...
+
+    @property
+    def Nz(self):
+        return self.wfile['meshDim'][2]
```
freyso: Get via mesh (notably  if cached, as suggested above)
```suggestion
        return mesh[2]
```

pyiron_atomistics/sphinx/output_parser.py
```diff
@@ -417,3 +418,98 @@ def results(self):
...
+
+    @property
```
freyso: cached property
```suggestion
    @cached_property
```

pyiron_atomistics/sphinx/output_parser.py
```diff
@@ -417,3 +418,98 @@ def results(self):
...
+    def n_states(self):
+        return self.wfile["nPerK"][0]
+
+    @property
```
freyso: cached property
```suggestion
    @cached_property
```

pyiron_atomistics/sphinx/output_parser.py
```diff
@@ -417,3 +418,98 @@ def results(self):
...
+    def n_spin(self):
+        return self.wfile["nSpin"].shape[0]
+
+    @property
```
freyso: cached property
```suggestion
    @cached_property
```

pyiron_atomistics/sphinx/output_parser.py
```diff
@@ -417,3 +418,98 @@ def results(self):
...
+    def k_weights(self):
+        return self.wfile["kWeights"][:]
+
+    @property
```
freyso: cached property
```suggestion
    @cached_property
```

pyiron_atomistics/sphinx/output_parser.py
```diff
@@ -417,3 +418,98 @@ def results(self):
...
+
+    """
+
+    def __init__(self, file_name="waves.sxb", cwd="."):
+        """
+        Args:
+            file_name (str): file name
+            cwd (str): directory path
+        """
+        self._eps = None
```
freyso: _eps can be removed (see below)

pyiron_atomistics/sphinx/output_parser.py
```diff
@@ -417,3 +418,98 @@ def results(self):
...
+    @property
+    def k_vec(self):
+        return self.wfile["kVec"][:]
+
+    @property
+    def eps(self):
+        """All eigenvalues (in Hartree) as (nk,n_states) block"""
+        if (self._eps is None):
+            self._eps = self.wfile['eps'][:].reshape(-1,self.n_spin,self.n_states)
+        return self._eps.T #change
```
freyso:
```suggestion
    @cached_property
    def eps(self):
        """All eigenvalues (in Hartree) as (nk,n_states) block"""
        return (self.wfile['eps'][:].reshape(-1,self.n_spin,self.n_states)).T
```
The _eps stuff was a manual implementation of what @cached_property does.

pyiron_atomistics/sphinx/output_parser.py
```diff
@@ -417,3 +418,98 @@ def results(self):
             if len(value) > 0:
                 results["dft"][key] = value
         return results
+
+
+class SphinxWavesParser:
```
freyso: is "parser" the best name? It actually is a "reader"...

For me, "parsing" sounds as if there is some human-readable text that needs to be processed to extract information. But most of this is plain reading from hdf5, plus a bit of reshaping/reordering.

pyiron_atomistics/sphinx/output_parser.py
```diff
@@ -402,7 +404,7 @@ def get_convergence(self):
     def get_fermi(self):
         pattern = r"Fermi energy:\s+(\d+\.\d+)\s+eV"
         return np.array(re.findall(pattern, self.log_main)).astype(float)
-
+
```
samwaseda:
```suggestion

```

pyiron_atomistics/sphinx/output_parser.py
```diff
@@ -417,3 +419,95 @@ def results(self):
...
+        off = 0
+        for ngk in self._n_gk:
+            fft_idx.append(self.wfile["fftIdx"][off : off + ngk])
+            off += ngk
+        return fft_idx
+
+    def get_psi_rec(self, i, ispin, ik):
+        """Loads a single wavefunction on full FFT mesh"""
+        if i < 0 or i >= self.n_states:
+            raise IndexError(f"i={i} fails 0 <= i < n_states={self.n_states}")
```
samwaseda:
```suggestion
        i = np.arange(self.n_states)[i]
```

Failing `i < 0` is actually a bit against the idea of python, as for example -1 means `self.n_states - 1` in this case. The operation I suggest above will already include the message when `i < n_states`. The same applies in the if statements below.

pyiron_atomistics/sphinx/output_parser.py
```diff
@@ -417,3 +419,95 @@ def results(self):
...
+        if ispin < 0 or ispin >= self.n_spin:
+            raise IndexError(f"ispin={ispin} fails 0 <= ispin < n_spin={self.n_spin}")
```
samwaseda:
```suggestion
        ispin = np.arange(self.n_spin)[ispin]
```

pyiron_atomistics/sphinx/output_parser.py
```diff
@@ -417,3 +419,95 @@ def results(self):
...
+        if ik < 0 or ik >= self.nk:
+            raise IndexError(f"ik={ik} fails 0 <= ik < nk={self.nk}")
```
samwaseda:
```suggestion
        ik = np.arange(self.nk)[ik]
```

pyiron_atomistics/sphinx/output_parser.py
```diff
@@ -417,3 +419,95 @@ def results(self):
...
+    def _fft_idx(self):
+        fft_idx = []
+        off = 0
+        for ngk in self._n_gk:
+            fft_idx.append(self.wfile["fftIdx"][off : off + ngk])
+            off += ngk
+        return fft_idx
+
+    def get_psi_rec(self, i, ispin, ik):
+        """Loads a single wavefunction on full FFT mesh"""
```
samwaseda: This doc string is not quite enough. It should at least state arguments and return value. In addition to this, I would like it to contain the information about how to get the shape. I presume in this case it's `self.mesh.shape`?

pyiron_atomistics/sphinx/output_parser.py
skatnagallu: I don't understand. If `i = -1 ` then the code should throw an error as states are indexed from 0 to `self.n_states`. And I would think if the user uses a state index higher than `n_states` or less than zero, this IndexError makes sense.

samwaseda: How is it different from `my_list[index]`? Here, `index` can be also a negative number

skatnagallu: by index I mean how a state is labeled. this function basically gives a wavefunction corresponding to a state labelled by (state,spin,k).

samwaseda: yeah, and -1 means the last state

skatnagallu: I understand this from python. But coming from SPHInX, I felt all the states are always labelled by positive numbers and negative number should through and error rather than giving last states.

skatnagallu: I do see your point. But I will wait to see @freyso opinion on this.

samwaseda: Sure, because SPHInX is written in C++, which doesn't allow negative indices, but it shouldn't have a fundamental problem with allowing it. Not allowing negative indices however breaks the python convention, and I don't think it's a good idea to do so only because we don't know what it could be useful for. If there's a fundamental problem, I'm also ok to forbid it.

freyso: I agree to the suggested workaround to ensure that negative indices are handled in a pythonic way. Just a comment in the code would be good, something like "translate indices from pythonic style to 0 ... n-1". Otherwise, the lines look like a complicated idempotency.

skatnagallu:
can this be merged?

pyiron_atomistics/sphinx/output_parser.py
```diff
@@ -417,3 +419,106 @@ def results(self):
...
+        ispin = np.arange(self.n_spin)[ispin]
+
+        res = np.zeros(shape=self.mesh, dtype=np.complex128)
+        off = self._n_gk[ik] * (i + ispin * self.n_states)
+        psire = self.wfile[f"psi-{ik+1}.re"][off : off + self._n_gk[ik]]
+        psiim = self.wfile[f"psi-{ik+1}.im"][off : off + self._n_gk[ik]]
+        res.flat[self._fft_idx[ik]] = psire + 1j * psiim
+        if compact:
+            compact_wave = psire + 1j * psiim
+            return compact_wave
```
freyso: I suggest to reorder code as follows.
```suggestion
        off = self._n_gk[ik] * (i + ispin * self.n_states)
        psire = self.wfile[f"psi-{ik+1}.re"][off : off + self._n_gk[ik]]
        psiim = self.wfile[f"psi-{ik+1}.im"][off : off + self._n_gk[ik]]
        compact_wave = psire + 1j * psiim
        if compact:
            return compact_wave
        res = np.zeros(shape=self.mesh, dtype=np.complex128)
        res.flat[self._fft_idx[ik]] = compact_wave
```

skatnagallu:
I want to add tests for this.

skatnagallu:
Added a test for SphinxWavesReader class. I created a 'fake.sxb' which is a fake wavefunction file. As the real wavefunction files even for small molecule and a single k point was already around 50 mb.

niklassiemer:
Couldn't you get a 'real' wavefunction for a single k point and (well I am not sure how sphinx does it - pure plane waves?) a very low energy cutoff? However, if the fake file has the same structure it should also be fine.

freyso:
an Al bulk cell (primitive fcc, 1 atom) with a single k should do.

skatnagallu:
I have added now a proper waves.sxb file and the tests work. I would suggest that we merge this into the main for now. I will be going on vacation now, and the wavefunction reader would be useful to others.

skatnagallu:
Can I merge this?

jan-janssen:
```
FAIL: test_waves (sphinx.test_parsers.TestSphinx)
  File "/home/runner/work/pyiron_atomistics/pyiron_atomistics/tests/sphinx/test_parsers.py", line 47, in test_waves
    self.assertTrue(np.mean(waves.get_psi_rec(0,0,0))>0)
AssertionError: False is not true

```

tests/sphinx/test_parsers.py
```diff
@@ -44,7 +44,7 @@ def test_waves(self):
         waves_path = Path.joinpath(test_path,'static/sphinx/sphinx_test_waves')
         waves = SphinxWavesReader('waves.sxb',cwd=waves_path)
         self.assertTrue(waves.nk==1)
-        self.assertTrue(np.mean(waves.get_psi_rec(0,0,0))>0)
+        self.assertTrue(np.mean(np.abs(waves.get_psi_rec(0,0,0)))>0)
```
niklassiemer: That test could only fail if all values in the waves are exactly zero?

skatnagallu: It is unlikely for the waves file, it is for a bulk Al supercell, which I created.

niklassiemer: And for the Al case, it is ok if the value of `waves.get_psi_rec(0,0,0)` are negative? I just think the test now is really weak and I am not sure if this is intended. I mean a single non-zero value in the full `waves.get_psi_rec(0,0,0)` would immediately cause a positive mean overall.

skatnagallu: But my understanding was to make sure the reading routine was correctly reading the waves file.

niklassiemer: Yes, the test would also fail if there is an error, however, the shape or some more specific value would test the _correct_ reading better than 'have one not zero value in that array'

tests/sphinx/test_parsers.py
```diff
@@ -0,0 +1,51 @@
...
+            cwd=self.get_path("sphinx_test_2_3")
+        )
+        self.assertTrue("scf_energy_free" not in E_dict)
+
+    def test_waves(self):
+        test_path = Path(__file__).parents[2-1]
+        waves_path = Path.joinpath(test_path,'static/sphinx/sphinx_test_waves')
+        waves = SphinxWavesReader('waves.sxb',cwd=waves_path)
+        self.assertTrue(waves.nk==1)
+        self.assertTrue(np.mean(np.abs(waves.get_psi_rec(0,0,0)))>0)
```
samwaseda:
```suggestion
        self.assertAlmostEqual(np.linalg.norm(waves.get_psi_rec(0,0,0)), 0.9984249664645706)
        self.assertEqual(waves.get_psi_rec(0,0,0).shape, (10, 10, 10))
```

To address @niklassiemer 's concern, which I share. I find it still ugly, but it fulfils the purpose.

samwaseda: Or it could also be `self.assertEqual(waves.get_psi_rec(0,0,0).shape, (10, 10, 10))` as @niklassiemer suggested.

niklassiemer: I would put both tests ;)

tests/sphinx/test_parsers.py
samwaseda: I violently committed my suggestion 😎

samwaseda (approved):

PR #1227, opened 2023-11-29
Make generic energy_pot of Vasp configurable

pmrv:
Vasp outputs three energies, smeared internal energy, free energy and smeared internal extrapolated to zero smearing.
This change allows to pick which of them is set as

output/generic/energy_pot

of the job.

Current defaults are smeared internal extrapolated to zero smearing when vasprun.xml can be read and free energy when OUTCAR is read.

### DIFF
--- a/pyiron_atomistics/vasp/base.py
+++ b/pyiron_atomistics/vasp/base.py
@@ -1927,6 +1927,14 @@ class Input:
         else:
             return eddrmm_value
 
+import enum
+class DftEnergyKind(enum.Enum):
+    ENERGY_FREE = enum.auto()
+    """The full free energy of the electronic system, i.e. smeared internal energy plus entropy."""
+    ENERGY_INTERNAL = enum.auto()
+    """The smeared internal energy of the electronic system."""
+    ENERGY_ZERO = enum.auto()
+    """The internal energy of the electronic system, extrapolated to zero smearing."""
 
 class Output:
     """
@@ -1938,8 +1946,9 @@ class Output:
         charge_density: Gives the charge density of the system
     """
 
-    def __init__(self):
+    def __init__(self, energy_kind=None):
         self._structure = None
+        self._energy_kind = energy_kind
         self.outcar = Outcar()
         self.oszicar = Oszicar()
         self.generic_output = GenericOutput()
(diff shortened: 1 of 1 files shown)


pmrv:
This also sidesteps [a bug](https://www.vasp.at/forum/viewtopic.php?p=19264&hilit=vasprun+energy#p19264) in the 5.4.4 xml output.

Ideally, I'd like to make ENERGY_FREE the default, because otherwise the "generic" forces are not consistent with the "generic" energies and that just doesn't make sense. For now I've kept the previous behavior though.

I'm not married to the name `energy_kind` because it's not very self explanatory, but couldn't think of anything right now.

samwaseda:
Hmmmmmmmmmmmm, I find it super dangerous to be able to change the default. Actually in a meeting long long time ago we kind of agreed to use the free energy for `energy_pot`. But anyway I have the feeling that this requires a bit more than just a review.

pmrv:
I definitely need a way for `output/generic/energy_pot` to be the free energy.  I'm ok if that's the only way, but it would break behavior so that's how I ended up this way.

samwaseda:
ok then I would rather support the idea of changing it to the free energy. Anything else would be for me nonsense because all the other derivative properties (forces, pressures etc.) come from the free energy.

samwaseda:
Also: if we decide to make it optional, I think we need a better concept to choose output values. That's something @jan-janssen and I talked about yesterday, and from the performance point of view and also in the context of workflow I fully support it, but I would not just want to create a new entry in the input and say this is how it works from now on.

pmrv:
Is there any other place that makes sense to configure it, though?  If output/parsing/etc. behavior can be changed, I would want it on the job level, because otherwise I'll need to keep global state in my head as I work between different projects.  If it is to live on a job, `input` seems the correct location?  One can think about a more "generic" input where options like this live, but a) we don't have one yet and b) these parser options would likely by very code specific.

pmrv:
Either way, the priority for me is to change the output to the free energy.  I can simplify this and we discuss the optional stuff later.

samwaseda:
> Is there any other place that makes sense to configure it, though?

No there isn't. For interactive LAMMPS we have something similar [here](https://github.com/pyiron/pyiron_atomistics/blob/bffc9d9fb3c6680e001f3554fcb30ccb8d5ea6cf/pyiron_atomistics/atomistics/job/interactive.py#L43) though, which is not really to be touched, but as we all know, this kind of stuff is not to be touched, until people start touching it, and in the end we have the same functionality with different standards everywhere.

> If output/parsing/etc. behavior can be changed, I would want it on the job level, because otherwise I'll need to keep global state in my head as I work between different projects. If it is to live on a job, `input` seems the correct location?

No I don't think it's the "correct" location, but it's the only one "not impossible" location, because our input currently does not talk about output (except for the phonopy job, which you and I hate). If we include the output info in the input, then I think we need a clear interface with a well defined structure.

> One can think about a more "generic" input where options like this live, but a) we don't have one yet and b) these parser options would likely by very code specific.

Yes, I agree, and just like the interactive class that I mentioned above, it's not necessarily code specific.

pmrv:
- bump vasp version number in job and parse the correct energy
- add a maintenance function that 
  1. changes the energy_pot to energy_free
  2. check whether there are any master jobs
- close this

PR #1242, opened 2023-12-03
Switch from setup.py to pyproject.toml

jan-janssen:


jan-janssen:
Documentation tests fail with `Could not import extension nbsphinx (exception: No module named 'nbsphinx')`

niklassiemer:
@jan-janssen is there a reason the caching of already built environments was removed here?

jan-janssen:
> @jan-janssen is there a reason the caching of already built environments was removed here?

The pulling of the cache was slower than reinstalling the packages.

PR #1249, opened 2023-12-05
Add AtomisticsMaintenance

pmrv:
Also add a method that corrects energy_pot of Vasp jobs.

This only adds the maintenance methods, the output parser is not yet updated.

### DIFF
--- a/pyiron_atomistics/vasp/base.py
+++ b/pyiron_atomistics/vasp/base.py
@@ -62,7 +62,7 @@ def _vasp_generic_energy_free_affected(job):
         energy_free = np.array(
             [e[-1] for e in job.project_hdf5["output/generic/dft/scf_energy_free"]]
         )
-        energy_pot = self.project_hdf5["output/generic/energy_pot"]
+        energy_pot = job.project_hdf5["output/generic/energy_pot"]
         return not np.allclose(energy_free, energy_pot)
     else:
         return False
--- a/pyiron_atomistics/vasp/base.py
+++ b/pyiron_atomistics/vasp/base.py
@@ -2070,7 +2070,6 @@ class Output:
             log_dict["forces"] = self.vp_new.vasprun_dict["forces"]
             log_dict["cells"] = self.vp_new.vasprun_dict["cells"]
             log_dict["volume"] = np.linalg.det(self.vp_new.vasprun_dict["cells"])
-            # log_dict["total_energies"] = self.vp_new.vasprun_dict["total_energies"]
             log_dict["energy_tot"] = self.vp_new.vasprun_dict["total_energies"]
             if "kinetic_energies" in self.vp_new.vasprun_dict.keys():
                 log_dict["energy_pot"] = (
--- a/pyiron_atomistics/vasp/base.py
+++ b/pyiron_atomistics/vasp/base.py
@@ -2242,8 +2242,12 @@ class Output:
                 self.vp_new.vasprun_dict["parameters"]["electronic"]["NELECT"]
             )
             if "kinetic_energies" in self.vp_new.vasprun_dict.keys():
+                # scf_energy_kin is for backwards compatibility 
                 self.generic_output.dft_log_dict[
                     "scf_energy_kin"
+                ] = self.vp_new.vasprun_dict["kinetic_energies"]
+                self.generic_output.dft_log_dict[
+                    "energy_kin"
                 ] = self.vp_new.vasprun_dict["kinetic_energies"]
 
         if (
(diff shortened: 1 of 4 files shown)


pyiron_atomistics/vasp/base.py
```diff
@@ -53,6 +53,17 @@
...
 __date__ = "Sep 1, 2017"
 
+def _vasp_generic_energy_free_affected(job):
+    """
+    Checks whether the value saved in output/generic/energy_pot matches the electronic free energy.
+    """
+    if self.project_hdf5.get("HDF_VERSION", "0.1.0") == "0.1.0":
+        energy_free = np.array([e[-1] for e in self.project_hdf5["output/generic/dft/scf_energy_pot"]])
+        energy_pot = self.project_hdf5["output/generic/energy_pot"]
+        return not np.allclose(energy_free, energy_pot)
```
samwaseda, line 66: Isn't it other way around? Shouldn't it be `True` when they are all close?

pmrv: Oh, it needs to be scf_energy_free.

samwaseda: ah ok now I understand your intention. I thought it was `energy_zero` XD

pmrv:
@ahmedabdelkawy and I also found two bugs in the way AIMD runs were parsed that are quite severe:

1. only the final kinetic energy was parsed
2. this was then subtracted from the total electronic energy instead of added

Bug fixes and another update method are included in this PR.

@ahmedabdelkawy I've tested this on a quick MD run, but would be great if you could also try it on your jobs.

pmrv:
Test failure seems to be related to importing Vasp runs from file, will have to check tomorrow.

pyiron_atomistics/vasp/base.py
```diff
@@ -2051,15 +2073,33 @@ def collect(self, directory=os.getcwd(), sorted_indices=None):
...
+                log_dict["energy_pot"] = self.vp_new.vasprun_dict["total_energies"]
             if "kinetic_energies" in self.vp_new.vasprun_dict.keys():
-                log_dict["energy_pot"] = (
-                    log_dict["energy_tot"]
-                    - self.vp_new.vasprun_dict["kinetic_energies"]
+                log_dict["energy_tot"] = (
+                    log_dict["energy_pot"]
+                    + self.vp_new.vasprun_dict["kinetic_energies"]
                 )
             else:
```
ahmedabdelkawy, line 2101: This will never be true, because the if condition only checks whether kinetic_energies key exists or not and not if there are actual values in it. This becomes a problem for static calculations where self.vp_new.vasprun_dict["kinetic_energies"] is an empty list but the key is still there. This will result in log_dict["energy_tot"] being empty.

pmrv: Correct!  I pushed a fix.

ahmedabdelkawy: parsing energy now works perfectly!

ahmedabdelkawy:
For misfortunate future generations who will have to work on this again: 

Parsing energies from vasp to pyiron now works as follows:
1. Parsing is done from vasprun.xml and not the outcar
2. Parsing of energy values is done by the following order of files: vasprun.xml -> vasprun.py -> base.py
3. There is a bug in vasprun.xml for vasp 5.xx versions described here: https://ww.vasp.at/forum/viewtopic.php?p=19242 and was taken into consideration
4. Parsed energy terminology is the following:
 * Energy minimization jobs:
  1. job.output.energy_pot  =  free energy (including smearing if used to be consistent with forces)
  2. job.output.energy_tot = job.output.energy_pot

 * MD jobs:
  1. job.output.energy_pot  =  free energy (including smearing if used to be consistent with forces)
  2. job.output.energy_tot = job.output.energy_pot  + job['output/generic/dft/energy_kin'] (kinetic energies)


More accessible quantities that are stored as a dictionary:
- arrays with MxN shape: M = number of ionic steps, N = number of electronic steps (ragged: differs from one ionic step to the other)
1. job['output/generic/dft/scf_energy_free'] -> free energy
2. job['output/generic/dft/scf_energy_int'] -> internal energy, aka energy when the entropy (eg: due to the degeneracy effect of the bands) is equal to zero
3. job['output/generic/dft/scf_energy_kin'] -> kinetic energies
5.  job['output/generic/dft/scf_energy_zero'] -> energy when the electronic temperature (aka sigma) is extrapolated to zero

- same as above but only at the last electronic step for each ionic step:
1. job['output/generic/dft/energy_free'] -> free energy
2. job['output/generic/dft/energy_int'] -> internal energy, aka energy when the entropy (eg: due to the degeneracy effect of the bands) is equal to zero
3. job['output/generic/dft/energy_kin'] -> kinetic energies
4.  job['output/generic/dft/energy_zero'] -> energy when the electronic temperature (aka sigma) is extrapolated to zero

pmrv:
@niklassiemer Please bump the minor version for the next release.

PR #1253, opened 2023-12-08
new sx structure parser that tolerates unexpected items

freyso:
In part, this is an experiment on parsing.

Idea 1 is to parse while reading the file, instead of reading the entire file first.
Idea 2 is to use a more tabular way of programming, by creating a table of keywords with associated parsing routines. 
Whenever the keyword is encountered, the corresponding parsing routine is called.
This is organized in a multi-level way, so certain keywords will simply add additional keywords and their associated parsing routines, which will be removed again from the when the next higher-level keyword appears. This allows to parse larger sections that have multiple subitems. The hope is to produce more readable (and thus maintainable) parsing routines.

I have implemented this for the rarely used sphinx.structure.read_atoms routines, which we need to parse 
some old data I created outside of pyiron. The new version supports multiple structures in the same file and ignores unexpected stuff (such as labels) much more reliably than the old version.

### DIFF
--- a/pyiron_atomistics/sphinx/parser_base.py
+++ b/pyiron_atomistics/sphinx/parser_base.py
@@ -1,0 +1,188 @@
+# coding: utf-8
+# Copyright (c) Max-Planck-Institut für Eisenforschung GmbH - Computational Materials Design (CM) Department
+# Distributed under the terms of "New BSD License", see the LICENSE file.
+
+__author__ = "Christoph Freysoldt"
+__copyright__ = (
+    "Copyright 2023, Max-Planck-Institut für Eisenforschung GmbH - "
+    "Computational Materials Design (CM) Department"
+)
+__version__ = "1.0"
+__maintainer__ = "Christoph Freysoldt"
+__email__ = "freysoldt@mpie.de"
+__status__ = "production"
+__date__ = "Dec 8, 2023"
+
+import re
+from types import GeneratorType
+import numpy
+
+class keyword_tree_parser:
+    """
+    A base class to parse files block by block via keyword-triggered
+    parsing routines organized in a tree. Parsing routines can
+    add more levels of keyword->parse function maps. The file
+    is read line by line on demand while parsing, so large files will not clobber
+    memory.
+
+    A parser routine can either return or yield (once!) to continue parsing. If it yields,
+    the rest of the routine (after yield) will be executed when the next keyword
+    of the current or a higher level is found.
+
+    Every parser routine MUST remove the keyword from the lineview.
+
+    A typical use will be
+
+    class my_parser(keyword_tree_parser):
+        def __init__(self,file)
+            super ().__init__({
+                    "key1" : self.parse_key1,
+                    "key2" : self.parse_key2  })
+            self.parse (file)
+
+    """
+    def __init__ (self,keylevels=[]):
+        if isinstance(keylevels,dict):
+            keylevels = [ keylevels ]
+        elif not isinstance(keylevels,list):
+            raise TypeError
+        self.keylevels = keylevels
+
+    def parse(self,filename):
+        """
+        Parse a file using the current keylevels
+
+        Args:
+            filename ... the filename of the file to parse
+        Returns: nothing
+        """
+        # --- initialization
+        if len(self.keylevels) == 0:
+           raise KeyError("No parsing functions available in keylevels")
+        filehandle = open(filename)
+        self.line = filehandle.__iter__ ()
+        self.lineview = ''
+        self.filename = filename
+        self.lineno = 0
+        while True:
+            for keymap in self.keylevels:
+                for key,func in keymap.items ():
+                    if key in self.lineview:
+                        self._cleanup(keymap)
+                        res = func ()
+                        if isinstance(res,GeneratorType):
+                            res.send (None)
+                            keymap['%finalize!'] = res
+                        break
+                else:
+                    continue
+                break
+            else:
+                try:
+                    self.lineview = next(self.line)
+                    self.lineno += 1
+                    self.line_from = self.lineno
+                except StopIteration:
+                    break
+        self._cleanup(self.keylevels[0])
+        if (hasattr(self,'finalize')):
+            self.finalize ()
+
+    def location(self):
+        """ Return the current parsing location (for error messages)"""
+        return f"in file '{self.filename}' line" + \
+               ( f" {self.lineno}" if self.lineno == self.line_from else
+                 f"s {self.line_from}..{self.lineno}" )
+
+    def read_until(self, match):
+        """
+        Appends more lines from input until match is found
+
+        Args:
+           match ... (str) what to wait for
+        Returns: nothing
+        """
+        while not match in self.lineview:
+            self.lineview += next(self.line)
+            self.lineno += 1
+            self.lineFrom = self.line
+
+    def extract_via_regex(self, regex):
+        """
+        Extracts and removes some text from current lineview
+
+        Args:
+           regex ... regular expression
+        Returns:
+           the extracted text
+        """
+        if isinstance(regex, str):
+            regex = re.compile (regex, re.DOTALL)
+        result = regex.search(self.lineview)
+        if result is None:
+            raise RuntimeError(f"Failed to extract '{regex.pattern}' "
+                               + self.location ()
+                               + "\n" + self.lineview)
+        self.lineview = regex.sub ('',self.lineview, count=1)
+        return result.group ()
+
+    def _cleanup(self, active):
+        """
+        (internal routine) remove levels below the current (active) level, and
+         call (optional) final blocks up to the current level
+
+        Args:
+           active ... the currently active map
+        Returns:
+           the extracted text
+        """
+        def try_finalize(keymap):
+            if '%finalize!' in keymap:
+                try:
+                    next(keymap['%finalize!'])
+                except StopIteration:
+                    pass
+                del keymap['%finalize!']
+        # roll back keylevels until active level
+        while (self.keylevels[-1] is not active):
+            try_finalize (self.keylevels[-1])
+            del self.keylevels[-1]
+        # and call optional finalize of currently active level
+        try_finalize(active)
+
+    def get_vector (self, key, txt):
+        """
+        (auxiliary function) Get a vector from 'key = [ ... ] ;'
+
+        Args:
+           key ... the key to look for
+        Returns:
+           one-dimensional vector containing the numbers
+        """
+        # get the relevant part between '=' and ';'
+        vecstring = re.sub ('.*' + key + r"\s*=\s*([^;]+);.*", r"\1", txt)
+        if vecstring is None:
+            raise RuntimeError(f"Cannot parse {key} from '{txt}' as vector "
+                               + self.location ())
+        # remove special characters [] , ; =
+        vecstring = re.sub(r"[][=,;$]",' ',vecstring)
+        return numpy.fromstring(vecstring, sep=' ')
+
+    def extract_var(self, key, startend='=;'):
+        """
+        Extract a block 'key = ... ;'
+
+        If the end pattern is not found in lineview, more lines are read.
+
+        Args:
+            key      ... the keyword
+            startend ... (optional) Override the = ; pair by two different patterns
+        Returns:
+            the extracted block
+        """
+        self.read_until (startend[1])
+        return self.extract_via_regex (key + r"\s*" + startend[0]
+                                       + r"\s*[^" + startend[1] + ']+'
+                                       + startend[1])
+
+
--- a/pyiron_atomistics/sphinx/parser_base.py
+++ b/pyiron_atomistics/sphinx/parser_base.py
@@ -17,6 +17,7 @@ import re
 from types import GeneratorType
 import numpy
 
+
 class KeywordTreeParser:
     """
     A base class to parse files block by block via keyword-triggered
@@ -41,14 +42,15 @@ class KeywordTreeParser:
             self.parse (file)
 
     """
-    def __init__ (self,keylevels=[]):
-        if isinstance(keylevels,dict):
-            keylevels = [ keylevels ]
-        elif not isinstance(keylevels,list):
+
+    def __init__(self, keylevels=[]):
+        if isinstance(keylevels, dict):
+            keylevels = [keylevels]
+        elif not isinstance(keylevels, list):
             raise TypeError
         self.keylevels = keylevels
 
-    def parse(self,filename):
+    def parse(self, filename):
         """
         Parse a file using the current keylevels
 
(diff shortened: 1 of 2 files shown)


pyiron_atomistics/sphinx/parser_base.py
```diff
@@ -0,0 +1,188 @@
...
+__maintainer__ = "Christoph Freysoldt"
+__email__ = "freysoldt@mpie.de"
+__status__ = "production"
+__date__ = "Dec 8, 2023"
+
+import re
+from types import GeneratorType
+import numpy
+
+class keyword_tree_parser:
```
samwaseda:
```suggestion
class KeywordTreeParser:
```

pyiron_atomistics/sphinx/parser_base.py
```diff
@@ -0,0 +1,188 @@
...
+)
+__version__ = "1.0"
+__maintainer__ = "Christoph Freysoldt"
+__email__ = "freysoldt@mpie.de"
+__status__ = "production"
+__date__ = "Dec 8, 2023"
+
+import re
+from types import GeneratorType
+import numpy
```
samwaseda: You don't want to use `np` instead?

pyiron_atomistics/sphinx/structure.py
```diff
@@ -2,100 +2,99 @@
...
+__email__ = "freysoldt@mpie.de"
 __status__ = "production"
-__date__ = "Feb 4, 2018"
+__date__ = "Dec 8, 2023"
 
 BOHR_TO_ANGSTROM = (
     scipy.constants.physical_constants["Bohr radius"][0] / scipy.constants.angstrom
 )
 
+class struct_parser(keyword_tree_parser):
```
samwaseda:
```suggestion
class StructParser(KeywordTreeParser):
```

pyiron_atomistics/sphinx/parser_base.py
```diff
@@ -0,0 +1,188 @@
...
+        Appends more lines from input until match is found
+
+        Args:
+           match ... (str) what to wait for
+        Returns: nothing
+        """
+        while not match in self.lineview:
+            self.lineview += next(self.line)
+            self.lineno += 1
+            self.lineFrom = self.line
```
samwaseda: `line_from`?

pyiron_atomistics/sphinx/parser_base.py
```diff
@@ -0,0 +1,188 @@
...
+        Returns: nothing
+        """
+        # --- initialization
+        if len(self.keylevels) == 0:
+           raise KeyError("No parsing functions available in keylevels")
+        filehandle = open(filename)
+        self.line = filehandle.__iter__ ()
+        self.lineview = ''
+        self.filename = filename
+        self.lineno = 0
```
samwaseda: I'm very uneasy with defining this kind of stuff outside of `__init__`, basically because if this class is used without calling `parse`, people would only see `AttributeError`.

samwaseda (commented):
ok after having added all the comments I finally understood what it does, but wouldn't it be easier to parse the entire file once and use whatever is needed? Alternatively, I would suggest to do it based solely on `regex`. I feel that it's very difficult to keep an eye on the consistency on `line`, `lineview`, `lineno` etc.

pyiron_atomistics/sphinx/parser_base.py
pmrv: I think the whole `while` could be replaced by 
```python
with open(file) as filehandle:
  for i, line in enumerate(filehandle):
    self.lineno = i
    ...
```

freyso:
> ok after having added all the comments I finally understood what it does, but wouldn't it be easier to parse the entire file once and use whatever is needed? 

This is what you guys currently do almost everywhere, but it becomes annoying and memory-wasteful when you need to parse very large files. It also renders certain parsing tasks more difficult than the stream-reading. Typical log files, for instance, have "sections" in their output with a rather well-defined beginning, but often no unique end marker. Instead, you will find a new output section, and it may depend on the input which other output sections appears next. The current parsing concept is adapted to this.

A sequential parser can parse GBs of data with a memory demand of a few KB (or whatever the maximum size of input is to be able extract information). 

> Alternatively, I would suggest to do it based solely on `regex`. I feel that it's very difficult to keep an eye on the consistency on `line`, `lineview`, `lineno` etc.

The KeywordTreeParser is supposed to encapsulate the line reading, the actual parser derived from it only uses the lineview. If you can parse from a single line alone, you need not worry at all. If you need more lines, currently there is only the read_until auxiliary, but one could easily think of additional auxiliaries such as "append_line" or "next_line".

pyiron_atomistics/sphinx/parser_base.py
freyso: > I think the whole `while` could be replaced by
> 
> ```python
> with open(file) as filehandle:
>   for i, line in enumerate(filehandle):
>     self.lineno = i
>     ...
> ```

No, because sometimes I need multiple lines at the same time. More importantly, the reading of additional lines can happen elsewhere than in the main parser loop.

freyso: > I'm very uneasy with defining this kind of stuff outside of `__init__`, basically because if this class is used without calling `parse`, people would only see `AttributeError`.

Anyone who implements a parser with such an engine and never calls "parse" didn't get the idea of the engine. Initializing in the parse function would allow to parse multiple times with the same object (if this makes sense...).

pmrv: Ah, I missed the `read_until` parts.  Well, I'd want at least the with statement, because currently I don't see the file ever being closed at all.

pyiron_atomistics/sphinx/parser_base.py
freyso: > You don't want to use `np` instead?

I currently use numpy in a single place for a single function (in get_vector), and I felt using the fully qualified name (numpy.fromstring) improves readability. But I am flexible if you prefer np over numpy.

pmrv:
I like the new architecture and agree that going line by line only once is important for both speed and memory consumption.  But I do have some python nits, that I'll add later.

pyiron_atomistics/sphinx/parser_base.py
```diff
@@ -0,0 +1,188 @@
...
+                    "key2" : self.parse_key2  })
+            self.parse (file)
+
+    """
+    def __init__ (self,keylevels=[]):
+        if isinstance(keylevels,dict):
+            keylevels = [ keylevels ]
+        elif not isinstance(keylevels,list):
+            raise TypeError
+        self.keylevels = keylevels
```
pmrv: Expecting sub classes and base classes to have different signatures for their `__init__` is imo problematic because it violates the [substitution principle](https://en.wikipedia.org/wiki/Liskov_substitution_principle).  If sub classes anyway should call `parse` in their init, can the base class init not be something like

```python
def __init__(self, file, keylevels=[]):
  if isinstance(keylevels,dict):
    keylevels = [ keylevels ]
  elif not isinstance(keylevels,list):
    raise TypeError
  self.keylevels = keylevels
  self.parse(file)
```

then subclasses can just to

```python
super.__init__(file=file, keylevels=...)
```

pyiron_atomistics/sphinx/parser_base.py
```diff
@@ -0,0 +1,188 @@
...
+            else:
+                try:
+                    self.lineview = next(self.line)
+                    self.lineno += 1
+                    self.line_from = self.lineno
+                except StopIteration:
+                    break
+        self._cleanup(self.keylevels[0])
+        if (hasattr(self,'finalize')):
+            self.finalize ()
```
pmrv:
```suggestion
        self.finalize()
    def finalize(self):
        pass
```

This makes it clearer on first glance that the class supports customizing this.

pyiron_atomistics/sphinx/parser_base.py
```diff
@@ -0,0 +1,188 @@
...
+        while True:
+            for keymap in self.keylevels:
+                for key,func in keymap.items ():
+                    if key in self.lineview:
+                        self._cleanup(keymap)
+                        res = func ()
+                        if isinstance(res,GeneratorType):
+                            res.send (None)
+                            keymap['%finalize!'] = res
+                        break
```
pmrv: I like the use of generators, but I wonder if this cannot be streamlined a bit.

If I understand correctly, `_cleanup` is called here first, because we cannot know if the `func` from a previous iteration didn't install some new `keylevels`.  Can we not require that by providing a method on the base parser that allows temporary modification of `keylevels` but also does the clean up as a context manager.  The parser functions/generators `func` can then use this manager around their `yield` statement.

```python
from contextlib import contextmanager
class keyword_tree_parser:
  ...
  @contextmanager
  def install_keylevel(self, new_keylevels):
    self.keylevels.append(new_keylevels)
    yield
    self.keylevels.pop(-1)
```

and sub classes would use this

```python
class sub(keyword_tree_parser):
  ...
  def my_parse_method(self):
    ... # some setup
    with self.install_keylevels({'cell': self.parse_cell}):
      yield
    ... # some processing
```

If we also require all keylevel functions to be generators, the hot part of the loop can then just be

```
res = func()
res.send(None)
```

The back and forth between the parser generator functions and the base class is not quite clear to me, so maybe I'm missing some steps.

pyiron_atomistics/sphinx/structure.py
```diff
@@ -2,100 +2,99 @@
...
+        self.keylevels.append ({
+            'cell': self.parse_cell,
+            'species' : self.parse_species})
+        self.extract_via_regex('structure')
+        # --- initialize for next structure
+        self.cell = None
+        self.positions = []
+        self.species = []
+        self.indices = []
+        self.ispecies=-1
```
pmrv: Do these need to be attributes or can they be just local variables?  They are only used to instantiate `Atoms` one L52 and below, or not?

jan-janssen:
I am wondering if we can combine this with the concepts we developed for the interactive LAMMPS parser. Here is a current example for the LAMMPS parser inside the `atomistics` package. https://github.com/pyiron/atomistics/blob/main/atomistics/calculators/lammps/calculator.py#L111

Writing the LAMMPS parser is easier, as LAMMPS already provides Python bindings, still the general principle is: 
```
def parse_lammps(
    file,
    quantities=("energy", "forces", "stress", ...),
    **kwargs,
):
    interactive_getter_dict = {
        "forces": function_to_parse_force_from_line,
        "energy": function_to_parse_energy_from_line,
        "stress": function_to_parse_stress_from_line,
        ...
    }
    result_dict = {q: [] for q in quantities}
    for line in file:
          for q in quantities:
                interactive_getter_dict[q](previous_result=result_dict[q], line=line)
    return result_dict
```

This approach combines both, the flexibility to parse only for specific properties and the option to parse each line only once.

pmrv:
One thing that makes it generally a bit hard to read is that the file status (current line, file object, etc.) is passed around implicitly via attributes.  It would be more readable, if that was passed around explicitly via arguments to the parser generator functions, imo, but I can't judge how much rewriting that would be.

jan-janssen:
@freyso Are you available to join the pyiron meeting today at 3pm? https://github.com/orgs/pyiron/discussions/207 Then we could start with this topic, that might be faster than the asynchronous discussion and separate developments at different locations.

pyiron_atomistics/sphinx/parser_base.py
freyso: filehandle was implicitly closed by python upon destructing the (local var) filehandle upon leaving the routine. I made that more explicit now, also removing the parse-time only properties.

pyiron_atomistics/sphinx/parser_base.py
freyso: I thought about your suggestion, but I am not convinced. If calling the base class initializer means to do the parsing, the base class initializer must be called at the end of the derived class's __init__ . This is contrary to what most people do elsewhere.

pmrv: Fair, but then I would also keep it out of the sub classes `__init__`.   An alternative would be to have the actual parsers not be sub classes of the keyword tree parser at all, but to implement a has-a relationship instead.

```python
class StructureParser:
  def __init__(self, file):
    ... # setup
    self._parser = KeywordTreeParser(keylevels=...)
    self._parser.parse(file)

File: pyiron_atomistics/sphinx/parser_base.py (hunk 2 above)
Comment (samwaseda) [reply pyiron_atomistics/sphinx/parser_base.py]: Readability is one thing, but if one of us starts editing the file we would definitely write `np` unconsciously, and if it fails, then we would probably simply think `numpy` is not imported at all.

File: pyiron_atomistics/sphinx/parser_base.py [hunk 10]
```diff
@@ -0,0 +1,201 @@
...
+                for key, func in keymap.items():
+                    if key in self.lineview:
+                        self._cleanup(keymap)
+                        res = func()
+                        if isinstance(res, GeneratorType):
+                            res.send(None)
+                            keymap["%finalize!"] = res
+                        break
+                else:
+                    continue
```
samwaseda, line 82:
```suggestion
                    else:
                        continue
```

pyiron_atomistics/sphinx/parser_base.py
```diff
@@ -0,0 +1,201 @@
...
+        while True:
+            for keymap in self.keylevels:
+                for key, func in keymap.items():
+                    if key in self.lineview:
+                        self._cleanup(keymap)
+                        res = func()
+                        if isinstance(res, GeneratorType):
+                            res.send(None)
+                            keymap["%finalize!"] = res
+                        break
```
samwaseda, line 80: From here on there are wrong indentations with different levels

pmrv: I think it's correct, but is making use of the not very well known `for/else` feature of python.

samwaseda: aha ok I didn't know that it was possible

samwaseda:
> This is what you guys currently do almost everywhere, but it becomes annoying and memory-wasteful when you need to parse very large files.

Annoying I don't know, but do we really have so large files in sphinx?

> It also renders certain parsing tasks more difficult than the stream-reading. Typical log files, for instance, have "sections" in their output with a rather well-defined beginning, but often no unique end marker. Instead, you will find a new output section, and it may depend on the input which other output sections appears next. The current parsing concept is adapted to this.

But the end of a section is currently recognised by a marker, right? We can do the same with regex.

> A sequential parser can parse GBs of data with a memory demand of a few KB (or whatever the maximum size of input is to be able extract information).

This part is anyway quite horrible right now in pyiron, because pyiron currently loads the whole input and output anyway. SPHInX allows @pmrv's lazy loading, but it's not supported elsewhere and is so buggy, so I don't think we should rely on this.

> The KeywordTreeParser is supposed to encapsulate the line reading, the actual parser derived from it only uses the lineview. If you can parse from a single line alone, you need not worry at all. If you need more lines, currently there is only the read_until auxiliary, but one could easily think of additional auxiliaries such as "append_line" or "next_line".

I understand the idea, and if python was a human, I would definitely think the way you describe it is how it should work. The reality, however, is that regex is so blazing fast, that anything we do with for loop and string checking becomes totally obsolete. As a matter of fact, reading line by line was what we used to do for SPHInX, but the parsing was slow beyond imagination at that time.

In addition to this, regex makes @jan-janssen's idea also easier to implement, because the output parsers become more modular, because parsing takes place only if the user asks for it, while in your case the parsing must take place again if there's something the user wants to have additionally.

PR #1261, opened 2023-12-15
Fix lammps parser

### LINKED ISSUE #1247: would be nice if pyiron can parse a failed lammps job
Dear all,

I recently realized that while pyiron can parse a failed vasp job to get the atomic positions, energies, etc, it cannot do so for a failed lammps job. This feature would be very useful for me to analyze my failed lammps calculations. I wonder if this could be implemented. 

Best,
Jing

samwaseda:


jyang2009:
Hi Sam,

Thanks a lot for the fix. I've tested it but still got an error when I try `job.collect_output()`
Screen Shot 2023-12-15 at 11 40 27 AM

samwaseda:
ok you are right, but it's a different problem, so I'm gonna merge this one anyway.

PR #1270, opened 2023-12-24
Separate input writer from pyiron

samwaseda:
My Christmas vacation work

Main changes:

- `InputWriter` does not exist anymore - it also made very little sense before
  - `copy_potentials` and `write_spin_constraints` are now independent functions
  - `copy_potentials` does very little right now - I'm not even sure if it merits to be an independent function
- structure group is separated


samwaseda:
As it turned out that there are quite a few things to be done, it’s probably going to stay draft for some time

samwaseda:
Obviously the input writers are not totally separated, but since there are already so many lines that got changed, I’m gonna stop for this PR. Following the [comments from @liamhuber and @niklassiemer](https://github.com/pyiron/pyiron_atomistics/issues/1274) I leave `DataContainer` inside.

PR #1272, opened 2023-12-26
Fixes non parsing of incar/job metainformation from vasprun

### LINKED ISSUE #1226: Vasprun parser doesn't parse everything that is intended
https://github.com/pyiron/pyiron_atomistics/blob/a66edac7bc0e3d2743278babe29651087a060040/pyiron_atomistics/vasp/vasprun.py#L84

This line makes it appear that we actually want to parse most meta data from a vasprun.xml file (version, date, etc.), but this conditional is never hit because `leaf.tag` is never, e.g. 'generator'.  I think the reason is probably because elementtree.iterparse seems to return xml elements depth first, i.e. `leaf` iterates over the tags inside `<generator>` but not that tag itself, but I don't have time to debug this in depth.

ligerzero-ai:
Fixes #1226

### DIFF
--- a/pyiron_atomistics/vasp/vasprun.py
+++ b/pyiron_atomistics/vasp/vasprun.py
@@ -29,7 +29,6 @@ __date__ = "Sep 1, 2017"
 
 
 class Vasprun(object):
-
     """
     This module is used to parse vasprun.xml files and store the data consistent with the pyiron input/output storage
     formats.
@@ -228,9 +227,9 @@ class Vasprun(object):
                                         species_dict[special_element]["n_atoms"] = int(
                                             elements[0].text
                                         )
-                                        species_dict[special_element][
-                                            "valence"
-                                        ] = float(elements[3].text)
+                                        species_dict[special_element]["valence"] = (
+                                            float(elements[3].text)
+                                        )
                                 else:
                                     species_key = elements[1].text
                                     species_dict[species_key] = dict()
--- a/pyiron_atomistics/vasp/vasprun.py
+++ b/pyiron_atomistics/vasp/vasprun.py
@@ -84,8 +84,7 @@ class Vasprun(object):
         for _, leaf in ETree.iterparse(filename):
             if leaf.tag in ["generator", "incar"]:
                 d[leaf.tag] = dict()
-                for items in leaf:
-                    d[leaf.tag] = self.parse_item_to_dict(items, d[leaf.tag])
+                d[leaf.tag] = self.parse_container_element(leaf, d[leaf.tag])
             if leaf.tag in ["kpoints"]:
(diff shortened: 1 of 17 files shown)


pyiron_atomistics/vasp/vasprun.py
```diff
@@ -538,18 +537,108 @@ def parse_structure_to_dict(self, node, d):
...
+                    if c.tag == "i":
+                        params[name] = self._parse_INCAR_generator_parameters(ptype, val)
+                    else:
+                        params[name] = self._parse_INCAR_generator_vector_parameters(ptype, val, self.filename, name)
+                except Exception as exc:
+                    if name == "RANDOM_SEED":
+                        # Handles the case where RANDOM SEED > 99999, which results in *****
+                        params[name] = None
+                    else:
+                        raise exc
```
pmrv, line 630:
```suggestion
                except Exception:
                    if name == "RANDOM_SEED":
                        # Handles the case where RANDOM SEED > 99999, which results in *****
                        params[name] = None
                    else:
                        raise
```
Using raise inside an except block re-raises automatically.  Can we also make the exception more specific?

pyiron_atomistics/vasp/vasprun.py
```diff
@@ -538,18 +537,108 @@ def parse_structure_to_dict(self, node, d):
...
+            val = [i == "T" for i in val.split()]
+        elif val_type == "int":
+            try:
+                val = [int(i) for i in val.split()]
+            except ValueError:
+                # Fix for stupid error in vasprun sometimes which displays
+                # LDAUL/J as 2****
+                val = self._parse_from_incar(filename, param_name)
+                if val is None:
+                    raise err
```
pmrv, line 587:
```suggestion
                    raise err from None
```
This will clear up the stacktrace.

It feels a bit strange, to pre define the exception.  It works and there's no down side, though.

pyiron_atomistics/vasp/vasprun.py
```diff
@@ -538,18 +537,108 @@ def parse_structure_to_dict(self, node, d):
...
+        elif val_type == "string":
+            val = val.split()
+        else:
+            try:
+                val = [float(i) for i in val.split()]
+            except ValueError:
+                # Fix for stupid error in vasprun sometimes which displays
+                # MAGMOM as 2****
+                # val = self._parse_from_incar(filename, param_name)
+                va
```
pmrv, line 597: Feels like there's something missing here?

pmrv (commented):
Thanks for taking this on!  Not sure if you are done with this already, but I left some small remarks anyway.  I like the escape hatch by reading from INCAR, even though it's not used yet.

PR #1303, opened 2024-02-02
Ignore eddrmm warning by default

pmrv:
This is the recommended behavior from upstream:
https://www.vasp.at/forum/viewtopic.php?f=3&t=17822

### DIFF
--- a/pyiron_atomistics/vasp/base.py
+++ b/pyiron_atomistics/vasp/base.py
@@ -1873,6 +1873,8 @@ class Input:
         self.kpoints = Kpoints(table_name="kpoints")
         self.potcar = Potcar(table_name="potcar")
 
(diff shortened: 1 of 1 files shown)


pyiron_atomistics/vasp/base.py
```diff
@@ -1873,7 +1873,7 @@ def __init__(self):
         self.kpoints = Kpoints(table_name="kpoints")
         self.potcar = Potcar(table_name="potcar")
 
-        self._eddrmm = "warn"
+        self._eddrmm = "ignore"
```
pmrv, line 1878:
```suggestion
        # "official" recommendation of VASP devs is to just ignore this warning
        # https://www.vasp.at/forum/viewtopic.php?f=3&t=17822
        self._eddrmm = "ignore"
```

PR #1307, opened 2024-02-04
Limit mp_api import

jan-janssen:
Only import `mp_api` - the API to access the materials project when needed. This accelerates the import by 69%
```
import time: self [us] | cumulative | imported package
import time:      2760 |   18156841 | pyiron_atomistics
import time:    123681 |    5643537 | pyiron_atomistics
```


pmrv:
Superceeded by #1310

PR #1315, opened 2024-02-09
Call `collect_output` when VASP job is not compressed at restart

samwaseda:
When a job is aborted due to runtime, usually `collect_output` is not called at the end, in which case `restart` fails. Here we make sure to call `collect_output` if the job is not compressed.

### DIFF
--- a/pyiron_atomistics/vasp/base.py
+++ b/pyiron_atomistics/vasp/base.py
@@ -1551,6 +1551,9 @@ class VaspBase(GenericDFTJob):
             new_ham (vasp.vasp.Vasp instance): New job
         """
         new_ham = super(VaspBase, self).restart(job_name=job_name, job_type=job_type)
+        if not self.is_compressed():
+            self.collect_output()
+            self.compress()
         if new_ham.__name__ == self.__name__:
             new_ham.input.potcar["xc"] = self.input.potcar["xc"]
         if new_ham.input.incar["MAGMOM"] is not None:


ahmedabdelkawy:
Tested and looks okay!
for NOT_CONVERGED Jobs for other reasons than a runtime error. For example for exceeding maximum number of NELM!
```
vasp_job = pr.create.job.Vasp('runtime_error', True)
vasp_job.structure = pr.create.structure.bulk('Al', cubic=True)
vasp_job.structure.set_cell(0.8 * vasp_job.structure.cell.array, scale_atoms=True)
vasp_job.calc_static()
vasp_job.input.incar["NELM"] = 3
vasp_job.input.incar["EDIFF"]="1E-6"
vasp_job.server.core = 1
vasp_job.server.queue = 'cmti'
vasp_job.run()
```
`vasp_job.is_compressed()`
```
rest_job = vasp_job.restart('runtime_error_rest')
rest_job.run()
```
`vasp_job.is_compressed()`

ahmedabdelkawy:
When a job is aborted due to runtime, the problem is more severe:
1. in the job table, the status is still running
2. the job is never collected

samwaseda:
> When a job is aborted due to runtime, the problem is more severe:
> 
> 1. in the job table, the status is still running
> 2. the job is never collected

ok that's a problem that cannot be solved, since we cannot do anything about the hard kill

PR #1334, opened 2024-02-21
update calphy interface

srmnitc:
Update calphy interface to support versions >= 1.3

### DIFF
--- a/pyiron_atomistics/calphy/job.py
+++ b/pyiron_atomistics/calphy/job.py
@@ -130,10 +130,13 @@ class Calphy(GenericJob, HasStructure):
         self.input.potential_initial_name = None
         self.input.potential_final_name = None
         self.input.structure = None
-        self.output = DataContainer(table_name="output")
+        self._output = DataContainer(table_name="output")
         self.input._pot_dict_initial = None
         self.input._pot_dict_final = None
         self.__version__ = calphy_version
+
+    def output(self):
+        return self._output
 
     @property
     def _default_input(self):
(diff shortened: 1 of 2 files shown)


srmnitc:
@pmrv Still, the job somehow does not collect and populate the output when running on the cluster. Any ideas you have this would also be super useful

jan-janssen:
```
ERROR: test_get_element_list (calphy.test_base.TestCalphy)
  File "/home/runner/work/pyiron_atomistics/pyiron_atomistics/tests/calphy/test_base.py", line 139, in test_get_element_list
    pm, pl = self.job._get_masses()
ValueError: not enough values to unpack (expected 2, got 1)
```

pyiron_atomistics/calphy/job.py
```diff
@@ -720,39 +733,45 @@ def collect_general_output(self):
...
+            f_ediff, b_ediff, flambda, blambda = self._collect_ediff(concentration)
             self.output["fe/forward/energy_diff"] = list(f_ediff)
             self.output["fe/backward/energy_diff"] = list(b_ediff)
             self.output["fe/forward/lambda"] = list(flambda)
             self.output["fe/backward/lambda"] = list(blambda)
 
             # get final structure
-            traj = PyscalTrajectory(
-                os.path.join(self.working_directory, "conf.equilibration.dump")
+            aseobj = read(
```
pmrv:
```suggestion
            aseobj = self.project.create.structure.read(
```

This should save you the `ase_to_pyiron` call below.

pmrv:
> @pmrv Still, the job somehow does not collect and populate the output when running on the cluster. Any ideas you have this would also be super useful

You mean that once the job on the cluster finishes, that the object in your local notebook doesn't show the output?  That's not really solvable so easily, because we have no communication between the process running on the cluster and the process running the notebook.

PR #1338, opened 2024-02-22
Lammps: use recursive=True in read_dict_from_hdf()

jan-janssen:
@pmrv Following our discussion, this uses the `recursive=True` flag in the `read_dict_from_hdf()` function to directly load the whole input. The limitation of this approach is that it is not backwards compatible, as the dictionary no longer contains `input` as primary key.

### DIFF
--- a/pyiron_atomistics/lammps/base.py
+++ b/pyiron_atomistics/lammps/base.py
@@ -740,16 +740,8 @@ class LammpsBase(AtomisticGenericJob):
         """
         super(LammpsBase, self).from_hdf(hdf=hdf, group_name=group_name)
         self._structure_from_hdf()
-        self.input.from_dict(
-            data_dict=self._hdf5.read_dict_from_hdf(
-                [
-                    "input",
-                    "input/control_inp",
-                    "input/potential_inp",
-                    "input/potential_inp/potential",
-                ]
-            )
-        )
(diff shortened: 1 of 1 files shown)


jan-janssen:
Fixed the error message

pmrv:
Seems to work in my testing together with [this](https://github.com/pyiron/pyiron_potentialfit/pull/49).  I'll merge and release this in a bit.

pyiron_atomistics/lammps/base.py
```diff
@@ -1056,10 +1048,10 @@ def _load_default_bond_params(self):
         self.bond_dict["O"]["angle_type_list"] = [1]
 
     def from_dict(self, data_dict):
-        self.control.from_dict(data_dict["input"]["control_inp"])
-        self.potential.from_dict(data_dict["input"]["potential_inp"])
-        if "bond_dict" in data_dict["input"].keys():
-            self.bond_dict = data_dict["input"]["bond_dict"]
+        self.control.from_dict(data_dict["control_inp"])
+        self.potential.from_dict(data_dict["potential_inp"])
```
pmrv:
```suggestion
        self.control.from_dict(data_dict[self.control.table_name])
        self.potential.from_dict(data_dict[self.potential.table_name])
```

PR #1339, opened 2024-02-23
add fix for output not being populated automatically

srmnitc:
Add fix to access output

### DIFF
--- a/pyiron_atomistics/calphy/job.py
+++ b/pyiron_atomistics/calphy/job.py
@@ -137,18 +137,11 @@ class Calphy(GenericJob, HasStructure):
 
     @property
     def output(self):
-        # fast path, if job is finished return output immediately
-        # accessing _status_dict breaks API, but exposed API of JobStatus
-        # automatically reloads the status from the database, which is what we
-        # want to avoid here
-        if self.status._status_dict["finished"]:
-            return self._output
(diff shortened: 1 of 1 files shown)


pyiron_atomistics/calphy/job.py
```diff
@@ -779,7 +772,7 @@ def collect_general_output(self):
             self._output["fe/backward/lambda"] = list(blambda)
 
             # get final structure
-            pyiron_atoms = self.project.create.structure.read(
+            pyiron_atoms = read(
```
pmrv, line 777: Ah, I had missed the project could be a base project instead of an atomistic one, so that's on me.  But if you using ase.read now, you will also need to do the conversion in line 781 again, I think.

jan-janssen: I have a fix for this in https://github.com/pyiron/pyiron_base/pull/1352

pmrv (approved):
lgtm.

https://github.com/pyiron/pyiron_base/issues/1353 Will be good to have in general, but we can merge that separately.

pmrv:
> lgtm.
> 
> [pyiron/pyiron_base#1353](https://github.com/pyiron/pyiron_base/issues/1353) Will be good to have in general, but we can merge that separately.

I actually meant https://github.com/pyiron/pyiron_base/pull/1352

PR #1341, opened 2024-02-23
Lammps: Fix final structure not being stored

### LINKED ISSUE #1340: Changes to lammps output hdf5
The format of lammps h5 output seems to have changed recently. This broke the pyrionFileReader of ovito https://github.com/ovito-org/pyironFileReader.
I implemented a fix in https://github.com/ovito-org/pyironFileReader/pull/1, which reads pbc and species from input/structure instead of
the removed output/structure. This works for both, old and new lammps hdf5 files. However, I am not sure whether it could cause problems due to internal mappings changing species or similar.
So the question is has output/structure been removed because it  doubled data or are there actual differences?

jan-janssen:
I guess this was broken in 0.4.2 - https://github.com/pyiron/pyiron_atomistics/releases/tag/pyiron_atomistics-0.4.2

### DIFF
--- a/pyiron_atomistics/lammps/base.py
+++ b/pyiron_atomistics/lammps/base.py
@@ -459,11 +459,11 @@ class LammpsBase(AtomisticGenericJob):
             ),
             group_name="output",
         )
-        if len(hdf_dict['output/generic/cells']) > 0:
+        if len(hdf_dict["output/generic/cells"]) > 0:
             final_structure = self.structure.copy()
-            final_structure.indices = hdf_dict['output/generic/indices'][-1]
-            final_structure.positions = hdf_dict['output/generic/positions'][-1]
-            final_structure.cell = hdf_dict['output/generic/cells'][-1]
+            final_structure.indices = hdf_dict["output/generic/indices"][-1]
+            final_structure.positions = hdf_dict["output/generic/positions"][-1]
+            final_structure.cell = hdf_dict["output/generic/cells"][-1]
             if final_structure is not None:
                 hdf_dict.update(
                     {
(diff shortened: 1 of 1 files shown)


pyiron_atomistics/lammps/base.py
```diff
@@ -459,9 +459,11 @@ def collect_output(self):
...
         )
-
-        if len(self.output.cells) > 0:
-            final_structure = self.get_structure(iteration_step=-1)
+        if len(hdf_dict["output/generic/cells"]) > 0:
+            final_structure = self.structure.copy()
+            final_structure.indices = hdf_dict["output/generic/indices"][-1]
+            final_structure.positions = hdf_dict["output/generic/positions"][-1]
+            final_structure.cell = hdf_dict["output/generic/cells"][-1]
             if final_structure is not None:
```
pmrv: Can never be `None` now, so if check is unnecessary.

pyiron_atomistics/lammps/base.py
```diff
@@ -459,9 +459,11 @@ def collect_output(self):
             ),
             group_name="output",
         )
-
-        if len(self.output.cells) > 0:
-            final_structure = self.get_structure(iteration_step=-1)
```
pmrv, line 464: Does that mean the `output.cells` and `get_structure` don't work anymore as well or is this somehow restored by writing the final structure below?

jan-janssen: It does not work until the output is stored in the HDF5 file, which only happens at the end of `collect_output()`.

Leimeroth (approved):

pyiron_atomistics/lammps/base.py
jan-janssen: I removed the if statement

PR #1348, opened 2024-02-28
Do not access files accidentely in interactive output

### LINKED ISSUE #1347: loading jobs very slow
For pyiron_base versions 0.7.4 - 0.7.8, loading jobs were very slow compared to previous versions. 

An example job is in the below link compressed file (I can't think of a better way to share it)! Not that this will only be noticed in relatively large jobs > ~100s MB jobs!
https://datashare.mpcdf.mpg.de/s/oi2i125mevjj2K7

pmrv:
Previously we indexed into the job itself, but that also triggers a search for (possibly) compressed files and is slow.  Since we only intend to look into the HDF anyway, we just do it directly now.

May fix https://github.com/pyiron/pyiron_base/issues/1347

### DIFF
--- a/pyiron_atomistics/atomistics/job/atomistic.py
+++ b/pyiron_atomistics/atomistics/job/atomistic.py
@@ -10,6 +10,8 @@ import os
 from typing import Callable, Union, Tuple
 from ase.io import write as ase_write
 import posixpath
+from functools import wraps
+
 from pyiron_atomistics.atomistics.structure.atoms import Atoms
 from pyiron_atomistics.atomistics.structure.neighbors import NeighborsTrajectory
 from pyiron_atomistics.atomistics.structure.has_structure import HasStructure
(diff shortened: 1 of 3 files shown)


pyiron_atomistics/atomistics/job/atomistic.py
```diff
@@ -983,26 +985,38 @@ def load_default(self):
 """
         self.load_string(file_content)
 
+def _suppress_notfound(meth):
+    @wraps(meth)
+    def f(*args, **kwargs):
+        try:
+            return meth(*args, **kwargs)
+        except ValueError:
+            return None
```
pmrv: Raise a warning additionally to annoy everyone.

pyiron_atomistics/sphinx/interactive.py
```diff
@@ -375,14 +375,15 @@ def check_band_occupancy(self, plot=True):
         """
         import matplotlib.pylab as plt
 
-        elec_dict = self._job["output/generic/dft"]["n_valence"]
-        if elec_dict is None:
+        try:
+            elec_dict = self._job.project_hdf5["output/generic/dft"]["n_valence"]
+        except ValueError:
             raise AssertionError("Number of electrons not parsed")
```
pmrv:
```suggestion
            raise AssertionError("Number of electrons not parsed") from None
```

samwaseda (approved):
Fine by me as long as integration tests pass

pmrv:
> Fine by me as long as integration tests pass

Mh, seems this is actually broken on main, as it also pop up in #1426.  I cannot actually run this test locally (hangs indefinitely), do you have an idea?

samwaseda:
Yeah it looks like it's a Sphinx problem right? Somehow it doesn't appear in my environment...

samwaseda:
I just checked everywhere but I just cannot reproduce the error...

samwaseda:
Ah now I updated the conda environment and it started appearing. Now I'll look into it.

samwaseda:
Now as [this PR](https://github.com/pyiron/pyiron_atomistics/pull/1444) got merged hopefully the tests will pass :)

PR #1350, opened 2024-03-01
Use input structure for get_number_of_species

pmrv:
This fixes an issue where tables are getting slowed down a lot for jobs run before #1341 was merged.  For those jobs output/structure/species is missing and accidentally trigger decompression of the jobs.

### DIFF
--- a/pyiron_atomistics/table/funct.py
+++ b/pyiron_atomistics/table/funct.py
@@ -109,7 +109,7 @@ def get_convergence_check(job):
 
 
 def get_number_of_species(job):
-    return {"Number_of_species": len(job["output/structure/species"])}
+    return {"Number_of_species": len(job["input/structure/species"])}
 
 
 def get_number_of_ionic_steps(job):


jan-janssen:
Initially I thought this breaks the compatibility with the `SQSJob` but it stores its structures in `output/structures` so it is not effected. Still this made me wonder if it would be better to search for a structure in multiple places, start with `output/structure` then maybe `output/structures` and finally `input/structure`. Another case where this breaks the functionality is a LAMMPS calculation with variational grand canonical sampling.

pmrv:
> Initially I thought this breaks the compatibility with the `SQSJob` but it stores its structures in `output/structures` so it is not effected. Still this made me wonder if it would be better to search for a structure in multiple places, start with `output/structure` then maybe `output/structures` and finally `input/structure`. 

I can do that.  Personally I like input/, because that's somewhat consistent with the other predefined functions.

> Another case where this breaks the functionality is a LAMMPS calculation with variational grand canonical sampling.

In a round about way actually no, but only because in case of SGC MC a lot of our code breaks when not all elements are present in the initial structure.

PR #1360, opened 2024-03-22
Lammps: fix species for atomstyle charge

### LINKED ISSUE #1353: Lammps not sorting atoms correctly
I am trying to run some reaxff calculations with a potential DF made by this function

```python
def REAXFF_DF(
    path,
    pot_file_name,
    elements,
    keywords='NULL checkqeq yes',
):
    elements_str = ""
    for e in elements:
        elements_str += f" {e}"
    
    if pot_file_name is None:
        fname = path
        pot_file_name = path.split("/")[-1]
    else:
        fname = f"{path}/{pot_file_name}"

    config = [
        f"pair_style reaxff {keywords}\n",
        f"pair_coeff * * {pot_file_name} {elements_str}\n",
        f'fix qeqreaxff all qeq/reaxff 1 0.0 10.0 1e-6 reaxff\n',
    ]
    files = [fname]
    
    pot = pot = pd.DataFrame(
        {
            "Name": "REAXFF",
            "Filename": [files],
            "Model": ["Custom"],
            "Species": [elements],
            "Config": [config],
        }
    )

    return pot
```
This requires to set 
```python
job.input.control['atom_style'] = 'charge'
```
Here I noticed that lammps does not follow the order of elements given by the potential when writing the structure.inp file.
This seems like a very critical bug and used to work as far as I can remember.
In the structure_full function there is a lot of code doing the sorting:

sorted_species_list = np.array(self._potential.get_element_lst())
```python
for species in self.structure.species:
            species_name = species.Abbreviation
            q_dict[species_name] = self.potential.get_charge(species_name)
            species_translate_list.append(self.potential.get_element_id(species_name))
        sorted_species_list = np.array(self._potential.get_element_lst())
```
This is completely lacking in structure_atomic for example. Instead there is 
something like
```python
        for id_eam, el_eam in enumerate(self._el_eam_lst):
            if el_eam in el_struct_lst:
                id_el = list(el_struct_lst).index(el_eam)
                el = el_obj_lst[id_el]
                el_dict[el] = id_eam + 1
```
which however, seems to not work.

Tagging you @pmrv and @jan-janssen, maybe one of you knows whats going on?

jan-janssen:
@Leimeroth I gave it a first try, but unfortunately I do not have an example to test this. Can you provide a test for this function?

### DIFF
--- a/pyiron_atomistics/lammps/structure.py
+++ b/pyiron_atomistics/lammps/structure.py
@@ -586,19 +586,31 @@ class LammpsStructure(object):
 
         masses = "Masses\n\n"
 
-        for ind, obj in enumerate(self._structure.get_species_objects()):
-            masses += "{0:3d} {1:f}".format(ind + 1, obj.AtomicMass) + "\n"
+        el_struct_lst = self._structure.get_species_symbols()
+        el_obj_lst = self._structure.get_species_objects()
+        el_dict = {}
+        for id_eam, el_eam in enumerate(self._el_eam_lst):
+            if el_eam in el_struct_lst:
+                id_el = list(el_struct_lst).index(el_eam)
+                el = el_obj_lst[id_el]
+                el_dict[el] = id_eam + 1
+                masses += "{0:3d} {1:f}".format(id_eam + 1, el.AtomicMass) + "\n"
+            else:
+                # element in EAM file but not used in structure, use dummy for atomic mass
+                masses += "{0:3d} {1:f}".format(id_eam + 1, 1.00) + "\n"
 
         atoms = "Atoms\n\n"
 
         coords = self.rotate_positions(self._structure)
         el_charge_lst = self._structure.get_initial_charges()
-        el_lst = self._structure.get_chemical_symbols()
-        el_alphabet_dict = {}
-        for ind, el in enumerate(self._structure.get_species_symbols()):
-            el_alphabet_dict[el] = ind + 1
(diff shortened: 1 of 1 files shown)


Leimeroth:
This correctly orders the elements which are actually present in the structure.
However, when using atom_style charge no species can be set  in the potential, which are not actually present in the structure, because the wrong number of atoms is written to structure.inp than:
```
Start File for LAMMPS
9 atoms
2 atom types

0. 4.921000003800000 xlo xhi
0. 4.261711015300000 ylo yhi
0. 5.400000095400000 zlo zhi
2.460500001800000 0E-15 0E-15 xy xz yz

Masses

  1 28.085000
  2 15.999000
  3 1.000000
```
Also the mass of C is not correct.
Without atom_style charge it sets the correct  N atom types and masses and than just never assigns the type.
This brings me back to the question why there are so many different structure writing functions that all use different ways of ordering? 
I think extracting the ordering functionality from the well tested atom style atomic writer into a separate function that is used by all writes would be the best option

Leimeroth:
I drafted a version refactoring the sorting in #1361. This is only working for charge and atom styles right now, because I am not sure how it works for bond and full styles.

jan-janssen:
> Also the mass of C is not correct. 

Atoms which are not used in the simulation cell but defined in the potential are set to `1.0`. In this case there are no C atoms in the simulation cell, correct?

PR #1361, opened 2024-03-22
Lammps refactor species ordering

### LINKED ISSUE #1353: Lammps not sorting atoms correctly

Leimeroth:
Refactor the sorting of species to have consisten behavior between different atom styles (see #1353 and #1360)

#TODO
- full and bond atom styles

### DIFF
--- a/pyiron_atomistics/lammps/structure.py
+++ b/pyiron_atomistics/lammps/structure.py
@@ -250,6 +250,20 @@ class LammpsStructure(object):
         self._string_input = input_str
 
     @property
+    def molecule_ids(self):
+        return self._molecule_ids
+
+    @molecule_ids.setter
+    def molecule_ids(self, molecule_ids=None):
+        if molecule_ids is None:
+            if "molecule_ids" in self.structure.get_tags():
+                self._molecule_ids = self.structure.molecule_ids
+            else:
+                self._molecule_ids = np.ones(len(self.structure))
+        else:
+            self._molecule_ids = molecule_ids
+
+    @property
     def el_eam_lst(self):
         """
 
@@ -340,6 +354,7 @@ class LammpsStructure(object):
 
         """
         self._set_lammps_id_dict()
+        self.molecule_ids = None
         # analyze structure to get molecule_ids, bonds, angles etc
         coords = self.rotate_positions(self._structure)
 
@@ -352,7 +367,7 @@ class LammpsStructure(object):
         format_str = "{0:d} {1:d} {2:d} {3:f} {4:f} {5:f} "
         if self._structure.dimension == 3:
             for id_atom, (x, y, z) in enumerate(coords):
-                id_mol = 1
+                id_mol = self.molecule_ids[id_atom]
                 atoms += (
                     format_str.format(
                         id_atom + 1,
@@ -366,7 +381,7 @@ class LammpsStructure(object):
                 )
         elif self._structure.dimension == 2:
             for id_atom, (x, y) in enumerate(coords):
-                id_mol = 1
+                id_mol = self.molecule_ids[id_atom]
                 atoms += (
                     format_str.format(
                         id_atom + 1,
@@ -441,6 +456,7 @@ class LammpsStructure(object):
 
         """
         self._set_lammps_id_dict()
+        self.molecule_ids = None
         coords = self.rotate_positions(self._structure)
 
         # extract electric charges from potential file
@@ -462,7 +478,6 @@ class LammpsStructure(object):
             neighbors = self._structure.get_neighbors_by_distance(
                 cutoff_radius=max_cutoff
             )
-            id_mol = 1
 
             # Draw bonds between atoms is defined in self._bond_dict
             # Go through all elements for which bonds are defined
@@ -524,7 +539,7 @@ class LammpsStructure(object):
             atoms += (
                 format_str.format(
                     id_atom + 1,
-                    id_mol,
+                    self.molecule_ids[id_atom],
                     self._species_lammps_id_dict[el],
                     q_dict[el],
                     coord[0],
--- a/pyiron_atomistics/lammps/structure.py
+++ b/pyiron_atomistics/lammps/structure.py
@@ -406,7 +406,7 @@ class LammpsStructure(object):
                                 bonds.append([ia + 1, ib + 1, b_type])
 
             self.structure.bonds = np.array(bonds)
-        bonds = self.structure.bond
+        bonds = self.structure.bonds
 
         bonds_str = "Bonds \n\n"
         for i_bond, (i_a, i_b, b_type) in enumerate(bonds):
--- a/pyiron_atomistics/lammps/structure.py
+++ b/pyiron_atomistics/lammps/structure.py
@@ -259,7 +259,7 @@ class LammpsStructure(object):
             if "molecule_ids" in self.structure.get_tags():
                 self._molecule_ids = self.structure.molecule_ids
             else:
-                self._molecule_ids = np.ones(len(self.structure))
+                self._molecule_ids = np.ones(len(self.structure), dtype=int)
         else:
             self._molecule_ids = molecule_ids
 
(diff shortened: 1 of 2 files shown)


Leimeroth:
I also modified structure_bond and structure_full. However, I am not sure what half their functionality is supposed to do, so it would be good to if someone has a look. For example in structure_full, there was some molecule id, which however was basically
just a counter? I set it to 1, which is what is done in structure_bond. Both actually seem wrong to me. I guess setting everything to 1 should be fine as long as they are undefined, so presumably that could be added via a tag or something.

jan-janssen:
> I also modified structure_bond and structure_full. However, I am not sure what half their functionality is supposed to do, so it would be good to if someone has a look. For example in structure_full, there was some molecule id, which however was basically just a counter? I set it to 1, which is what is done in structure_bond. Both actually seem wrong to me. I guess setting everything to 1 should be fine as long as they are undefined, so presumably that could be added via a tag or something.

That sounds wrong to me. I guess the only example we used so far was simulating water, so in that case you basically count the oxygen atoms and every water molecule - H2O - gets one molecule ID. In your changes all water molecules together just act as one molecule. Is it necessary to touch the `bond` style at this point? I thought you are primarily using `charge`.

Leimeroth:
> In your changes all water molecules together just act as one molecule.

This behavior is unchanged from the previous. Before it was set via 
```
id_mol, id_species = 1, el_dict[elements[id_atom]]
```
so it was permanently 1 for bond and as 
```
indices = self._structure.indices
 for id_el, id_species in enumerate(indices):
                id_mol += 1
```
so it was basically just the same as lammps index and each atom was a separate molecule.
Hence, my opinion that both previous implementations probably did not do what they are supposed to do at all.
In fact , I only changed the behavior for full and not for bond. Now both at least to the same.

IMO the cleanest solution would be to check  whether molecule indices are defined by the user.
Else leaving them on 1 should be reasonable.

pyiron_atomistics/lammps/structure.py
```diff
@@ -424,11 +462,8 @@ def structure_full(self):
             neighbors = self._structure.get_neighbors_by_distance(
                 cutoff_radius=max_cutoff
             )
-            id_mol = 0
-            indices = self._structure.indices
-            for id_el, id_species in enumerate(indices):
-                id_mol += 1
-                molecule_lst.append([id_el, id_mol, species_translate_list[id_species]])
```
jan-janssen, line 431: Here the molecule ID was modified in the previous version but I have a hard time to see this part in your modified version.

Leimeroth: The previous implementation assigns a new molecule ID for each atom. While I agree that setting it to 1 is wrong, assigning the same mol and atom ids seems rather pointless also.
However, structure_bond sets it to 1, so my change at least leads to consistent behavior between the two methods.

I have to admit that I normally do not use pair styles needing molecule IDs, but I think there is no way to automatically set them correctly. IMO this could be solved via the tag system, as for example done for selective dynamics (see my previous comment in the thread). If wanted, I could implement that I guess, but since I don't really use it someone would need to test it.

jan-janssen: > I have to admit that I normally do not use pair styles needing molecule IDs, but I think there is no way to automatically set them correctly. IMO this could be solved via the tag system, as for example done for selective dynamics (see my previous comment in the thread). If wanted, I could implement that I guess, but since I don't really use it someone would need to test it.

I looked at other codes like ASE https://gitlab.com/ase/ase/-/blob/master/ase/io/lammpsdata.py and they also set the molecule Id to 0 if no molecule is defined. So most likely we are fine. @ahmedabdelkawy Do you know anybody using LAMMPS with molecule IDs? I remember Sudarsan used it for his water simulation at some point, but I am not sure what the status is.

jan-janssen:
Seeing that we do not really have anybody who knows why the current parser is written in the way it is, I wonder if it would be better to replace it with an ASE or pymatgen parser instead. The pymatgen parser is available in https://github.com/materialsproject/pymatgen/blob/master/pymatgen/io/lammps/data.py but from briefly looking at it, I feel the ASE parser would be easier to integrate. Unfortunately, the current ASE version `3.22.1` is not as fully featured as their current master branch, for example the option to write the `masses` is still missing. So for the moment switching to the ASE parser does not seem to be an option.

pyiron_atomistics/lammps/structure.py
Leimeroth: I added a possibility to manually set molecule ids via the tag system

ahmedabdelkawy: No, I am not familiar with the reason why it is implemented this way.

Leimeroth:
With all tests passing I guess this can be merged?

pyiron_atomistics/lammps/structure.py
```diff
@@ -195,6 +195,7 @@ def __init__(self, bond_dict=None, job=None):
         self._bond_dict = bond_dict
         self._force_skewed = False
         self._job = job
+        self._species_lammps_id_dict = {}
```
jan-janssen:
```suggestion
        self._species_lammps_id_dict = {}
        self._molecule_ids = []
```

jan-janssen: Introduce `self._molecule_ids` already in the `__init__()` Function.

pyiron_atomistics/lammps/structure.py
```diff
@@ -269,6 +284,47 @@ def el_eam_lst(self, el_eam_lst):
         """
         self._el_eam_lst = el_eam_lst
 
+    def _set_lammps_id_dict(self, el_eam_lst=None):
```
jan-janssen: Would not it be easier if this would be a static function? The static function would then return the `species_lammps_id_dict`. The reason why I prefer static functions is that they are easier to debug.

pyiron_atomistics/lammps/structure.py
```diff
@@ -269,6 +284,47 @@ def el_eam_lst(self, el_eam_lst):
...
+                    "el_eam_list is empty. Can not determine order of species"
+                )
+
+            else:
+                el_eam_lst = self.el_eam_lst
+
+        for idx, el in enumerate(el_eam_lst):
+            self._species_lammps_id_dict[el] = idx + 1
+
+    def lammps_header(
```
jan-janssen, line 294: Same here, I would also make this a static function.

jan-janssen:
> With all tests passing I guess this can be merged?

I had a few stylistic remarks but apart from this it is fine from my side.

jan-janssen (approved):
Looks good to me

PR #1366, opened 2024-03-26
Add "region" to control.py

prince-mathews:
Allows for the `region` command to be added to the control file for LAMMPS jobs.


pmrv (approved):
lgtm

PR #1379, opened 2024-04-12
changed the generic cells property

### LINKED ISSUE #1378: get_structure() for sphinx jobs gives error
**Summary**

running get_structure() for sphinx job gives error (IndexError: ('cell at step ', 25, ' not found')).  Sphinx Job is a DFT relaxation job. and cell typically doesn't change for each step.

```python
job = pr.load('sphinx_job')
job.get_structure(i) # i>0 gives the error.
```

skatnagallu:
Changed the cells property to repeat based on number of steps to fix issue #1378

### DIFF
--- a/pyiron_atomistics/sphinx/base.py
+++ b/pyiron_atomistics/sphinx/base.py
@@ -2147,7 +2147,8 @@ class Output:
         if "positions" not in self.generic.list_nodes():
             self.generic.positions = np.array([self._job.structure.positions])
         if "cells" not in self.generic.list_nodes():
-            self.generic.cells = np.tile(np.array([self._job.structure.cell.tolist()]),(len(self.generic.energy_tot),1,1))
+            self.generic.cells = np.tile(np.array([self._job.structure.cell.tolist()]),
+                                         (len(self.generic.dft.energy_free),1,1))
         self.generic.to_hdf(hdf=hdf)
 
         with hdf.open("output") as hdf5_output:
--- a/pyiron_atomistics/sphinx/base.py
+++ b/pyiron_atomistics/sphinx/base.py
@@ -2146,7 +2146,8 @@ class Output:
             self.generic.energy_pot = self.generic.dft.energy_free
         if "positions" not in self.generic.list_nodes():
             self.generic.positions = np.array([self._job.structure.positions])
-        if "cells" not in self.generic.list_nodes():
+        if "cells" not in self.generic.list_nodes() and "cell" in self.generic.list_nodes():
+            self.generic.cells = self.generic.cell
             self.generic.cells = np.tile(np.array([self._job.structure.cell.tolist()]),
                                          (len(self.generic.dft.energy_free),1,1))
         self.generic.to_hdf(hdf=hdf)
--- a/pyiron_atomistics/sphinx/base.py
+++ b/pyiron_atomistics/sphinx/base.py
@@ -2146,7 +2146,10 @@ class Output:
             self.generic.energy_pot = self.generic.dft.energy_free
         if "positions" not in self.generic.list_nodes():
             self.generic.positions = np.array([self._job.structure.positions])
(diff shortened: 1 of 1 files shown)


skatnagallu:
Unittest fails with AttributeError: energy_free. I am not sure why

pyiron_atomistics/sphinx/base.py
```diff
@@ -2147,7 +2147,7 @@ def get_last(arr):
         if "positions" not in self.generic.list_nodes():
             self.generic.positions = np.array([self._job.structure.positions])
         if "cells" not in self.generic.list_nodes():
-            self.generic.cells = np.array([self._job.structure.cell.tolist()])
+            self.generic.cells = np.tile(np.array([self._job.structure.cell.tolist()]),(len(self.generic.energy_tot),1,1))
```
```suggestion
np.tile(self._job.structure.cell,(len(self.generic.energy_tot),1,1))
```
This shouldn't have an effect, right?

skatnagallu: prolly not. But the unittest still fails.

freyso:
I think this approach is suboptimal. 

The relaxHist.sx output replicates the cell, so it should be parsable (and it actually seems to be parsed into "output.cell" rather than ...".cells" (with s), see sphinx/output_parsers.py lines 192-196. I'd rather rename the parsing target in the parser, because then we would be ready for future implementations of automatic cell optimization.

samwaseda:
Based on @freyso's comment, I suggest to copy `generic.cell`, which will probably also solve the problem with CI

pyiron_atomistics/sphinx/base.py
```diff
@@ -2147,7 +2147,8 @@ def get_last(arr):
         if "positions" not in self.generic.list_nodes():
             self.generic.positions = np.array([self._job.structure.positions])
         if "cells" not in self.generic.list_nodes():
```
samwaseda:
```suggestion
        if "cells" not in self.generic.list_nodes() and "cell" in self.generic.list_nodes():
            self.generic.cells = self.generic.cell
```

pyiron_atomistics/sphinx/base.py
```diff
@@ -2146,8 +2146,10 @@ def get_last(arr):
             self.generic.energy_pot = self.generic.dft.energy_free
         if "positions" not in self.generic.list_nodes():
             self.generic.positions = np.array([self._job.structure.positions])
-        if "cells" not in self.generic.list_nodes():
-            self.generic.cells = np.array([self._job.structure.cell.tolist()])
+        if "cells" not in self.generic.list_nodes() and "cell" in self.generic.list_nodes():
+            self.generic.cells = self.generic.cell
+            self.generic.cells = np.tile(np.array([self._job.structure.cell.tolist()]),
```
samwaseda, line 2151:
```suggestion
```

Sorry apparently I missed these lines

pyiron_atomistics/sphinx/base.py
```diff
@@ -2146,8 +2146,10 @@ def get_last(arr):
...
+                                         (len(self.generic.dft.energy_free),1,1))
```
samwaseda:
```suggestion
```

Sorry apparently I missed these lines

skatnagallu:
Shouldn't we change the key to 'cells'[ here ](https://github.com/pyiron/pyiron_atomistics/blob/0602a398b0d7149e44f6be27cc37c524a584f67c/pyiron_atomistics/sphinx/output_parser.py#L198) ?

samwaseda:
> Shouldn't we change the key to 'cells'[ here ](https://github.com/pyiron/pyiron_atomistics/blob/0602a398b0d7149e44f6be27cc37c524a584f67c/pyiron_atomistics/sphinx/output_parser.py#L198) ?

I guess we should discuss it at the pyiron meeting. For now if the tests pass we can live with what we get here.

skatnagallu:
can be merged now

samwaseda (approved):
As soon as Black runs I'm fine with merging it

PR #1375, opened 2024-04-09
Add ElasticMatrixJob

jan-janssen:
```python
from pyiron_atomistics import Project

pr = Project("elastic")
job_lmp = pr.create.job.Lammps("lmp")
job_lmp.structure = pr.create.structure.ase.bulk("Al", cubic=True)
job_elastic = pr.create.job.ElasticMatrixJob("elastic")
job_elastic.ref_job = job_lmp
job_elastic.run()
```

PR #1388, opened 2024-04-15
Remove elastic tensor job

jan-janssen:
To minimise the confusion with `ElasticMatrixJob` the `ElasticTensor` job is transferred to `pyiron_contrib`, as discussed in https://github.com/pyiron/pyiron_atomistics/pull/1375


pmrv:
Would be good to use the module conversion setup as in contrib

https://github.com/pyiron/pyiron_contrib/blob/1409ce51314a32af27bdc2273b7a11fc2552d9ea/pyiron_contrib/__init__.py#L93

Users of the old elastic job should be able to use already existing maintenance functions once those module conversions are added.

jan-janssen:
I opened the corresponding pull request in https://github.com/pyiron/pyiron_contrib/pull/1040

PR #1394, opened 2024-04-19
Add OutcarCollectError and catch it in Vasp.collect

pmrv:
If an OUTCAR is present but does not contain NIONS the collect method breaks the Outcar objects raises an unexpected ValueError.

This changes the error to a more specific one and correctly catches the error during collect.


jan-janssen (approved):
Looks good to me, still having a short test for this error would be great.

jan-janssen (approved):
Looks good to me

PR #1395, opened 2024-04-19
Fix vasp potentials for GenericMasters

### LINKED ISSUE #1377: Children of master job not copying potential from the reference job
**Summary**

Children of a master job (ex. a Murnaghan job) do not copy the "potential" from the reference job when they are created. 

pyiron_atomistics version '0.4.17' (cm cluster)

**Expected Behavior**

```python
ref_job.potential['Nb'] = 'Nb_pv'
...
child_job = next(job_murn.iter_jobs())
child_job.potential

>>> {'Nb': 'Nb_pv}
```

**Actual Behavior**

```python
ref_job.potential['Nb'] = 'Nb_pv'
...
child_job = next(job_murn.iter_jobs())
child_job.potential

>>> {'Nb': None}
```

**Steps to Reproduce**

```python
ref_job = pr.create.job.Vasp(job_name='vasp_ref')
ref_job.structure = structure.copy()
ref_job.potential['Nb'] = 'Nb_pv'
ref_job.set_encut(encut)
ref_job.set_kpoints(mesh=kpoint_mesh, scheme='MP')

murn_job = pr.create.job.Murnaghan(job_name='murn_job')
murn_job.ref_job = ref_job
murn_job.input['strains'] = strains.tolist()
murn_job.run()

child_job = next(job_murn.iter_jobs())
child_job.potential
```

jan-janssen:
@raynol-dsouza Can you check if this fixes your issue?

### DIFF
--- a/pyiron_atomistics/vasp/base.py
+++ b/pyiron_atomistics/vasp/base.py
@@ -134,15 +134,7 @@ class VaspBase(GenericDFTJob):
 
         """
         GenericDFTJob.structure.fset(self, structure)
-        if structure is not None and any(
-            [
-                el not in self._potential._potential_dict.keys()
-                for el in set(structure.get_chemical_symbols())
-            ]
-        ):
-            self._potential = VaspPotentialSetter(
-                element_lst=structure.get_species_symbols().tolist()
-            )
+        self._reinit_potential_setter(structure=structure)
 
     @property
     def potential(self):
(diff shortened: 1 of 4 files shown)


raynol-dsouza:
@jan-janssen Thanks! It works correctly now for VASP.

pyiron_atomistics/vasp/base.py
```diff
@@ -1856,6 +1747,127 @@ def list_potentials(self):
         """
         return self.potential_list
 
+    def _set_kpoints(
```
jan-janssen, line 1750: I am just reordering the functions to have the private functions at the end of the class definition.

pyiron_atomistics/vasp/base.py
```diff
@@ -1856,6 +1747,127 @@ def list_potentials(self):
...
+
+    def _reinit_potential_setter(self, structure):
+        if structure is not None:
+            self._potential.to_dict().update(
+                {
+                    el: None
+                    for el in set(structure.get_chemical_symbols())
+                    if el not in self._potential.to_dict().keys()
+                }
+            )
```
jan-janssen, line 1850: This is the only new function I introduce.

pyiron_atomistics/vasp/base.py
```diff
@@ -1856,6 +1747,127 @@ def list_potentials(self):
...
+
+    def _get_path_for_kpoints(self, path_name):
```
jan-janssen, line 1852: I am just reordering the functions to have the private functions at the end of the class definition.

pyiron_atomistics/vasp/base.py
pmrv, line 1850: If I understand correctly this relies on `self._potential.to_dict()` returning its internal `dict` directly and that modifications on the return value are reflected in the object.  That can be dangerous, I think.

```suggestion
            for el in structure.get_species_symbols():
                if el not in self._potential.to_dict():
                    self._potential[el] = None
```

Though it would require that `VaspPotentialSetter.__getitem__/.__getattr__` don't raise errors on new elements anymore.

jan-janssen: > If I understand correctly this relies on `self._potential.to_dict()` returning its internal `dict` directly and that modifications on the return value are reflected in the object. That can be dangerous, I think.

I agree that it is dangerous, that is why I added the test. When a copy of the dictionary is returned the test breaks - I just checked this again. From my perspective this keeps the functionality of the `VaspPotentialSetter` class as defined previously, like raising errors when users mistype element names or try to set an element which does not exist in the structure and it does not directly access private variables which causes IDEs like Pycharm to complain. If anybody is willing to rewrite the handling of pseudo potentials in VASP I am open for suggestions, still that would be a separate pull request. 

So from my perspective this is the minimal change required to fix the bug Raynol reported and I would like to merge it as is, to prevent other users from struggling with the same issue.

pmrv: The only other issue I thought of is:  What happens when the newly set structure has less elements than the `VaspPotentialSetter` had previously defined.  When POTCAR is written are only those elements written that are actually in the structure or all that are in `VaspPotentialSetter`?  I'm not familar enough with the code to immediately tell, but that may cause issues if we accidentally write POTCARs of additional elements.  If that works, go ahead.

jan-janssen: We loop over the element types in the structure when we gather the paths of the potentials:
https://github.com/pyiron/pyiron_atomistics/blob/main/pyiron_atomistics/vasp/potential.py#L429
So having elements in the potential list which do not appear in the structure is not harmful. The more risky part is changing the element type of the structure, like:
```
job_vasp.structure[:] = "Al"
```
This was an issue before and that is still an issue now, but I also do not know a clever way to change it at the moment.

pmrv: Fair enough then. I suppose your case could be avoided if we added the potentials as tags or something in the structure itself, but that's a different change entirely.

PR #1416, opened 2024-05-07
Update the data mining example

jan-janssen:
@HaithamGaafer Maybe you can take a look if you spot anything else we should change. I tried to adopt the changes from your notebook while staying close to the original example.


HaithamGaafer:
It's the same as the one I have.

PR #1418, opened 2024-05-15
Additional inputs to phonopy

raynol-dsouza:
1. In order to calculate the projected DOS (either internally with phonopy or externally using a custom script), one needs the `eigenvectors` of the dynamical matrix at each (ideally irreducible) qpoint. This is by default hard-coded to be `False` in pyiron. I add an additional input to allow to save these eigenvectors if necessary.

2. Phonopy has 2(?) methods for the integration over the Brillouin zone: the `tetrahedron method` (phonopy default, hard coded in pyiron) and the `Gaussian smearing method` (`sigma` calculated internally by phonopy). I also add an additional input to allow to change the default.


raynol-dsouza:
@jan-janssen Thanks for the format black commit!

pmrv (approved):
If I remember correctly one can get the eigenvector also via some post processing from phonopy even from a loaded job, so it might not be necessary to save them and rather just compute them on demand (not sure how long this takes anymore).  But that would be a bigger change and this is self contained and works, so looks good to me.

raynol-dsouza:
@pmrv yes, you are right. But this involves executing `job.phonopy.run_mesh()` again, which is basically repeating what could just be done once. It is also quite memory intensive to do it on demand depending on how dense the qpoint mesh is.

jan-janssen (approved):
Looks good to me - please go ahead and merge the pull request so we include the changes in the next release.

PR #1426, opened 2024-05-24
Refactor Vasp output classes

pmrv:
The vasp.interactive sub module defined subclasses that did not change the functionality of the base classes.  As far as I can tell they were not used except in `VaspInteractive.interactive_close`, so I have removed them but left the imports in place in case any HDF5 files still refer to them.

### DIFF
--- a/pyiron_atomistics/vasp/base.py
+++ b/pyiron_atomistics/vasp/base.py
@@ -2026,6 +2026,7 @@ class Output:
         self.outcar = Outcar()
         self.oszicar = Oszicar()
         self.generic_output = GenericOutput()
+        self.dft_output = DFTOutput()
         self.description = (
             "This contains all the output static from this particular vasp run"
         )
--- a/pyiron_atomistics/vasp/interactive.py
+++ b/pyiron_atomistics/vasp/interactive.py
@@ -345,9 +345,6 @@ class GenericOutput(GenericOutputBase):
         log_dict (dict): A dictionary of all tags and values of generic data (positions, forces, etc)
     """
 
-    def __init__(self):
-        super(GenericOutput, self).__init__()
-
     def to_hdf(self, hdf):
         """
         Save the object in a HDF5 file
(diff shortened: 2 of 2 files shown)


jan-janssen:
I think that is a very reasonable suggestion, originally there was:
```python
    def to_hdf(self, hdf):
        """
        Save the object in a HDF5 file
        Args:
            hdf (pyiron_base.objects.generic.hdfio.ProjectHDFio): HDF path to which the object is to be saved
        """
        with hdf.open("dft") as hdf_dft:
            # hdf_go["description"] = self.description
            for key, val in self.log_dict.items():
                if isinstance(val, list) or isinstance(val, np.ndarray):
                    hdf_dft[key] = val[:-1]
                else:
                    hdf_dft[key] = val
```
This was then fixed around five years ago https://github.com/pyiron/pyiron_atomistics/commit/f2d6e66f16419e15381b83f529d9d4fd5d57ac56 and afterwards nobody cleaned it up.

pyiron_atomistics/vasp/base.py
```diff
@@ -2026,6 +2026,7 @@ def __init__(self):
         self.outcar = Outcar()
         self.oszicar = Oszicar()
         self.generic_output = GenericOutput()
+        self.dft_output = DFTOutput()
```
jan-janssen, line 2029: I feel this is not included in the `to_dict()` and `from_dict()` as the VASP base implementation stores the corresponding entries in `Output.generic_output.dft_log_dict` while the VASP interactive implementation uses `Output.dft_output.log_dict`.

pmrv: I'm not sure how it is handled, but the `vasp.interactive.Output` was literally not used as far as I could tell.  When I create a test vasp job in a notebook (`pr.create.job.Vasp`) its `job._output_parser` object is the `vasp.base.Output`, only when `VaspInteractive.interactive_close` is called this was overwritten.  I can test again with a "proper" interactive vasp job, though, in case there are any weird edge cases.

ligerzero-ai (approved):
LGTM, up to you to see if edge cases exist, but I don't think it's worth the time.

pmrv:
Test failure discussed in #1444, merging anyway.

PR #1432, opened 2024-05-31
[patch] use deprecate from pyiron_snippets

liamhuber:


pyiron_atomistics/atomistics/job/atomistic.py
```diff
@@ -17,8 +17,8 @@
     GenericParameters,
     GenericMaster,
     GenericJob as GenericJobCore,
-    deprecate,
 )
+from pyiron_snippets.deprecate import deprecate
```
jan-janssen, line 21: The import feels a bit redundant `from pyiron_snippets.deprecate import deprecate`. Can we change it to `from pyiron_snippets import deprecate` ?

pyiron_atomistics/atomistics/master/phonopy.py
```diff
@@ -18,7 +18,8 @@
 from pyiron_atomistics.atomistics.structure.phonopy import (
     publication as phonopy_publication,
 )
-from pyiron_base import state, JobGenerator, deprecate
+from pyiron_base import state, JobGenerator
+from pyiron_snippets.deprecate import deprecate
 from pyiron_snippets.import_alarm import ImportAlarm
```
jan-janssen, line 23: Especially this line would be much cleaner as: 
```suggestion
from pyiron_snippets import deprecate, ImportAlarm
```

pyiron_atomistics/atomistics/job/atomistic.py
liamhuber: Certainly not on this version of snippets. Personally, to maintain the idea that snippets is exceedingly lightweight and that all its pieces are independent, I prefer to leave its `__init__` empty. But I'm not totally closed to it so feel free to open an issue on snippets and see what the others think

pyiron_atomistics/atomistics/master/phonopy.py
liamhuber: As mentioned above, I don't want this because if snippets grows I don't want to burn time importing everything in it just because you use one thing in it. But tbh I'm not sure that even with a large module this burns significant time, so if you have data to prove it won't impact performance down the line, then I'm open to it.

pmrv: The current way keeps the different submodules of snippets and their API uncoupled as is intended and so I think it is strongly preferable to importing/exporting everything in the top module.  Especially since it would be for purely aesthetic reasons.  `from pyiron_snippets.deprecate import *` would work just as well, if it's really unbearable.

PR #1438, opened 2024-06-03
[patch] replace logger with snippets version

liamhuber:
i.e. pyiron_base.state.logger.logger with pyiron_snippets.logger.logger. The original location leaves a re-direct for backwards compatibility.


jan-janssen (approved):
Looks good to me

PR #1444, opened 2024-06-06
[patch] remove initial magnetic moments

samwaseda:
Apparently the latest ASE adds default magnetic moments to magnetic atoms, which breaks the Sphinx tests, so I remove the tags if they are present in the tests.


pmrv:
Looks ok for me in the tests, but also seems like something that could bite us otherwise.  Like, I would normally except the structure creation to always return non magnetic structures, unless explicitly asked otherwise.

samwaseda:
I don’t understand why it’s failing but this one causes problems in many repositories, so I merge it

PR #1446, opened 2024-06-10
LAMMPS: Implement calculate() function

jan-janssen:
Switch the LAMMPS job to the `calculate()` function as it is introduced in https://github.com/pyiron/pyiron_base/pull/1477 . The pull request is expected to fail until https://github.com/pyiron/pyiron_base/pull/1477 is merged. 

With these new changes it is possible to evaluate LAMMPS jobs as functions. For a classical LAMMPS job you would use: 
```python

pr = Project("test")
job = pr.create.job.Lammps("lmp")
job.structure = pr.create.structure.ase.bulk("Al", cubic=True)
job.run()
```

This still works, but in addition you can also use an external executor:
```python
from concurrent.futures import ProcessPoolExecutor

pr = Project("test")
job = pr.create.job.Lammps("lmp")
job.structure = pr.create.structure.ase.bulk("Al", cubic=True)

with ProcessPoolExecutor() as exe:
    fs = exe.submit(job.get_calculate_function(), **job.calculate_kwargs)
    shell_output, output_dict, job_crashed = fs.result()

job.save_output(output_dict=output_dict, shell_output=shell_output)
```


pyiron_atomistics/lammps/base.py
```diff
@@ -393,13 +387,32 @@ def write_input(self):
...
+        return {
+            "files_to_create": {
+                "structure.inp": lmp_structure._string_input,
+                "control.inp": "".join(self.input.control.get_string_lst()),
+                "potential.inp": "".join(self.input.potential.get_string_lst()),
+            },
+            "files_to_copy": files_to_copy,
+        }
+
+    def write_input(self):
```
samwaseda: Shouldn't this go to `AtomisticGenericJob`?

jan-janssen: Yes, the corresponding pull request is already open on `pyiron_base` https://github.com/pyiron/pyiron_base/pull/1472 .

samwaseda: Then why is a copy here?

jan-janssen: This is primarily for historic reasons, this was the first pull request, before the one in `pyiron_base`. The copy is removed in https://github.com/pyiron/pyiron_atomistics/pull/1451

samwaseda: But doesn't it make more sense to not add it in the first place in this PR?

jan-janssen: Unless a new version of base is released the tests fail without this temporary function.

samwaseda: I'm ok with it, but it seems to me like a bigger evil to copy and paste lines of code only to be deleted in the next PR if the sole reason to do it is you want to merge it earlier, because it's gonna stay in the commit history and might confuse the future developers. Maybe someone like @pmrv or @liamhuber has an opinion?

pyiron_atomistics/lammps/base.py
```diff
@@ -371,19 +372,12 @@ def enable_h5md(self):
...
         )
 
-    def write_input(self):
-        """
-        Call routines that generate the code specific input files
-
-        Returns:
-
-        """
+    def get_input_file_dict(self):
```
samwaseda: I want a doc string here, which indicates that it's part of the new routine even though inside lammps base it is not obvious.

jan-janssen: Added the Docstring

samwaseda: Actually what I wanted was more something like `executed by GenericJob.write_input`, because this kind of functions which belong to the workflow routines are extremely difficult to follow when you just look at the child classes.

jan-janssen: As we discussed on Monday, the process of converting the job model in pyiron from an object model to a functional approach still takes multiple pull requests, the change to more dictionary based data types is one of them. The second is the introduction of calculate functions, which take these input dictionaries as input, call the executable and return the output as dictionary. For the `ExecutorContainerJob` I created a first draft of this change in https://github.com/pyiron/pyiron_base/pull/1477 . This allows pyiron to directly use the executor infrastructure from `pympipool` not only to execute serial python processes but also call executables. So from the `get_input_file_dict()` function we get the input for the `calculate()` function and we need these `get_input_file_dict()` functions to give expert users the option to directly manipulate the input files before they are written.

samwaseda: Wherever it is used, it is just extremely difficult to figure out for new developers which functions belong to the workflow routine and which ones not. So it's not so much about what this function does, but it's more about why this function is there. It can also be a comment in the code.

jan-janssen: Yes, I agree that during the migration from the previous object based execution where we always transferred the job object from one function to the next to the new more functional approach we have redundancies and the code it harder to read. Still I think this is a necessary migration as we can leverage the whole infrastructure already developed for the submission of functions in the `concurrent.futures.Executor` classes and more recently in `pympipool` as we discussed in the pyiron meeting this Monday.

samwaseda: But I'm not saying you shouldn't do it. I'm saying there should be one line of comment in the code saying why this function exists.

pyiron_atomistics/lammps/base.py
```diff
@@ -371,19 +372,21 @@ def enable_h5md(self):
...
         )
 
-    def write_input(self):
+    def get_input_file_dict(self) -> dict:
         """
-        Call routines that generate the code specific input files
+        Get an hierarchical dictionary of input files. On the first level the dictionary is divided in file_to_create
+        and files_to_copy. Both are dictionaries use the file names as keys. In file_to_create the values are strings
+        which represent the content which is going to be written to the corresponding file. In files_to_copy the values
+        are the paths to the source files to be copied.
```
jan-janssen, line 371:
```suggestion
        are the paths to the source files to be copied.

        The get_input_file_dict() function is called before the write_input() function to convert the input specified on
        the job object to strings which can be written to the working directory as well as files which are copied to the
        working directory. After the write_input() function wrote the input files the executable is called. 
```

jan-janssen: @samwaseda Like so?

samwaseda: Yep. It didn't have to be so detailed but it looks good to me.

samwaseda (approved):

pyiron_atomistics/lammps/base.py
liamhuber: @samwaseda, what you are saying sounds perfectly reasonable to me. If this PR ought to depend on a PR in base, unless there is some pressing need to get it merged and released, just wait for what's needed to be available in base.

jan-janssen:
This comment is no longer valid.

jan-janssen:
Waiting for https://github.com/pyiron/pyiron_base/pull/1477

pmrv (approved):
lgtm

PR #1452, opened 2024-06-10
VASP: Encapsulate the output parser in the parse_vasp_output() function

jan-janssen:
Create a `parse_vasp_output()` function which can be used to parse VASP output files as standalone function:
```python
parse_vasp_output(
    working_directory,
    structure,
    sorted_indices,
)
```

### DIFF
--- a/pyiron_atomistics/vasp/base.py
+++ b/pyiron_atomistics/vasp/base.py
@@ -16,31 +16,23 @@ from pyiron_atomistics.vasp.potential import (
     Potcar,
     strip_xc_from_potential_name,
 )
-from pyiron_atomistics.atomistics.structure.atoms import (
-    Atoms,
-    CrystalStructure,
-    structure_dict_to_hdf,
-    dict_group_to_hdf,
-)
+from pyiron_atomistics.atomistics.structure.atoms import CrystalStructure
 from pyiron_base import state, GenericParameters
 from pyiron_snippets.deprecate import deprecate
-from pyiron_atomistics.vasp.parser.outcar import Outcar, OutcarCollectError
-from pyiron_atomistics.vasp.parser.oszicar import Oszicar
-from pyiron_atomistics.vasp.procar import Procar
+from pyiron_atomistics.vasp.output import (
+    VaspCollectError,
+    Output,
+    get_final_structure_from_file,
+    output_dict_to_hdf,
+    parse_vasp_output,
+)
 from pyiron_atomistics.vasp.structure import read_atoms, write_poscar, vasp_sorter
 from pyiron_atomistics.vasp.vasprun import Vasprun as Vr
-from pyiron_atomistics.vasp.vasprun import VasprunError, VasprunWarning
-from pyiron_atomistics.vasp.volumetric_data import (
-    VaspVolumetricData,
-    volumetric_data_dict_to_hdf,
-)
+from pyiron_atomistics.vasp.vasprun import VasprunError
+from pyiron_atomistics.vasp.volumetric_data import VaspVolumetricData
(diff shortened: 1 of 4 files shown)


pyiron_atomistics/vasp/base.py
```diff
@@ -835,27 +773,12 @@ def get_final_structure_from_file(self, cwd, filename="CONTCAR"):
...
-                )
-            except (IndexError, ValueError, IOError):
-                raise IOError("Unable to read output structure")
-        return input_structure
+        return get_final_structure_from_file(
+            working_directory=cwd,
+            filename=filename,
+            structure=self.structure,
+            sorted_indices=self.sorted_indices,
+        )
```
ligerzero-ai, line 790: Just import from the output.py? why is this necessary in base.py?

jan-janssen: Primarily for backwards compatibility, I am not sure if anybody is using this function at the moment.

pyiron_atomistics/vasp/base.py
```diff
@@ -400,81 +392,27 @@ def write_input(self):
...
-                )
-        return self._output_parser.to_dict()
+    def _store_output(self, output_dict):
+        output_dict_to_hdf(
+            data_dict=output_dict,
+            hdf=self._hdf5,
+            group_name="output",
+        )
+        if len(self._exclude_groups_hdf) > 0 or len(self._exclude_nodes_hdf) > 0:
+            self.project_hdf5.rewrite_hdf5()
```
ligerzero-ai: Can't we just put this directly in collect_output? Is this being used somewhere else?

jan-janssen: Yes, the idea is to move the writing of the input, the execution of the executable and the parsing of the output into a separate function which can then be submitted to an `pympipool` executor. This is implemented in https://github.com/pyiron/pyiron_atomistics/pull/1454 which is based on the changes in this pull request.

ligerzero-ai: if they are the same function (as I am seeing here), what is the advantage of creating a new hidden method vs just calling collect_output?

jan-janssen: The current `collect_output()` function consists of two parts, the parsing of the output files and the storing of the output inside the HDF5 file. These two parts are separated in future into the `parse_vasp_output()` function which could even be called as a standalone output parser and the `job._store_output() function. To remain compatible to the previous interface the `collect_output()` function calls these two functions. In future, the parsing of the output might happen on one compute node while the storing in the HDF5 file happens on the central node of the allocation, to prevent multiple nodes accessing the same HDF5 file.

pyiron_atomistics/vasp/base.py
ligerzero-ai: I think if someone cared enough, they'd use ASE or pymatgen to read from file and convert it over, not worth the additional obfuscation/confusion potential this has. Anyone that knows how to dig this deep for a structure reader probably doesn't need it. my2c

ligerzero-ai (approved):
moves Output, DFTOutput classes, output_dict_to_hdf method from base.py to output.py underneath pyiron_atomistics.vasp

Splits old `collect_output` into `collect_output` and `_store_output`. 

collect_output Method:

    This method is responsible for collecting outputs.
    It calls parse_vasp_output to parse the output data.
    It then calls _store_output to store this parsed data.

_store_output Method:

    This method handles the storage of the parsed output data into the HDF5 file.
    It also includes logic to rewrite the HDF5 file if certain conditions are met (self._exclude_groups_hdf or self._exclude_nodes_hdf).

suggest removal of get_final_structure, but final call up to you.

PR #1453, opened 2024-06-10
Vasp: Refactor bader analysis

jan-janssen:
Rather than requiring the `job` object as input the `Bader` class now only requires the `structure` and the `working_directory`. This simplifies the refactoring of the output in https://github.com/pyiron/pyiron_atomistics/pull/1452

### DIFF
--- a/pyiron_atomistics/dft/bader.py
+++ b/pyiron_atomistics/dft/bader.py
@@ -6,6 +6,9 @@ import numpy as np
 import os
 import subprocess
 
+from pyiron_atomistics.vasp.volumetric_data import VaspVolumetricData
+
+
 __author__ = "Sudarsan Surendralal"
 __copyright__ = (
     "Copyright 2021, Max-Planck-Institut für Eisenforschung GmbH - "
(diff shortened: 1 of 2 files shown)


pyiron_atomistics/dft/bader.py
```diff
@@ -26,22 +29,23 @@ class Bader:
     .. _Bader code: http://theory.cm.utexas.edu/henkelman/code/bader
     """
 
-    def __init__(self, job):
+    def __init__(self, structure, working_directory):
```
ligerzero-ai, line 32: if this now accepts a structure, working directory, instead of job, it will break previous jobs - I don't think anyone is using this code, so just a passing comment here on this change.

ligerzero-ai (approved):
LGTM

pyiron_atomistics/vasp/base.py
```diff
@@ -435,7 +435,9 @@ def collect_output_parser(self, cwd):
         if os.path.isfile(os.path.join(cwd, "AECCAR0")) and os.path.isfile(
             os.path.join(cwd, "AECCAR2")
         ):
-            bader = Bader(self)
+            bader = Bader(
+                working_directory=self.working_directory, structure=self.structure
+            )
             try:
```
jan-janssen, line 441: @ligerzero-ai The `Bader` class is only used internally.

PR #1457, opened 2024-06-12
remove calc_static from DFT

### LINKED ISSUE #428: `fix_symmetry` is overwritten
https://github.com/pyiron/pyiron_atomistics/blob/2f1a45bb7964f127478d746e53549de7191005c7/pyiron_atomistics/dft/job/generic.py#L323

The line above overwrites `job.fix_symmetry` (which is required if the k-point symmetry in the box should be turned off), i.e. the following happens:

```python
>>> job = pr.create.job.Sphinx('foo')
>>> job.structure = pr.create.structure.bulk('Al', cubic=True)
>>> job.fix_symmetry = False
>>> job.calc_static()
>>> print(job.fix_symmetry):
True
```

### LINKED ISSUE #1103: Overwirte the SPINX input file by pyiron
1. I have been trying to do DFT+U calculations using Spinx with Pyiron. And I met a problem when I tried to switch off the symmetry.
2. According to the Spinx manual, I can switch off the symmetry of the structure by creating an empty symmetry group. I have successfully done this by "job_Sphinx.input.sphinx.structure.create_group("symmetry")" and can see the group in "job.input".
3. But if I save the job by "job_Sphinx.save()" and run the calculation "job_Sphinx.run()", pyiron will automatically overwrite the input file which doesn't contain symmetry anymore.
4. I don't know what caused this erroneous overwrite, so in order to conduct a detailed inspection, I have attached my code.
[DFT+U Spinx.txt](https://github.com/pyiron/pyiron_atomistics/files/11950568/DFT%2BU.Spinx.txt)
5. I would like to set the parameters of Sphinx through PyIron without being automatically overwritten by PyIron. Thank you for any contributions.

samwaseda:
It’s mainly to close [this issue](https://github.com/pyiron/pyiron_atomistics/issues/428) and [this issue](https://github.com/pyiron/pyiron_atomistics/issues/1103), but it looks to me that this function shouldn’t have the current form, as the arguments are simply ignored. So I just removed it and let’s see if the tests pass

### DIFF
--- a/pyiron_atomistics/dft/job/generic.py
+++ b/pyiron_atomistics/dft/job/generic.py
@@ -333,7 +333,6 @@ class GenericDFTJob(AtomisticGenericJob):
         ionic_forces=None,
         volume_only=False,
     ):
-        self._generic_input["fix_symmetry"] = True
         super(GenericDFTJob, self).calc_minimize(max_iter=max_iter, pressure=pressure)
 
     def calc_md(
--- a/pyiron_atomistics/dft/job/generic.py
+++ b/pyiron_atomistics/dft/job/generic.py
@@ -318,22 +318,25 @@ class GenericDFTJob(AtomisticGenericJob):
             path_name=path_name,
         )
 
+    def calc_static(
+        self,
+        electronic_steps=60,
+    ):
+        super().calc_static()
+
     def calc_minimize(
         self,
         electronic_steps=60,
         ionic_steps=100,
         max_iter=None,
         pressure=None,
-        algorithm=None,
-        retain_charge_density=False,
-        retain_electrostatic_potential=False,
         ionic_energy_tolerance=None,
         ionic_force_tolerance=None,
         ionic_energy=None,
         ionic_forces=None,
         volume_only=False,
     ):
-        super(GenericDFTJob, self).calc_minimize(max_iter=max_iter, pressure=pressure)
+        super().calc_minimize(max_iter=max_iter, pressure=pressure)
 
     def calc_md(
         self,
@@ -341,12 +344,10 @@ class GenericDFTJob(AtomisticGenericJob):
         n_ionic_steps=1000,
         n_print=1,
         time_step=1.0,
-        retain_charge_density=False,
-        retain_electrostatic_potential=False,
         **kwargs,
     ):
         self._generic_input["fix_symmetry"] = False
-        super(GenericDFTJob, self).calc_md(
+        super().calc_md(
             temperature=temperature,
             n_ionic_steps=n_ionic_steps,
             n_print=n_print,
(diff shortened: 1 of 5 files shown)


pyiron_atomistics/dft/job/generic.py
```diff
@@ -318,16 +318,6 @@ def set_kpoints(
...
         )
 
-    def calc_static(
-        self,
-        electronic_steps=100,
-        algorithm=None,
-        retain_charge_density=False,
-        retain_electrostatic_potential=False,
-    ):
-        self._generic_input["fix_symmetry"] = True
```
pmrv, line 328: Isn't it just this line that causes the problem discussed in the linked issues?  I rather like having `calc_static` on the generic DFT to give us a place where we can enforce generic parameters like `electronic_steps`.  (I suspect that the other three are not actually generic, they could be removed then.)

pmrv:
I'm not to deep into it now, but I saw in the issue that you gave an example of running Sphinx without calling `calc_static`. Imo there should be a call to any `calc_*` method before running an atomistic job, for consistency.

ligerzero-ai:
I'll get to this later tonight

pyiron_atomistics/dft/job/generic.py
samwaseda: Fine, I brought it back.

pyiron_atomistics/sphinx/base.py
```diff
@@ -628,9 +622,6 @@ def calc_static(
 
         Args:
             electronic_steps (float): max # of electronic steps
-            retain_electrostatic_potential:
-            retain_charge_density:
-            algorithm (str): CCG or blockCCG (not implemented)
             electronic_steps (int): maximum number of electronic steps
                 which can be used to achieve convergence
         """
```
ligerzero-ai, line 625: What is this if statement doing (below)? Is electronic steps ever actually None?

ligerzero-ai (approved):
Small question about the if statement in calc_static, else LGTM

pyiron_atomistics/sphinx/base.py
samwaseda: Which if statement?

ligerzero-ai: the one right below, sorry, vscode won't let me select it

samwaseda: ah yeah it's kind of complicated (actually super sketchy) because in `interactive.py` I guess it's `None` by default, which makes it extremely difficult to understand. I'm ready to change it but since it's not really related to this PR I might want to open another one if you think it should be changed.

pyiron_atomistics/dft/job/generic.py
pmrv: Thanks for the additional clean up, too!

PR #1465, opened 2024-06-25
parse ediel_sol from OUTCAR

jyang2009:
This is a minor change of the vasp parser to be able to parse the E_diel_sol in OUTCAR, which is the dielectric solvation energy.


pmrv (approved):
Looks good to me.

PR #1481, opened 2024-07-04
AbstractPotential: remove dead parameter

pmrv:
backward_compatibility_name is simply never used.


jan-janssen (approved):
Looks good to me

PR #1484, opened 2024-07-08
Remove unused argument

pmrv:
n_iterations is not available in get_interstitials from structuretoolkit.


samwaseda:
Ah I wanted to approve it but apparently I merged it instead. I hope it was ok.

PR #1500, opened 2024-07-19
Sphinx: Fix check_setup() for band structure calculation

### LINKED ISSUE #1499: Check_setup needs to be fixed for restarted band structure calculations in sphinz
Restarted sphinx jobs for band structure calculations, creates a "kpoints" group in input, and not "kpoint". So when check_setup() is called for the job, it throws a keyerror.

https://github.com/pyiron/pyiron_atomistics/blob/0f38395584bccb319648822049a5a347b514fe68/pyiron_atomistics/sphinx/base.py#L1561

skatnagallu:
added a if statement to check if "kPoint" is in the input


samwaseda (approved):
I'm fine if the tests pass

PR #1504, opened 2024-07-23
Enable Python 3.9 tests again - for lower bound

jan-janssen:


jan-janssen:
```
  ├─ python 3.9**  is installable with the potential options
  │  ├─ python [3.9.0|3.9.1|...|3.9.9] would require
  │  │  └─ python_abi 3.9.* *_cp39, which can be installed;
  │  └─ python [3.9.0|3.9.1|...|3.9.7], which can be installed;
  └─ scipy 1.14.0**  is not installable because there are no viable options
     ├─ scipy 1.14.0 would require
     │  ├─ python >=3.10,<3.11.0a0 , which conflicts with any installable versions previously reported;
     │  └─ python_abi 3.10.* *_cp310, which conflicts with any installable versions previously reported;
     ├─ scipy 1.14.0 would require
     │  ├─ python >=3.11,<3.12.0a0 , which conflicts with any installable versions previously reported;
     │  └─ python_abi 3.11.* *_cp311, which conflicts with any installable versions previously reported;
     └─ scipy 1.14.0 would require
        └─ python_abi 3.12.* *_cp312, which conflicts with any installable versions previously reported.
```

Issue #1508, opened 2024-07-26
Cell (along with other properties) cannot be accessed from `project_hdf5` in Sphinx


samwaseda:
https://github.com/pyiron/pyiron_atomistics/blob/4361b7b02f89148e249aac04d17dd6eeffdd1bff/pyiron_atomistics/atomistics/job/atomistic.py#L1032

The reason why this line fails is because Sphinx uses `DataContainer` and has tags like `cells__index_2`, which obviously is not the same as `cells` when it's accessed from `project_hdf5`. I guess this topic really requires an overhaul discussion because it's a recurring issue with no structural solution in the current format

jan-janssen:
I have a prototype for accessing the output for Sphinx in the HDF5 file directly https://github.com/pyiron-dev/pyiron-hdf5-format

pmrv:
You should be able to just change `project_hdf5` to `content` in the sphinx output, then it'll be resolve the data container automatically.  I added this to base some time ago, but never got to using it here.

PR #1509, opened 2024-07-28
Use .content rather than .project_hdf5

pmrv:
This allows .output to work with Sphinx jobs that use strange output format.

### DIFF
--- a/pyiron_atomistics/atomistics/job/atomistic.py
+++ b/pyiron_atomistics/atomistics/job/atomistic.py
@@ -1029,22 +1029,22 @@ class GenericOutput(object):
     @property
     @_suppress_notfound
     def cells(self):
-        return self._job.project_hdf5["output/generic/cells"]
+        return self._job.content["output/generic/cells"]
 
     @property
     @_suppress_notfound
     def energy_pot(self):
-        return self._job.project_hdf5["output/generic/energy_pot"]
+        return self._job.content["output/generic/energy_pot"]
 
     @property
     @_suppress_notfound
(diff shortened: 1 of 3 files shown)


jan-janssen:
> This allows .output to work with Sphinx jobs that use strange output format.

Does this break LAMMPS and VASP?

pmrv:
> > This allows .output to work with Sphinx jobs that use strange output format.
> 
> Does this break LAMMPS and VASP?

The only difference between .content and .project_hdf5 is that the former tries to `to_object()` any `DataContainers` it finds, but otherwise returns the same, so I can't see why it would make problems.  Technically .content raises KeyErrors instead of ValueErrors which is what caused the test failures on the weekend, but since .output anyway always silenced those to None, no outside code should depend on this.

@samwaseda Can you add a test case for #1508?

jan-janssen:
@pmrv Should we adjust the pyiron table to also use `content` rather than `output` ?

pmrv:
> @pmrv Should we adjust the pyiron table to also use `content` rather than `output` ?

You mean the predefined datamining functions?  Yes, that would make sense.

jan-janssen (approved):
Looks good to me. Once @samwaseda confirms it fixes his issues, I guess we should merge this and a release a new minor version.

jan-janssen:
@samwaseda Can you confirm this is working for SPHInx? I would like to release a new `pyiron_atomistics` version today.

samwaseda (approved):
I guess it's fine.

PR #1527, opened 2024-08-16
Use ResourceResolvers in lammps/vasp/sphinx

pmrv:
Minimal changes to make atomistics use the resolvers.  See #1528 for clean up.

### DIFF
--- a/.ci_support/environment-old.yml
+++ b/.ci_support/environment-old.yml
@@ -13,7 +13,7 @@ dependencies:
 - pandas =2.0.3
 - phonopy =2.20.0
 - pint =0.18
-- pyiron_base =0.9.4
+- pyiron_base =0.9.12
 - pyiron_snippets =0.1.3
 - pylammpsmpi =0.2.18
 - pyscal3 =3.2.5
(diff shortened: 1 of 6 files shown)


pmrv:
Test failure was due to the backwards version testing.  I've updated `environment-old.yml`, but not sure if that's all it takes to propagate that change to packaging.

samwaseda (approved):
If it passes the integration tests I'm fine.

jan-janssen (approved):
Looks good to me

pmrv:
Like I wrote in #1528 and I added those commits here now as well.

@jan-janssen Is changing `environment-old.yml` enough to update the version bounds in the packages or do I need to change something else as well?

jan-janssen:
> @jan-janssen Is changing `environment-old.yml` enough to update the version bounds in the packages or do I need to change something else as well?

Updating the  `environment-old.yml` is sufficient. The information is combined in the https://github.com/pyiron/pyiron_atomistics/blob/main/.ci_support/release.py script which is called in https://github.com/pyiron/pyiron_atomistics/blob/main/.github/workflows/deploy.yml

PR #1537, opened 2024-08-23
Nicer error message when no potential tables configured

pmrv:
Previously the method raised a ValueError, but #1527 swallowed that by accident.


jan-janssen (approved):
Looks good to me

pmrv:
I think the integration test failure is unrelated and maybe a bit unreliable, because #1542 used to have the same failure but in went away after I commited something unrelated.

PR #1541, opened 2024-08-26
VASP: Add collect_output() function again

jan-janssen:
Primarily for users who want to debug failed jobs


jan-janssen:
@jyang2009 as discussed in the FAQ meeting

PR #1543, opened 2024-08-26
Pass axis in QHA job plotting

pmrv:
Previously `pyplot.colorbar` picked `pyplot.gca()` by default I think, but now it raises a hard error.  So I just resolved this.


jan-janssen (approved):
Looks good to me

PR #1549, opened 2024-08-28
clears output for restarted jobs

skatnagallu:
Temporary fix for the [issue](https://github.com/pyiron/pyiron_atomistics/issues/1575) mentioned in pyiron base. 

This is a temporary solution.

### DIFF
--- a/pyiron_atomistics/sphinx/base.py
+++ b/pyiron_atomistics/sphinx/base.py
@@ -837,7 +837,7 @@ class SphinxBase(GenericDFTJob):
         if recreate_guess:
             new_job.load_guess_group()
 
-        new_job.output.clear_output_for_restart()  # temporary fix for clearing output for restarted jobs (restarted jobs copy output of parent job)
+        new_job.output.clear()  # temporary fix for clearing output for restarted jobs (restarted jobs copy output of parent job)
         return new_job
 
     def relocate_hdf5(self, h5_path=None):
@@ -1894,7 +1894,7 @@ class Output:
         self.generic = None
         self.charge_density = None
         self.electrostatic_potential = None
-        self.clear_output_for_restart()
+        self.clear()
         self.old_version = False
 
     def collect_spins_dat(self, file_name="spins.dat", cwd=None):
@@ -2154,7 +2154,7 @@ class Output:
             self.old_version = True
             pass
 
-    def clear_output_for_restart(self):
+    def clear(self):
         """(Re)set  to a clean state."""
         # This function is part of a temporary fix for restart jobs (see SphinxBase.restart).
         # It could also be used when existing jobs should be reparsed (e.g. after code fixes, updates, etc.)
--- a/pyiron_atomistics/sphinx/base.py
+++ b/pyiron_atomistics/sphinx/base.py
@@ -836,6 +836,8 @@ class SphinxBase(GenericDFTJob):
             )
         if recreate_guess:
             new_job.load_guess_group()
+
+        new_job.output.clear_output_for_restart() # temporary fix for clearing output for restarted jobs (restarted jobs copy output of parent job)
         return new_job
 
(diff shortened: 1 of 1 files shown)


pyiron_atomistics/sphinx/base.py
```diff
@@ -2151,6 +2153,13 @@ def from_hdf(self, hdf):
             self.old_version = True
             pass
 
+    def clear_output_for_restart(self):
```
jan-janssen: Can you call the same `clear_output_for_restart()` function in the `__init__()` function of the Output class? This hopefully prevents us from changing both independent from each other.

pyiron_atomistics/sphinx/base.py
```diff
@@ -1890,10 +1890,7 @@ class Output:
 
     def __init__(self, job):
         self._job = job
-        self.generic = DataContainer(table_name="output/generic")
-        self.charge_density = SphinxVolumetricData()
-        self.electrostatic_potential = SphinxVolumetricData()
-        self.generic.create_group("dft")
+        self.clear_output_for_restart()
```
jan-janssen:
```suggestion
        self.generic = None
        self.charge_density = None
        self.electrostatic_potential = None
        self.clear_output_for_restart()
```
Some code linters complain when the variables are not initialised in the `__init__()` function.

freyso:
What about simply calling the function `clear`  or `setup_fresh` or something like this?
I find it confusing for newcomers to see a function `clear_output_for_restart` in an `__init__` function. It might also be advantageous to keep the function once the output copying issue is fixed, in case one would like to redo the output parsing and ensure that no unexpected items are left in (say: if we discover additional parsing bugs and want to redo the parsing for an existing expensive calculation).

pyiron_atomistics/sphinx/base.py
```diff
@@ -1888,10 +1890,10 @@ class Output:
 
     def __init__(self, job):
         self._job = job
-        self.generic = DataContainer(table_name="output/generic")
-        self.charge_density = SphinxVolumetricData()
-        self.electrostatic_potential = SphinxVolumetricData()
-        self.generic.create_group("dft")
+        self.generic = None
```
freyso, line 1894:
```suggestion
        # --- define internal variables (overwritten in clear_output_for_restart) as None to silence code linters
        self.generic = None
```

pyiron_atomistics/sphinx/base.py
```diff
@@ -2151,6 +2153,13 @@ def from_hdf(self, hdf):
             self.old_version = True
             pass
 
+    def clear_output_for_restart(self):
+        """Clears the output for restarted job. Temporary fix."""
```
freyso:
```suggestion
        """(Re)set  to a clean state."""
        # This function is part of a temporary fix for restart jobs (see SphinxBase.restart).
        # It could also be used when existing jobs should be reparsed (e.g. after code fixes, updates, etc.)
```

jan-janssen (approved):
Looks good to me, still I would like @samwaseda to take a quick look as he is currently the maintainer of the Sphinx interface in pyiron.

samwaseda (approved):
Fine by me.

freyso:
fine for me.

pyiron_atomistics/sphinx/base.py
```diff
@@ -1888,10 +1890,11 @@ class Output:
 
     def __init__(self, job):
         self._job = job
-        self.generic = DataContainer(table_name="output/generic")
-        self.charge_density = SphinxVolumetricData()
-        self.electrostatic_potential = SphinxVolumetricData()
-        self.generic.create_group("dft")
+        # --- define internal variables (overwritten in clear_output_for_restart) as None to silence code linters
```
```suggestion
        # --- define internal variables (overwritten in clear()) as None to silence code linters
```

freyso (approved):

PR #1556, opened 2024-09-06
Sphinx log parser fix

### LINKED ISSUE #1552: Sphinx job parsing issues
1. In SphinxLogParser, the `get_fermi()` function is currently searching for the wrong pattern. The correct pattern should be `pattern = r"Fermi energy:\s+(-?\d+\.\d+)\s+eV"`.

2. It would be beneficial to enhance the `convergence_check()` function by providing more informative print statements that suggest potential actions to take.

3. There appears to be a warning missing in the `set_empty_states()` function if the user input is significantly below the default value. Furthermore, there is an inconsistency in the documentation and default value for non-magnetic calculations. In load_default_groups. For the non-magnetic calculation there should be a 0.5 missing for empty states.

4.   Spin detection in log file is better determined with presence of  "Spin moment:" in log file.
```python
def spin_enabled(self):
    return len(re.findall("Spin moment:", self.log_file)) > 0
```

skatnagallu:
Fixes the issues mentioned [here](https://github.com/pyiron/pyiron_atomistics/issues/1552). All the mentioned issues were causing problems with restarted jobs. In my case restarted jobs for band structure calculations.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,4 @@ apidoc/
 .ipynb_checkpoints/
 test_times.dat
 core.*
+.vscode/settings.json
--- a/pyiron_atomistics/sphinx/base.py
+++ b/pyiron_atomistics/sphinx/base.py
@@ -1235,7 +1235,7 @@ class SphinxBase(GenericDFTJob):
             if self._spin_enabled:
                 self.input["EmptyStates"] = int(1.5 * len(self.structure) + 3)
             else:
-                self.input["EmptyStates"] = int(len(self.structure) + 3)
+                self.input["EmptyStates"] = int(0.5 * len(self.structure) + 3)
 
         if not self.input.sphinx.basis.read_only:
             self.load_basis_group()
@@ -1606,8 +1606,11 @@ class SphinxBase(GenericDFTJob):
                 else:
                     warnings.warn(
                         "Number of empty states was not specified. Default: "
-                        + "3+NIONS for non-magnetic systems"
+                        + "3+NIONS*0.5 for non-magnetic systems"
                     )
+            else:
+                if self.input["EmptyStates"] < 0.5*int(len(self.structure) + 3):
+                    warnings.warn("Number of empty states seem to be too low. Hopefully you know what you are doing.")
 
             return len(w) == 0
 
--- a/pyiron_atomistics/sphinx/base.py
+++ b/pyiron_atomistics/sphinx/base.py
@@ -1488,6 +1488,7 @@ class SphinxBase(GenericDFTJob):
         if not self.nbands_convergence_check():
             print(
                 """
+                The highest orbital is not empty (has a finite occupation).
                 Number of empty states might be too few.
                 Try changing to a higher number of empty states or use default values.
                 """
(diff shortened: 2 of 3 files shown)


pyiron_atomistics/sphinx/base.py
```diff
@@ -1486,7 +1486,21 @@ def convergence_check(self):
         """
         # Checks if sufficient empty states are present
         if not self.nbands_convergence_check():
+            print(
+                """
+                Number of empty states might be too few.
```
freyso, line 1492:
```suggestion
                The highest orbital is not empty (has a finite occupation).
                Number of empty states might be too few.
```

pyiron_atomistics/sphinx/base.py
```diff
@@ -1486,7 +1486,21 @@ def convergence_check(self):
...
+                Try changing to a higher number of empty states or use default values.
+                """
+            )
             return False
+        if not self.output.generic.dft.scf_convergence[-1]:
+            print(
+                """
+                scf convergence not reached. Check residue and number of steps. 
+                If residue is low, try increasing number of steps.
```
freyso:
```suggestion
                If residue goes down systematically, but slowly, try increasing number of steps. 
```

freyso (approved):
Looks fine to me. Minor comments to improve explanations of convergence issues.

skatnagallu:
@samwaseda  can we merge this?

.gitignore
```diff
@@ -11,3 +11,4 @@ apidoc/
 .ipynb_checkpoints/
 test_times.dat
 core.*
+.vscode/settings.json
```
samwaseda, line 14:
```suggestion
```

Or does this have a particular reason?

pyiron_atomistics/sphinx/base.py
```diff
@@ -1592,7 +1607,12 @@ def check_setup(self):
                 else:
                     warnings.warn(
                         "Number of empty states was not specified. Default: "
-                        + "3+NIONS for non-magnetic systems"
+                        + "3+NIONS*0.5 for non-magnetic systems"
+                    )
+            else:
+                if self.input["EmptyStates"] < 0.5 * int(len(self.structure) + 3):
```
samwaseda, line 1613: Isn't this normal for semi conductors? @skatnagallu @freyso

.gitignore
skatnagallu: I use vscode and it usually pushes changes in my settings to git sometimes

jan-janssen: I think it is very helpful for all vscode users, just like including the `.idea/` folder in the `.gitignore` is helpful for Pycharm users. So we could debate if this should be a separate pull request, but in general I think it is a helpful addition.

samwaseda (changes requested):
This PR does more than what the title says. Fix the problem in one PR first then we can talk about changing the default values.

pyiron_atomistics/sphinx/base.py
skatnagallu: For semiconductors this doesnt matter right? You shouldn't set it for semiconductors. And then SPHInX defaults to 0.

skatnagallu:
As suggested, I moved the changes to 3 PRs each dealing with different issues. Shall I close this PR and delete this branch?

jan-janssen:
I am going to close this pull request, as the sphinx output parser is now available as part of https://github.com/pyiron/sphinx_parser .

PR #1567, opened 2024-09-17
Changes default values for empty states and adds warnings for SPHInX

skatnagallu:
As a part of fixing the issues mentioned [here](https://github.com/pyiron/pyiron_atomistics/issues/1552), we thought it would be nice to have some warnings. Also we saw that there was discrepency between default empty states value and documentation of the function.

### DIFF
--- a/pyiron_atomistics/sphinx/base.py
+++ b/pyiron_atomistics/sphinx/base.py
@@ -1490,6 +1490,7 @@ class SphinxBase(GenericDFTJob):
         if not self.nbands_convergence_check():
             print(
                 """
+                The highest orbital is not empty (has a finite occupation).
                 Number of empty states might be too few.
                 Try changing to a higher number of empty states or use default values.
                 """
--- a/pyiron_atomistics/sphinx/base.py
+++ b/pyiron_atomistics/sphinx/base.py
@@ -1488,7 +1488,21 @@ class SphinxBase(GenericDFTJob):
         """
         # Checks if sufficient empty states are present
         if not self.nbands_convergence_check():
+            print(
+                """
+                Number of empty states might be too few.
+                Try changing to a higher number of empty states or use default values.
+                """
+            )
             return False
+        if not self.output.generic.dft.scf_convergence[-1]:
+            print(
+                """
+                scf convergence not reached. Check residue and number of steps. 
+                If residue is low, try increasing number of steps. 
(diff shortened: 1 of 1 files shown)


skatnagallu:
This is the previous [PR](https://github.com/pyiron/pyiron_atomistics/pull/1556#issuecomment-2339763506)

samwaseda:
This is what I don’t understand. You’re right that the default may not have been properly chosen, but the new one appears also arbitrary to me. Can you maybe tell why you chose this value?

skatnagallu:
> This is what I don’t understand. You’re right that the default may not have been properly chosen, but the new one appears also arbitrary to me. Can you maybe tell why you chose this value?

The default was already mentioned in the [docstring](https://github.com/pyiron/pyiron_atomistics/blob/8af2f30a9fdc25ca5093841f6f3cd00c7ad5e4ae/pyiron_atomistics/sphinx/base.py#L1055) of `set_empty_states() `function. I am not sure who originally implemented it. But the value I implemented is based on disscusion with @freyso and also what we saw in VASP, to match what was in the docstring.

skatnagallu:
@samwaseda Are you happy with this?

samwaseda:
ok I originally implemented it but I didn't realise that there was this discrepancy :D I'm still concerned about the warning, which will keep appearing to everyone, regardless of what kind of material we are talking about. I guess it's slightly unfair regarding other more serious warnings.

In addition, I don't quite understand the difference between the two print statements. Isn't one warning for the filling check enough?

skatnagallu:
For convergence there are two convergence checks so two print statements. We thought the print statement for scf convergence would be a helpful hint for new users of sphinx.

samwaseda:
> For convergence there are two convergence checks so two print statements. We thought the print statement for scf convergence would be a helpful hint for new users of sphinx.

Sorry what I meant was `scf_convergence` is set with the `nbands_convergence_check`, so if the first print statement is given, then the second one is not required, right? Maybe I'm badly mistaken somewhere.

skatnagallu:
`scf_convergence` is set from parsing of the log file [here](https://github.com/pyiron/pyiron_atomistics/blob/b1f555df434b64a46bafb2ce794086388fcdd54a/pyiron_atomistics/sphinx/output_parser.py#L394).  And it doesn't have information on the `nbands_convergence_check`.  The final `convergence_check` has both informations in it.

PR #1583, opened 2024-10-10
Add new class to calculate electronic F from DFT

pmrv:
The new class FermiSmearing is accessible via

job.get_electronic_structure().fermi

for all GenericDFTJobs.  it calculates the free energy of the electronic system and can output it as a pandas dataframe.  It also calculates the density at the fermi level for comparison with the Sommerfeld method, though the method itself is not implemented.


jan-janssen:
Looks good to me - Can you add a test?

samwaseda (approved):

PR #1601, opened 2024-11-13
Increase lammps parsing robustness

Leimeroth:
As the title says, this increases the robustness of  parsing lammps output.
The occasion here was a simulation where atoms got lost during the run.
With these changes:
- warnings in the log file are catched and logged with logger
- the amount of atoms in a dump file can change

The latter is done by modifying how and which output is written to hdf5, so it could have impacts on other code parts I guess, would be great if you have a look @jan-janssen

### DIFF
--- a/pyiron_atomistics/lammps/base.py
+++ b/pyiron_atomistics/lammps/base.py
@@ -440,8 +440,8 @@ class LammpsBase(AtomisticGenericJob):
                 }
             )
         else:
-            logger.warning("
-                "The number of atoms changed during the simulation. This can be a sign of massive issues in your simulation."
+            logger.warning(
+                "The number of atoms changed during the simulation. This can be a sign of massive issues in your simulation.\n"
                 "Not storing 'output/structure' to HDF"
             )
         self.project_hdf5.write_dict_to_hdf(data_dict=hdf_dict)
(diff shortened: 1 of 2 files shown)


pyiron_atomistics/lammps/base.py
```diff
@@ -427,10 +428,18 @@ def save_output(
             data_dict=output_dict,
             group_name="output",
         )
-        final_structure = self.structure.copy()
-        final_structure.indices = hdf_dict["output/generic/indices"][-1]
-        final_structure.positions = hdf_dict["output/generic/positions"][-1]
-        final_structure.cell = hdf_dict["output/generic/cells"][-1]
+        try:
```
jan-janssen: Rather than using a `try except` statement, can we compare the length of the initial structure and the indices ?

pyiron_atomistics/lammps/output.py
```diff
@@ -75,7 +76,11 @@ def parse_lammps_output(
 
     for k, v in dump_dict.items():
         if len(v) > 0:
-            hdf_generic[k] = convert_units(np.array(v), label=k)
+            try:
+                hdf_generic[k] = convert_units(np.array(v), label=k)
+            except ValueError:
+                vals = [convert_units(np.array(val), label=k) for val in v]
+                hdf_generic[k] = vals
```
jan-janssen:
```suggestion
                hdf_generic[k] = [convert_units(np.array(val), label=k) for val in v]
```

jan-janssen (approved):
Looks good to me

PR #1622, opened 2024-12-10
Use either pyscal2/3 in calphy job

pmrv:
MPIE env recently switched from pyscal2 to pyscal3 because calphy dropped the dependency on pyscal2.


srmnitc (approved):
Looks good! Thanks for updating.

PR #1627, opened 2024-12-16
add logic to treat multiple log steps

Leimeroth:
Logic to parse output of multiple run commands in a single lammps simulation. Behavior is unchanged for Lammps simulations with a single run command.

### DIFF
--- a/tests/lammps/test_base.py
+++ b/tests/lammps/test_base.py
@@ -858,10 +858,13 @@ class TestLammps(TestWithCleanProject):
         file_directory = os.path.join(
             self.execution_path, "..", "static", "lammps_test_files"
         )
-        _generic_keys_lst, _pressure_dict, _df = _collect_output_log(
+        _generic_keys_lst, _pressure_dict, df = _collect_output_log(
             file_name=os.path.join(file_directory, "multi_log_step_test.lammps"),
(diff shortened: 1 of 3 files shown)


jan-janssen:
@Leimeroth Can you add a corresponding test?

Leimeroth:
Took a while due to vacation. I added a test that simply parses an exemplary log file, so it will fail if someone changes the log function to not handle multiple run commands anymore. Do you want to test for something more specific?

tests/lammps/test_base.py
```diff
@@ -844,6 +844,25 @@ def test_bonds_input(self):
...
+            cell=[3.61, 3.61, 3.61],
+        )
+        file_directory = os.path.join(
+            self.execution_path, "..", "static", "lammps_test_files"
+        )
+        _generic_keys_lst, _pressure_dict, _df = _collect_output_log(
+            file_name=os.path.join(file_directory, "multi_log_step_test.lammps"),
+            prism=UnfoldingPrism(structure.cell),
+        )
+
```
jan-janssen: Can you validate the output here? For example you could check the length of the two individual runs and that all properties are parsed.

jan-janssen (approved):
Looks good to me

PR #1676, opened 2025-02-24
Use pyiron_lammps LAMMPS parser

jan-janssen:
The LAMMPS parser which was previously contained in `pyiron_atomistics` was moved to `pyiron_lammps`. This allows us to use the LAMMPS parser without depending on `pyiron_atomistics` or `pyiron_base`. For example in `atomistics` I plan to add a file based LAMMPS interface which is using the same parser.


jan-janssen:
While the coverage decreased slightly here, the over 1000 lines which were removed here are now 90% covered in https://github.com/pyiron/pyiron_lammps .

PR #1757, opened 2025-07-03
Fix parsing bugs in Calpy when n_print_steps > 0

pmrv:
- Fix typo in calphy trajectory usage

### DIFF
--- a/pyiron_atomistics/calphy/job.py
+++ b/pyiron_atomistics/calphy/job.py
@@ -966,13 +966,13 @@ class Calphy(GenericJob, HasStructure):
 
             def get_pos_cell(block):
                 di = block.to_dict()[0]
-                cc = di['box']
-                if cc.shape[1] == 3: # triclinic box, bail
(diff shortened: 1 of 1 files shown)


pmrv:
@srmnitc The second commit was necessary for me to speed up collecting calphy jobs with a lot of written structures (5x500, roughly), but I couldn't make it general for the triclinic case.  I see that you did this manually in the lammps parser in pyscal, can you adapt something here?

pmrv:
```
FAIL: test_hdf (atomic.structure.test_structurestorage.TestContainer.test_hdf)
Containers written to, then read from HDF should match.
  File "/Users/runner/work/pyiron_atomistics/pyiron_atomistics/tests/atomic/structure/test_structurestorage.py", line 286, in test_hdf
    self.assertTrue(
AssertionError: False is not true : per atom array spins read is not the same as written
```

Cannot reproduce locally, re-running the jobs...

srmnitc (approved):

PR #1759, opened 2025-07-06
Fix release to include multiple pyiron_base versions

jan-janssen:


newfadel (commented):
The pull request modifies `.ci_support/pypi_vs_conda_names.json` to include support for multiple `pyiron_base` versions by changing the value associated with `pyiron_base` from a single version string to a list of versions. This allows the CI pipeline to test `pyiron_base` against various compatible versions, ensuring broader compatibility and more robust releases.

This change is a good practice for enhancing the robustness of the `pyiron_base` releases. It significantly improves testing coverage by validating the package against a range of `pyiron_base` versions, which can help catch potential compatibility issues earlier. This directly contributes to a more stable and reliable library. 😊

PR #1809, opened 2025-09-15
Fix initialization of dicitionaries

jan-janssen:


liamhuber (approved):
Yep!

Aesthetically, I prefer the one-line formulation `calc_kwargs = {} if calc_kwargs is None else calc_kwargs`, but some of the variable names are long enough that might wind up being two lines anyway. At any rate, it's purely taste and not necessary.

PR #1869, opened 2025-10-30
Use pyxtal via assyst

jan-janssen:


jan-janssen:
Wait for https://github.com/eisenforschung/assyst/pull/36

PR #1876, opened 2025-11-10
LAMMPS: Add support for compute_modify

### LINKED ISSUE #1875: How to insert a command line which does not exist in block_dict into control.inp for LAMMPS calculations
I need to use "compute_modify" command in the control.inp file and tried to add it using `job.input.control["compute_modify"]`. But it was complained` "pyiron_log - WARNING - Unknown parameter (does not exist in block_dict): compute_modify"`. Then the command line was added after "run" command, which would have no effect for the whole calculations.

Is there any easy way to add this line?

jan-janssen:


lfzhu-phys:
I would like to know how to apply the modification:) Would it be automatically available? I tried to restart the kernel and restart the server, but the problem remains. How to access the modifications? Thanks a lot in advance.

PR #1954, opened 2026-03-16
Install with uv

niklassiemer:


jan-janssen:
@niklassiemer How does this differ from the `pip check`?

niklassiemer:
The main intent is to _build_ the packages also via the pip/PyPI route with a reliable tool. @XzzX reported, that he is not able to just install packages out of the box with pip.

